PR #2, opened 2019-11-03
Cleanup

franciozzy:
First part of my clean up, mostly cosmetic.

### DIFF
--- a/lib/muser.h
+++ b/lib/muser.h
@@ -439,11 +439,9 @@ lm_get_region(loff_t pos, size_t count, loff_t *off);
 uint8_t *
 lm_get_pci_non_std_config_space(lm_ctx_t *lm_ctx);
 
-#ifdef DEBUG
 void
 dump_buffer(lm_ctx_t *lm_ctx, const char *prefix,
             const unsigned char *buf, uint32_t count);
-#endif
 
 #endif /* LIB_MUSER_H */
 
--- a/lib/muser.h
+++ b/lib/muser.h
@@ -92,7 +92,7 @@ struct lm_sparse_mmap_areas {
 typedef ssize_t (lm_region_access_t) (void *pvt, char * const buf, size_t count,
                                       loff_t offset, const bool is_write);
 
-/** 
+/**
  * Prototype for memory access callback. The program MUST first map device
  * memory in its own virtual address space using lm_mmap, do any additional
  * work required, and finally return that memory. When a region is memory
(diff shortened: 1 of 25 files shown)


lib/python_bindings.c
```diff
@@ -52,7 +52,7 @@ handle_read(char *dst, PyObject *result, int count)
  *
  * FIXME need a way to provide private pointer.
  */
-static ssize_t 
+static ssize_t
 region_access_wrap(void *pvt, char * const buf, size_t count, loff_t offset,
      const bool is_write, int region)
```
swapnili: I think you missed to fix this one. This line should start after '('.

lib/muser.h
```diff
@@ -226,9 +226,9 @@ typedef void (lm_log_fn_t) (void *pvt, const char *const msg);
  *
  * @returns the number of bytes read or written
  */
-typedef ssize_t (lm_cap_access_t) (void *pvt, const uint8_t id,
-                                   char * const buf, size_t count,
-                                   loff_t offset, const bool is_write);
+typedef ssize_t (lm_cap_access_t) (void *pvt, uint8_t id,
+                                   char *buf, size_t count,
+                                   loff_t offset, bool is_write);
```
swapnili: Looks like this can be done in 3 lines, considering the 80C width.
e.g:
typedef ssize_t (lm_cap_access_t) (void *pvt, uint8_t id, char *buf,
                                                            size_t count, loff_t offset, bool is_write);

lib/muser_priv.h
```diff
@@ -36,12 +36,14 @@
 #include "muser.h"
 
 int
-muser_pci_hdr_access(lm_ctx_t * const lm_ctx, size_t * const count,
-                     loff_t * const pos, const bool write,
-                     unsigned char *const buf);
+muser_pci_hdr_access(lm_ctx_t *lm_ctx, size_t *count,
+                     loff_t *pos, bool write,
```
swapnili: This can also fit in two lines, considering the 80C width.

lib/muser.h
franciozzy: Do you mean 2 lines? Yeah, it could. I slightly prefer 3 to keep related variables grouped together (eg. buf and count). I think some of these will need to be further review wrt the usage of const. Let me know if you think I should pack it now or later and I can resend.

lib/muser_priv.h
franciozzy: Yes. I'll repost. The order of these variables is not very natural to me. Maybe we should rearrange them altogether.

lib/muser.h
swapnili: Yes I mean 2 lines. But its fine, you can also fix this later.

lib/python_bindings.c
swapnili: If this is already fixed in later PR's, then please ignore this.

PR #3, opened 2019-11-04
libmuser: Organise lm_ctx_create and friends

franciozzy:
Current state of lm_ctx_create is quite messy. This cleans it up
considerably, moving the pci config space creation to a separate helper.
It gets rid of the 'extended' bool from dev_info, meaning that a client
that wants to have a 4KiB config header should specify it on the
corresponding region.

This may all need reviewing later if we decide that other regions may be
always handled by libmuser, in which case we can hide config, vga (and
maybe rom?) entirely from our clients.

Signed-off-by: Felipe Franciosi <felipe@nutanix.com>

### DIFF
--- a/lib/libmuser.c
+++ b/lib/libmuser.c
@@ -80,7 +80,6 @@ struct lm_ctx {
     void                    *pvt;
     dma_controller_t        *dma;
     int                     fd;
-    bool                    extended;
     int (*reset)            (void *pvt);
     lm_log_lvl_t            log_lvl;
     lm_log_fn_t             *log;
@@ -141,7 +140,7 @@ irqs_disable(lm_ctx_t *lm_ctx, uint32_t index)
         lm_ctx->irqs.type = IRQ_NONE;
         for (i = 0; i < lm_ctx->irqs.max_ivs; i++) {
             if (lm_ctx->irqs.efds[i] >= 0) {
-                (void) close(lm_ctx->irqs.efds[i]);
+                (void)close(lm_ctx->irqs.efds[i]);
                 lm_ctx->irqs.efds[i] = -1;
             }
         }
@@ -665,13 +664,6 @@ lm_get_region(loff_t pos, size_t count, loff_t *off)
 }
 
 static ssize_t
-noop_cb(void *pvt __attribute((unused)), char *buf __attribute((unused)),
-        size_t count __attribute((unused)), loff_t offset __attribute((unused)),
-        bool is_write __attribute((unused))) {
-	return count;
-}
-
-static ssize_t
 handle_pci_config_space_access(lm_ctx_t *lm_ctx, char *buf, size_t count,
                                loff_t pos, bool is_write,
                                lm_region_access_t *pci_config_fn)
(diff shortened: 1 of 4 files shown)


lib/libmuser.c
```diff
@@ -1148,89 +1180,70 @@ lm_ctx_create(lm_dev_info_t *dev_info)
...
-
-    lm_ctx->log = dev_info->log;
-    lm_ctx->log_lvl = dev_info->log_lvl;
-
-    err = lm_caps_init(lm_ctx, dev_info->caps, dev_info->nr_caps);
-
 out:
     if (err) {
         if (lm_ctx) {
+            dma_controller_destroy(lm_ctx, lm_ctx->dma);
```
tmakatos: That should have been an individual commit.

lib/libmuser.c
```diff
@@ -1105,39 +1088,88 @@ free_sparse_mmap_areas(lm_reg_info_t *reg_info)
...
-    if (!nr_caps) {
-        return 0;
+    // Convenience pointer.
+    cfg_reg = &lm_ctx->pci_info.reg_info[LM_DEV_CFG_REG_IDX];
+
+    // Set a default config region if none provided.
+    memset(&zero_reg, 0, sizeof(zero_reg));
+    if (memcmp(cfg_reg, &zero_reg, sizeof(*cfg_reg)) == 0) {
+        cfg_reg->flags = LM_REG_FLAG_RW;
+        cfg_reg->size = PCI_CFG_SPACE_SIZE;
```
tmakatos: Missed `.fn = noop_cb`.

lib/libmuser.c
```diff
@@ -1105,39 +1088,88 @@ free_sparse_mmap_areas(lm_reg_info_t *reg_info)
...
+    } else {
+        // Validate the config region provided.
+        if ((cfg_reg->flags != LM_REG_FLAG_RW) ||
+            ((cfg_reg->size != PCI_CFG_SPACE_SIZE) &&
+             (cfg_reg->size != PCI_CFG_SPACE_EXP_SIZE))) {
+            return EINVAL;
```
tmakatos: A debug message explaining why it failed would be nice. Then the comment could go away.

franciozzy:
> I think this PR should have been broken into smaller commits to make the reviewing process easier.

It probably could... but I only had a few hours to kick off the work and I wanted to get this out there sooner rather than later. My bad. Now that you've reviewed this, let's try to get this in and I'll try to make future commits smaller.

lib/libmuser.c
```diff
@@ -1105,39 +1088,88 @@ free_sparse_mmap_areas(lm_reg_info_t *reg_info)
...
-        lm_log(lm_ctx, LM_ERR, "failed to create PCI capabilities: %m\n");
-        return err;
+    // Allocate a buffer for the config space.
+    lm_ctx->pci_config_space = calloc(1, cfg_reg->size);
+    if (lm_ctx->pci_config_space == NULL) {
+        return -1;
+    }
+
+    // Bounce misc PCI basic header data.
+    memcpy(&lm_ctx->pci_config_space->hdr.id, &dev_info->pci_info.id,
```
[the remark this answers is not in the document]
franciozzy: I have a slight preference for memcpy when it comes to bouncing structs, but agree that this looks better with direct assignment. I will fix it.

lib/libmuser.c
```diff
@@ -1105,39 +1088,88 @@ free_sparse_mmap_areas(lm_reg_info_t *reg_info)
...
+    assert(lm_ctx != NULL);
+    assert(dev_info != NULL);
 
-    if (!nr_caps) {
-        return 0;
+    // Convenience pointer.
+    cfg_reg = &lm_ctx->pci_info.reg_info[LM_DEV_CFG_REG_IDX];
+
+    // Set a default config region if none provided.
+    memset(&zero_reg, 0, sizeof(zero_reg));
```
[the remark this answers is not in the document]
franciozzy: I don't remember why I changed this. Thanks for flagging. I'll move it back to zeroing it on the initialisation (with const, given this is meant to always be zeros), but I don't see value in making it static.

lib/libmuser.c
franciozzy: Probably. As per comment above, I made a trade-off to try and clean up the whole function in one go.

lib/libmuser.c
franciozzy: Actually, I meant to take it out entirely. I'll fix it so that noop_cb actually goes for good. I added it as a hack to get the Piter demo ready, but in reality we don't need it. Libmuser needs to take care of everything config-space related, so let's focus on that. We should actually refuse a config access callback on dev_info if one is provided. On the config space, (I believe) we only need callbacks for the vendor capabilities.

lib/libmuser.c
franciozzy: I agree. But I think we should do a separate review of our debug/log messages and make it all consistent there. Right now, it's a bit unclear when we log what. If that's ok I'll not add this now, but instead do it separately across the entire codebase.

PR #5, opened 2019-11-04
general fixes and cleanup

swapnili:


kmod/muser.c
```diff
@@ -999,7 +996,9 @@ static ssize_t muser_read(struct mdev_device *mdev, char __user *buf,
 	muser_dbg("R %lx@%llx", mucmd.muser_cmd.rw.count,
 		  mucmd.muser_cmd.rw.pos);
 
-	/* Process mudev_cmd in libmuser context. TODO move into function */
+	/* TODO: move following into function */
+
+	/* Process mudev_cmd in libmuser context.*/
```
franciozzy: If we're going to fix the comment, then let's be pedantic on the format. :)

`/* Process mudev_cmd in libmuser context */`

(without the period at the end, or at least with a space before the `*/`.

swapnili: Sure will fix it :).

PR #6, opened 2019-11-05
fix compilations problems for kernel 4.19.67 plus clean up

tmakatos:
Most likely this fixes many more 4.x kernels.


tmakatos:
I messed up the commit message: `access_ok()` is missing from older kernels. Also, the uuid problem can't be fixed without an `#ifdef`. I'll fix both issues with by using an `#ifdef`, which we'll have to remove when upstreaming `muser.ko`.

tmakatos:
Fixed and tested and both kernel versions, please review.

PR #7, opened 2019-11-05
Integrate with Travis CI

franciozzy:
This adds a simple .travis.yml file for integration with Travis CI. It
tries to build muser.git on a Ubuntu Bionic environment using the Linux
kernel 5.2.21.

Signed-off-by: Felipe Franciosi <felipe@nutanix.com>

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,18 @@
+language: c
+dist: bionic
+compiler: gcc
+before_install:
+  - muser_dir=$(pwd)
+  - linux_dir=/tmp/linux
+  - mkdir "${linux_dir}"
+  - cd "${linux_dir}"
+  - git init
+  - git remote add origin git://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git
+  - git fetch --depth 1 origin v5.2.21
+  - git checkout FETCH_HEAD
+  - patch -p1 < "${muser_dir}/patches/vfio.diff"
+  - make olddefconfig
+  - make prepare
+  - cd "${muser_dir}"
+script:
+  - KDIR=/tmp/linux make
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -0,0 +1,17 @@
+language: c
+dist: bionic
+compiler: gcc
+before_install:
+  - src=$(pwd)
+  - mkdir /tmp/linux
```
tmakatos: It would bee better to use `/tmp/linux` from a variable.

.travis.yml
```diff
@@ -0,0 +1,17 @@
...
+  - cd /tmp/linux
+  - git init
+  - git remote add origin git://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git
+  - git fetch --depth 1 origin e91ef5bcdeda8956eb9f1972ed90198b698dca0f # 5.2.21
```
tmakatos: It's cleaner to use the tag `v5.2.21` than the SHA.

lib/libmuser.c
```diff
@@ -45,6 +45,7 @@
 #include <stddef.h>
 #include <sys/mman.h>
 #include <stdarg.h>
+#include <linux/vfio.h>
```
tmakatos: Why is this needed?

.travis.yml
franciozzy: Sure. I'll also fix the other variable names.

.travis.yml
franciozzy: I thought only SHA1 worked (from reading some online doc) and never tried the tag. Tried it now separately and it works, so I'll do that.

lib/libmuser.c
franciozzy: Actually that slipped here, should probably have been a separate commit, but I don't think it's worth separating because only Travis breaks. Travis couldn't build libmuser without it, presumably because `libmuser.c` uses VFIO pound-defined constants that happen to be available via some other include or source file in the same compilation unit. I didn't spend absolutely any time thinking about why it didn't work, I just added it here.

PR #8, opened 2019-11-06
Fix the return values type

swapnili:
Fix the functions which returns negative values but the return type is unsigned. Also return valid errno instead of return -1 in case of errors.

### DIFF
--- a/kmod/muser.c
+++ b/kmod/muser.c
@@ -689,7 +689,7 @@ err:
 	return ret;
 }
 
-static int has_anonymous_pages(struct vfio_dma_mapping *dma_map)
+static bool has_anonymous_pages(struct vfio_dma_mapping *dma_map)
 {
 	int i, nr_pages = NR_PAGES(dma_map->length);
 
@@ -697,11 +697,11 @@ static int has_anonymous_pages(struct vfio_dma_mapping *dma_map)
 		if (PageAnon(dma_map->pages[i])) {
 			muser_dbg("ignore IOVA=%lx, page(s) not shared",
 				  dma_map->iova);
-			return 1;
+			return true;
 		}
 	}
 
-	return 0;
+	return false;
 }
 
 static int muser_iommu_dma_map(struct muser_dev *mudev,
@@ -1107,7 +1107,7 @@ err:
 	return ret;
 }
 
-static unsigned int get_minsz(unsigned int cmd)
+static ssize_t get_minsz(unsigned int cmd)
 {
 	switch (cmd) {
 	case VFIO_DEVICE_GET_INFO:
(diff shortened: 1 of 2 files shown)


kmod/muser.c
```diff
@@ -1119,10 +1119,10 @@ static unsigned int get_minsz(unsigned int cmd)
 	case VFIO_DEVICE_SET_IRQS:
 		return offsetofend(struct vfio_irq_set, count);
 	}
-	return -1;
+	return -EOPNOTSUPP;
```
tmakatos: `EINVAL` would be more suitable.

kmod/muser.c
```diff
@@ -1135,20 +1135,19 @@ static unsigned int get_argsz(unsigned int cmd, struct mudev_cmd *mucmd)
 		return mucmd->muser_cmd.ioctl.data.irq_set.argsz;
 	}
 
-	return -1;
+	return -EOPNOTSUPP;
```
tmakatos: ditto

kmod/muser.c
```diff
@@ -1480,23 +1479,25 @@ static inline int mmap_done(struct mudev_cmd * const mucmd)
...
 		mucmd->pg_map.nr_pages = 0;
 	}
+
 	return ret;
 }
 
-static long libmuser_unl_ioctl(struct file *filep,
-				     unsigned int cmd, unsigned long arg)
+static long libmuser_unl_ioctl(struct file *filep, unsigned int cmd,
+			       unsigned long arg)
```
tmakatos: white space

kmod/muser.c
```diff
@@ -1552,20 +1549,20 @@ static long libmuser_unl_ioctl(struct file *filep, unsigned int cmd,
 		if (ret)
 			goto out;
 
+		mucmd_err = mucmd->muser_cmd.err;
```
tmakatos: Is this just for readability?

kmod/muser.c
swapnili: The invalid(EINVAL) cmds are already checked at vfio layer, vfio only sends us the valid commands. It is at our layer either we support them or not, that is the reason EOPNOTSUPP looks more better.

kmod/muser.c
swapnili: ditto

kmod/muser.c
tmakatos: OK, I think it's worth a comment in our code that VFIO does this.

kmod/muser.c
swapnili: Its strange I do not see a white space in the file. Although the line is been adjusted to be started after the '('.
like: 
static long libmuser_unl_ioctl(struct file *filep, unsigned int cmd,
                                                  unsigned long arg)

kmod/muser.c
swapnili: yes

PR #9, opened 2019-11-06
enable -Wall, -Wextra, and -Werror and fix all warnings

tmakatos:

### DIFF
--- a/lib/dma.h
+++ b/lib/dma.h
@@ -124,7 +124,7 @@ dma_controller_remove_region(dma_controller_t *dma, dma_addr_t dma_addr,
 
 // Helper for dma_addr_to_sg() slow path.
 int
-_dma_addr_sg_split(lm_ctx_t *ctx, const dma_controller_t *dma,
+_dma_addr_sg_split(const dma_controller_t *dma,
                    dma_addr_t dma_addr, uint32_t len,
                    dma_sg_t *sg, int max_sg);
 
@@ -140,7 +140,7 @@ _dma_addr_sg_split(lm_ctx_t *ctx, const dma_controller_t *dma,
  *     necessary to complete this request.
  */
 static inline int
-dma_addr_to_sg(lm_ctx_t *ctx, const dma_controller_t *dma,
+dma_addr_to_sg(const dma_controller_t *dma,
                dma_addr_t dma_addr, uint32_t len,
                dma_sg_t *sg, int max_sg)
 {
@@ -159,7 +159,7 @@ dma_addr_to_sg(lm_ctx_t *ctx, const dma_controller_t *dma,
         return 1;
     }
     // Slow path: search through regions.
-    cnt = _dma_addr_sg_split(ctx, dma, dma_addr, len, sg, max_sg);
+    cnt = _dma_addr_sg_split(dma, dma_addr, len, sg, max_sg);
     if (likely(cnt > 0)) {
         region_hint = sg->region;
     }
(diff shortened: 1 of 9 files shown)


lib/dma.h
```diff
@@ -174,7 +174,12 @@ void
 dma_unmap_region(dma_memory_region_t *region, void *virt_addr, size_t len);
 
 static inline int
-dma_map_sg(dma_controller_t *dma, int prot,
+dma_map_sg(dma_controller_t *dma,
+#if DMA_MAP_FAST_IMPL
```
swapnili: Should we get rid of DMA_MAP_FAST_IMPL? I do not understand what is the use of it.
With DMA_MAP_FAST_IMPL disabled I do not know how the dma_mapping will work.

lib/libmuser.c
```diff
@@ -170,7 +170,8 @@ irqs_set_data_none(lm_ctx_t *lm_ctx, struct vfio_irq_set *irq_set)
     long ret;
     eventfd_t val;
 
-    for (i = irq_set->start; i < irq_set->start + irq_set->count; i++) {
+    for (i = (int)irq_set->start; i < (int)(irq_set->start + irq_set->count);
```
swapnili: Instead of typecasting it everywhere, you can just declare 'i' as unsigned int ?

lib/libmuser.c
```diff
@@ -196,7 +197,8 @@ irqs_set_data_bool(lm_ctx_t *lm_ctx, struct vfio_irq_set *irq_set, void *data)
 
     assert(data != NULL);
 
-    for (i = irq_set->start, d8 = data; i < irq_set->start + irq_set->count;
+    for (i = (int)irq_set->start, d8 = data;
```
swapnili: Same 'i' can be declared as unsigned int.

lib/libmuser.c
```diff
@@ -220,7 +222,8 @@ irqs_set_data_eventfd(lm_ctx_t *lm_ctx, struct vfio_irq_set *irq_set, void *data
     int efd, i;
 
     assert(data != NULL);
-    for (i = irq_set->start, d32 = data; i < irq_set->start + irq_set->count;
+    for (i = (int)irq_set->start, d32 = data;
+         i < (int)(irq_set->start + irq_set->count);
```
swapnili: ditto

lib/libmuser.c
tmakatos: It's better to use signed because overflow of a signed int is defined while overflow of unsigned is not.

lib/dma.h
tmakatos: `DMA_MAP_FAST_IMPL` is the default behavior, see the top of the file for more details. In any case this is beyond the scope of this patch.

swapnili: Yes I had a look into the comment, the comment and the macros looks misleading it also needs to be fixed. Also It does not talks about DMA_MAP_FAST_IMPL.
Anyways I think currently with DMA_MAP_FAST_IMPL disabled, DMA_MAPS will not work, so these ifdefs with macros are not required and we can just get rid of it.
If you agree I will fix this and send the patch around.

lib/libmuser.c
swapnili: irq_set->start and irq_set->count are __u32(unsigned int). So if 'i' is (int) in reality it cannot accommodate all the values of __u32 it will wrap-around.
Its other way the overflow of unsigned is defined,
https://www.gnu.org/software/autoconf/manual/autoconf-2.63/html_node/Integer-Overflow-Basics.html

lib/dma.h
tmakatos: sure

lib/libmuser.c
tmakatos: you're right, I'll fix it.

tmakatos: fixed

lib/libmuser.c
tmakatos: fixed

lib/libmuser.c
tmakatos: fixed

PR #11, opened 2019-11-12
README improvements

tmakatos:


README.md
```diff
@@ -168,3 +168,12 @@ Making libmuser Multi-threaded
...
 implement this we'll have to maintain a private context in struct file.
+
+Troubleshooting
+---------------
+
+If you get the following error when starting QEMU:
+
+    qemu-system-x86_64: -device vfio-pci,sysfsdev=/sys/bus/mdev/devices/00000000-0000-0000-0000-000000000000: vfio 00000000-0000-0000-0000-000000000000: failed to read device config space: Bad address
+
+it means that you haven't properly patched your kernel.
```
swapnili: I am not sure if the reason for this error is always improper patched kernel. Should this be rephrased to something like "it may mean that you haven't properly applied the patch to the kernel"?

tmakatos: Good point, I'll rephrase it.

PR #12, opened 2019-11-13
Fix pci config space read

tmakatos:


lib/libmuser.c
```diff
@@ -663,29 +664,33 @@ lm_get_region(loff_t pos, size_t count, loff_t *off)
...
-    r1 = cap_maybe_access(lm_ctx->caps, lm_ctx->pvt, buf, count, pos, is_write);
-    if (r1 < 0) {
+    count = MIN(pci_config_space_size(lm_ctx), count);
+    ret = cap_maybe_access(lm_ctx->caps, lm_ctx->pvt, buf, count, pos, is_write);
+    if (ret < 0) {
         lm_log(lm_ctx, LM_ERR, "bad access to capabilities %u@%#x\n", count,
                pos);
-        return r1;
-    }
-    buf += r1;
```
swapnili: It looks like after this change, pci_config_fn for LM_DEV_CFG_REG_IDX will never be called even if it falls in PCI_STD_HEADER_SIZEOF?

tmakatos: `pci_config_fn` should never be called for `PCI_STD_HEADER_SIZEOF`. Now that you raised this point, the question is whether there should be a user callback for accesses past the header and the capabilities. I seem to remember that we should handle all accesses to the PCI cofiguration space but I might be wrong. @franciozzy ?

franciozzy: I don't think there should be any callbacks for the config space besides vendor capabilities. We originally tried to make it so config accesses that were not handled by libmuser would fallback somehow, but I don't think that was a good decision. Libmuser handle everything there.

PR #17, opened 2019-11-13
forward support for other types of devices (non-PCI) plus relevant cleanup

### LINKED ISSUE #13: allow device types other than PCI

tmakatos:
closes #13, #17

### DIFF
--- a/lib/libmuser.c
+++ b/lib/libmuser.c
@@ -1133,8 +1133,9 @@ pci_config_setup(lm_ctx_t *lm_ctx, const lm_dev_info_t *dev_info)
     }
 
     // Initialise capabilities.
-    if (dev_info->nr_caps > 0) {
-        lm_ctx->caps = caps_create(dev_info->caps, dev_info->nr_caps);
+    if (dev_info->pci_info.nr_caps > 0) {
+        lm_ctx->caps = caps_create(dev_info->pci_info.caps,
+                                   dev_info->pci_info.nr_caps);
         if (lm_ctx->caps == NULL) {
             lm_log(lm_ctx, LM_ERR, "failed to create PCI capabilities: %m\n");
             goto err;
--- a/lib/libmuser.c
+++ b/lib/libmuser.c
@@ -1167,6 +1167,11 @@ lm_ctx_create(const lm_dev_info_t *dev_info)
         return NULL;
     }
 
+    if (dev_info->dev_type != LM_DEV_TYPE_PCI) {
+        errno = EINVAL;
+        return NULL;
+    }
+
     /*
      * FIXME need to check that the number of MSI and MSI-X IRQs are valid
      * (1, 2, 4, 8, 16 or 32 for MSI and up to 2048 for MSI-X).
--- a/lib/libmuser.c
+++ b/lib/libmuser.c
@@ -77,6 +77,13 @@ typedef struct {
     _Static_assert(sizeof(s) - offsetof(s, m) == sizeof(t), \
         #t " " #m " must be last member in " #s)
 
+struct pci {
+    uint32_t                irq_count[LM_DEV_NUM_IRQS];
+    lm_reg_info_t           reg_info[LM_DEV_NUM_REGS];
+    lm_pci_config_space_t   *config_space;
+    struct caps             *caps;
+};
+
 struct lm_ctx {
     void                    *pvt;
     dma_controller_t        *dma;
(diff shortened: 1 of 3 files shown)


lib/libmuser.c
```diff
@@ -77,16 +77,21 @@ typedef struct {
     _Static_assert(sizeof(s) - offsetof(s, m) == sizeof(t), \
         #t " " #m " must be last member in " #s)
 
+struct pci {
```
franciozzy: Should this be `lm_ctx_pci`?

lib/libmuser.c
```diff
@@ -77,16 +77,21 @@ typedef struct {
...
     void                    *pvt;
     dma_controller_t        *dma;
     int                     fd;
     int (*reset)            (void *pvt);
     lm_log_lvl_t            log_lvl;
     lm_log_fn_t             *log;
-    lm_pci_info_t           pci_info;
-    lm_pci_config_space_t   *pci_config_space;
-    struct caps             *caps;
+    struct pci              pci;
```
franciozzy: Do we want to put this in a union already, or leave it for later?

lib/libmuser.c
```diff
@@ -1207,8 +1212,13 @@ lm_ctx_create(const lm_dev_info_t *dev_info)
     lm_ctx->log_lvl = dev_info->log_lvl;
     lm_ctx->reset = dev_info->reset;
 
-    // Bounce the provided pci_info into the context.
-    memcpy(&lm_ctx->pci_info, &dev_info->pci_info, sizeof(lm_pci_info_t));
+    // Bounce info for IRQs and regions.
+    assert(sizeof(lm_ctx->pci.irq_count) == sizeof(dev_info->pci_info.irq_count));
```
franciozzy: These should be static asserts closer to the definition of these structs. You only want the test to happen during compilation, not at runtime every time.

lib/libmuser.c
```diff
@@ -1207,8 +1212,13 @@ lm_ctx_create(const lm_dev_info_t *dev_info)
...
+    memcpy(lm_ctx->pci.irq_count, dev_info->pci_info.irq_count,
+           sizeof(lm_ctx->pci.irq_count));
+    assert(sizeof(lm_ctx->pci.reg_info) == sizeof(dev_info->pci_info.reg_info));
```
franciozzy: Same here.

lib/muser.h
```diff
@@ -165,6 +165,45 @@ enum {
...
+     */
+    size_t size;
+
+    /*
+     * Function to call back when the capability gets read or written.
+     */
+    lm_cap_access_t *fn;
+} lm_cap_t;
+
+#define LM_MAX_CAPS (PCI_CFG_SPACE_SIZE - PCI_STD_HEADER_SIZEOF) / PCI_CAP_SIZEOF
```
franciozzy: Math in defines (the whole thing) should always be within brackets.
(I can see this was already wrong, but this is a good opportunity to fix it.)

lib/muser.h
```diff
@@ -188,6 +227,19 @@ typedef struct {
      * Class code.
      */
     lm_pci_hdr_cc_t     cc;
+
+    /*
+     * PCI capabilities. The user needs to only define the ID and size of each
+     * capability. The actual capability is not maintained by libmuser. When a
+     * capability is accessed the appropriate callback function is called.
+     */
+    lm_cap_t        caps[LM_MAX_CAPS];
```
franciozzy: I think the alignment got lost, or there's a mix of spaces and tabs (not sure how github displays it).

For struct definitions where there's a lot of comments between elements, I'm not sure if there's value in trying to align everything. Maybe we want to "copy" (adapt) our internal "C" CODING_STYLE to this project.

lib/libmuser.c
tmakatos: This struct is internal to the library, so there's no problem of namespace clash. Do we want to prefix internal structs as well?

lib/libmuser.c
tmakatos: Let's leave it for later, it makes no difference.

lib/libmuser.c
tmakatos: I wanted to do it as you suggest however I don't know how to statically assert that a member of a struct is of a specific size.

lib/libmuser.c
tmakatos: Ditto.

lib/muser.h
tmakatos: OK will remove alignment for those.

tmakatos:
@franciozzy I've addressed your comments, please take another look.

tmakatos:
Superseded by https://github.com/nutanix/libvfio-user/issues/106.

PR #22, opened 2019-12-04
Cleanup 3

franciozzy:
Various changes in preparation for Larmier unit tests, including some fixes for bugs found during testing.


lib/dma.c
```diff
@@ -137,6 +137,10 @@ dma_controller_remove_regions(lm_ctx_t *ctx, dma_controller_t *dma)
 void
 dma_controller_destroy(lm_ctx_t *lm_ctx, dma_controller_t *dma)
 {
+    if (dma == NULL) {
```
swapnili: I think the caller should not pass NULL *dma. It looks better to fix the caller and maybe have a assert() here.

PR #26, opened 2019-12-17
Mmap fixes

tmakatos:


kmod/muser.c
```diff
@@ -1267,7 +1267,10 @@ static int muser_mmap(struct mdev_device *const mdev,
 
 	/* Process mudev_cmd in server context. */
 	err = muser_process_cmd(mudev, &mucmd);
-	if (unlikely(err)) {
+	if (likely(err == 0)) {
```
swapnili: As in case of error err will be <0, I think we can avoid the extra "if" by doing following:
if (unlikely(err< 0)) {
  muser_info("failed to mmap: %d", err);
  return err;
}
err = mucmd.muser_cmd.err;

tmakatos: `muser_process_cmd` only returns an error if `down_timeout` fails, so this check is needed.

tmakatos: Ignore the above, the reason it's done this way is so that we can have a single point of printing the error message.

swapnili: Sorry I did not got how single point of printing the error message is related to this change. If the error is there then we print the error message. down_timeout() can return -ETIME. Even with following same behaviour should be there,
if (unlikely(err< 0)) {
muser_info("failed to mmap: %d", err);
return err;
}
err = mucmd.muser_cmd.err;

tmakatos: It's not 100% related to this change, should have been a different patch, but it's a good thing to have anyway because it makes identifying failures much easier. Otherwise we'd have to correlate the -ETIME failure with the mmap failure.

swapnili: Ok got it, I misunderstood it.

Issue #34, opened 2020-01-27
kernel warning if muser app dies while guest is doing PCI read


tmakatos:
Because of an unrelated bug, the process providing device emulation died (specifically it's muser SPDK, https://github.com/tmakatos/spdk):
```
[    5.793367] ACPI: PCI Interrupt Link [LNKD] enabled at IRQ 10
qemu-system-x86_64: vfio_region_write(00000000-0000-0000-0000-000000000000:region0+0x14, 0x0,4) failed: Operation not permitted
qemu-system-x86_64: vfio_region_read(00000000-0000-0000-0000-000000000000:region0+0x1c, 4) failed: Timer expired
[   10.837434] nvme nvme0: Removing after probe failure status: -19
qemu-system-x86_64: vfio_region_read(00000000-0000-0000-0000-000000000000:region0+0x1c, 4) failed: Timer expired
qemu-system-x86_64: vfio_pci_write_config(00000000-0000-0000-0000-000000000000, 0x4a, 0x3, 0x2) failed: Operation not permitted
qemu-system-x86_64: vfio_pci_read_config(00000000-0000-0000-0000-000000000000, 0x3d, 0x1) failed: Timer expired
qemu-system-x86_64: vfio 00000000-0000-0000-0000-000000000000: Failed to set up TRIGGER eventfd signaling for interrupt INTX-0: VFIO_DEVICE_SET_IRQS failure: Operation not permitted
qemu-system-x86_64: vfio_pci_read_config(00000000-0000-0000-0000-000000000000, 0x4, 0x2) failed: Timer expired
qemu-system-x86_64: vfio_pci_write_config(00000000-0000-0000-0000-000000000000, 0x4, 0xfbc2, 0x2) failed: Operation not permitted
qemu-system-x86_64: vfio_pci_read_config(00000000-0000-0000-0000-000000000000, 0x5c, 0x2) failed: Timer expired
qemu-system-x86_64: vfio_pci_write_config(00000000-0000-0000-0000-000000000000, 0x5c, 0xffc0, 0x2) failed: Operation not permitted
qemu-system-x86_64: vfio_pci_read_config(00000000-0000-0000-0000-000000000000, 0x4, 0x2) failed: Timer expired
qemu-system-x86_64: vfio_pci_read_config(00000000-0000-0000-0000-000000000000, 0x0, 0x4) failed: Timer expired
```
The host kernel:
```
[...]

PR #35, opened 2020-01-28
Fix for #34 issue

swapnili:


swapnili:
We do not need this with libmuser_open / libmuser_release. The module reference is incremented/decremented when the device emulation open()/close() the mudev control device.

PR #37, opened 2020-02-04
use vmalloc for allocating array of pages

### LINKED ISSUE #30: DMA notifier fails if using hugepages and  GB of guest memory
When using hugepages and 3 GB of guest memory, QEMU fails as follows:
```
qemu-system-x86_64: -device vfio-pci,sysfsdev=/sys/bus/mdev/devices/00000000-0000-0000-0000-000000000000: vfio 00000000-0000-0000-0000-000000000000: error getting device from group 0: Inappropriate ioctl for device
Verify all devices in group 0 are bound to vfio-<bus> or pci-stub and not already in use
```
The kernel emits the following warning:
```
[ 2484.634813] notifier failed for iova=100000 vaddr=7f50a3700000 size=bff00000
```

As kun.qiu@intel.com mentioned this might be because of the `kmalloc` in `muser_iommu_dma_map`. If this is the case then it should be trivial to fix e.g. by using `vmalloc`.

tmakatos:
Otherwise kmalloc can fail to allocate memory for larger guest VM RAM (>= 3GB).

fixes #30 

### DIFF
--- a/kmod/muser.c
+++ b/kmod/muser.c
@@ -639,7 +639,7 @@ static void put_dma_map(struct muser_dev *mudev,
 		put_page(dma_map->pages[i]);
 	}
 
-	kfree(dma_map->pages);
+	vfree(dma_map->pages);
 }
 
 static int
@@ -654,7 +654,7 @@ get_dma_map(struct muser_dev *mudev, struct vfio_dma_mapping *dma_map,
 	struct page **pages;
 
 	length = map->size;
-	pages = kmalloc_array(NR_PAGES(length), sizeof(*pages), GFP_KERNEL);
+	pages = vmalloc(NR_PAGES(length) * sizeof(*pages));
 	if (!pages)
 		return -ENOMEM;
 


kmod/muser.c
```diff
@@ -654,7 +654,7 @@ get_dma_map(struct muser_dev *mudev, struct vfio_dma_mapping *dma_map,
 	struct page **pages;
 
 	length = map->size;
-	pages = kmalloc_array(NR_PAGES(length), sizeof(*pages), GFP_KERNEL);
+	pages = vmalloc(NR_PAGES(length) * sizeof(*pages));
```
swapnili: I think using kvmalloc() or kvmalloc_array should be better, it first tries to allocate physically contiguous pages otherwise returns virtually contiguous pages.

tmakatos: Why do we care if pages contiguous?

swapnili: kmalloc() is faster than vmalloc() and also the accesses to the kmalloc'ed memory is faster as additional translation is not required. So mostly in the non-huge page cases we can leverage the kmalloc'ed memory.
I do not have a strong opinion on this, its just a suggestion.

tmakatos: Fair enough, I'll create an issue ti improve this in the future.

PR #42, opened 2020-02-24
support for huge pages

### LINKED ISSUE #29: bad page map kernel warning when QEMU guest uses hugepages for guest memory
When using hugepages for the guest memory:
```
-object memory-backend-file,id=mem,size=1G,mem-path=/dev/hugepages,share=on -mem-prealloc -numa node,memdev=mem
```
the kernel (5.4.0-rc8) emits the following warning multiple times:
```
[2065774.775141] BUG: Bad page map in process gpio-pci-idio-1  pte:80000001bcea0027 pmd:22f4c4067
[2065774.775980] page:ffffddd2c6f3a800 refcount:0 mapcount:-1 mapping:dead000000000400 index:0xa0 compound_mapcount: -1
[2065774.776641] hugetlbfs_aops name:"qemu_back_mem.mem.NGh5A4"
[2065774.777329] flags: 0x2ffff8000000000()
[2065774.778116] raw: 02ffff8000000000 ffffddd2c6f38001 ffffddd0c6f3a808 dead000000000400
[2065774.778733] raw: 0000000000000000 0000000000000000 0000000000000000 0000000000000000
[2065774.779350] page dumped because: bad pte
[2065774.779965] addr:00007fb15f4f4000 vm_flags:1c0000fb anon_vma:0000000000000000 mapping:ffff8c0329d59570 index:a0
[2065774.780608] file:00000000-0000-0000-0000-000000000000 fault:0x0 mmap:libmuser_mmap [muser] readpage:0x0
[2065774.781307] CPU: 4 PID: 96138 Comm: gpio-pci-idio-1 Tainted: G           O      5.4.0-rc8+ #11
[2065774.782361] Hardware name: Nutanix AHV, BIOS 1.9.1-5.el6 04/01/2014
[2065774.783352] Call Trace:
[2065774.784611]  dump_stack+0x66/0x8b
[2065774.785866]  print_bad_pte+0x1d1/0x2a0
[2065774.786652]  unmap_page_range+0x7b6/0xab0
[2065774.787410]  ? __switch_to_asm+0x40/0x70
[2065774.788114]  unmap_vmas+0x81/0xf0
[2065774.788811]  unmap_region+0xae/0x120
[2065774.789501]  __do_munmap+0x2aa/0x4d0
[2065774.790190]  __vm_munmap+0x6f/0xc0                                                                                                                                                                            [2065774.790868]  __x64_sys_munmap+0x27/0x30
[2065774.791531]  do_syscall_64+0x52/0x180
[2065774.792184]  entry_SYSCALL_64_after_hwframe+0x44/0xa9
[2065774.792828] RIP: 0033:0x7fb19f5386e7
[2065774.793476] Code: c7 c0 ff ff ff ff eb 8d 48 8b 15 ac 47 2b 00 f7 d8 64 89 02 e9 5b ff ff ff 66 2e 0f 1f 84 00 00 00 00 00 b8 0b 00 00 00 0f 05 <48> 3d 01 f0 ff ff 73 01 c3 48 8b 0d 81 47 2b 00 f7 d8 64 89
01 48
[2065774.794832] RSP: 002b:00007ffdc6f671d8 EFLAGS: 00000206 ORIG_RAX: 000000000000000b
[2065774.795524] RAX: ffffffffffffffda RBX: 0000000000000000 RCX: 00007fb19f5386e7
[2065774.796232] RDX: 0000000040000000 RSI: 0000000040000000 RDI: 00007fb15f454000
[2065774.796934] RBP: 00007ffdc6f67200 R08: 0000000000000000 R09: 0000000000000000
[2065774.797628] R10: 0000000000000001 R11: 0000000000000206 R12: 000055e3b81fb830
[2065774.798324] R13: 00007ffdc6f67a40 R14: 0000000000000000 R15: 0000000000000000
[2065774.799088] Disabling lock debugging due to kernel taint
```
reported by kun.qiu@intel.com

Surprisingly the GPIO sample seems to work, although attempting to unload muser.ko results in much more serious errors.

The problem might be that `libmuser` calls `mmap` without the `MAP_HUGE` flags in this case.

tmakatos:
This PR fixes #29 by injecting the fd backing the VMA in the DMA registration and then requests libmuser to mmap this fd. Apart from allowing huge pages to work in muser, it also avoids having to pin guest pages.

Because we no longer call `vfio_pin_pages`, VFIO no longer sends us the DMA unregistration event, we need to fix this (the problem is that we get overlapping/duplicates DMA regions).

Also, we might be injecting multiple times the same fd, so we might want to optimize that.

This PR also includes some other minor improvements.

### DIFF
--- a/kmod/CMakeLists.txt
+++ b/kmod/CMakeLists.txt
@@ -33,7 +33,11 @@
 configure_file(muser.c ${CMAKE_CURRENT_BINARY_DIR}/muser.c COPYONLY)
 configure_file(muser.h ${CMAKE_CURRENT_BINARY_DIR}/muser.h COPYONLY)
 # FIXME need to pass "CFLAGS_muser.o := -DDEBUG" for debug builds
-file(WRITE ${CMAKE_CURRENT_BINARY_DIR}/Kbuild "obj-m := muser.o")
+set(KMOD_MAKEFILE_CONTENT "obj-m := muser.o")
+IF(CMAKE_BUILD_TYPE MATCHES Debug)
+	set(KMOD_MAKEFILE_CONTENT "CFLAGS_muser.o := -DDEBUG\n${KMOD_MAKEFILE_CONTENT}")
+ENDIF(CMAKE_BUILD_TYPE MATCHES Debug)
+file(WRITE ${CMAKE_CURRENT_BINARY_DIR}/Kbuild ${KMOD_MAKEFILE_CONTENT})
 
 # Build module using kernel's Makefile.
 set(KBUILD_CMD ${CMAKE_MAKE_PROGRAM} -C ${KDIR} M=${CMAKE_CURRENT_BINARY_DIR} modules)
--- a/kmod/muser.c
+++ b/kmod/muser.c
@@ -1279,9 +1279,9 @@ static int muser_mmap(struct mdev_device *const mdev,
 		err = mucmd.muser_cmd.err;
 	}
 	if (unlikely(err != 0)) {
-		muser_info("failed to mmap %#lx@%#lx: %d",
-		           mucmd.muser_cmd.mmap.request.len,
+		muser_info("failed to mmap %#lx-%#lx: %d",
 		           mucmd.muser_cmd.mmap.request.addr,
+		           mucmd.muser_cmd.mmap.request.addr + mucmd.muser_cmd.mmap.request.len,
 		           err);
 		return err;
 	}
--- a/kmod/muser.c
+++ b/kmod/muser.c
@@ -1019,7 +1019,7 @@ static ssize_t muser_read(struct mdev_device *mdev, char __user *buf,
 		_count = mucmd.muser_cmd.err;
 
 	if (_count < 0)
-		muser_dbg("failed to process read: %d, %d\n", err,
+		muser_dbg("failed to process read: %d, %d", err,
 		          mucmd.muser_cmd.err);
 
 	*ppos = mucmd.muser_cmd.rw.pos;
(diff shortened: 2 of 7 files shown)


tmakatos:
> I think it is important to vfio_pin_pages() which are behind the GPA, we cannot get rid of it as VFIO relies on that. After this change we will no longer be pinning those pages and then they can be swapped out anytime in between.

We do want the pages to be able to be swapped out, that was one of the key points raised during the talk. If we pin the pages then there's no memory overcommit etc.

swapnili:
I think the point was not to pin each dma regions, instead pin only when required. Without pinning the pages there is no guarantee that page will not be moved/swapped-out during the bi-directional IO to device. As per my understanding the DMA'ble area should always be pinned in memory.

tmakatos:
> I think the point was not to pin each dma regions, instead pin only when required. Without pinning the pages there is no guarantee that page will not be moved/swapped-out during the bi-directional IO to device. As per my understanding the DMA'ble area should always be pinned in memory.

The "device" in this case is libmuser, so if the page is swapped out and libmuser attempts to touch it then it will be swapped in. Are you referring to the scenario where libmuser has to send this request to an actual device?

swapnili:
Yes in case of real device. But now am also thinking do we need to obey those semantics even in case of virtual device?

tmakatos:
> Yes in case of real device. But now am also thinking do we need to obey those semantics even in case of virtual device?

Even if libmuser uses this memory for a real device, the kernel will pin that memory because it has been properly set up. So pinning is completely unnecessary in both cases.

PR #43, opened 2020-02-25
a few additional fixes/improvements regarding the recent hugepages fix

tmakatos:


kmod/muser.c
```diff
@@ -708,16 +715,16 @@ static int muser_iommu_dma_unmap(struct muser_dev *const mudev,
 	dma_map = __find_dma_map(mudev, unmap->iova);
 	if (!dma_map) {
 		mutex_unlock(&mudev->dev_lock);
-		muser_dbg("failed to find dma map for iova:%llu\n", unmap->iova);
-		return -EINVAL;
+		return 0;
```
swapnili: Why this error is ignored now?

tmakatos: Previously we used `vfio_pin_pages`, so VFIO would only send DMA unmap events for DMA regions we have pinned. Since we no longer use `vfio_pin_pages`, VFIO doesn't send a DMA unmap event at all. This is a problem for which we haven't got a solution yet, but one possible solution would be to force VFIO to send the DMA unmap event at least once, independently of whether or not the DMA region is mapped (more on this here https://www.redhat.com/archives/vfio-users/2020-February/msg00016.html). So in my setup I'm testing using this hack in VFIO, and since we don't send to libmuser every DMA region VFIO sends us (some regions are not shareable), in the end VFIO will tell us to unmap a DMA  region we haven't mapped.

swapnili: This does not looks correct to me, looks like a hack to ignore the error cases which may be a valid bug(dma unmap sent for an invalid map).
I would suggest write a big comment over here explaining why are we doing this and needs to be fixed later.

tmakatos: Sure I can put a comment clarifying why we do this, but if the proposed solution is accepted in VFIO then it's not a hack.

swapnili: If it is accepted then the comment can be removed later. But better to have it now.

tmakatos: Done.

PR #46, opened 2020-02-26
don't install the same fd multiple times plus refactor

tmakatos:
This patch avoids installing the same fd multiple times, and this saves system resources. The DMA library still calls dup(2) on the fd, so we might want to remove that and completely avoid having fd duplicates.

This patch also refactors libmuser_unl_ioctl, essentially it moves the main two cases into separate functions.


tmakatos:
Closing for now as we've decided to install the fd every time a DMA region is added without checking whether it has already been installed.

Issue #44, opened 2020-02-25
don't install the same fd multiple times


tmakatos:
During DMA map we grad the fd and blindly install it in libmuser, however we might have already installed an fd for that file. We should keep track of that and avoid installed a new fd for the same file multiple times.

jlevon:
What's the status of this now?

PR #49, opened 2020-02-27
don't duplicate DMA region fd when adding DMA region, plus minor improvements

tmakatos:


tmakatos:
This relies on #43

tmakatos:
> Looks like some changes in this patch are also in https://github.com/nutanix/muser/pull/43/files

Yes, that why I need #43 to be reviewed first.

tmakatos:
@swapnili I've rebased this PR on top of master so the redundant commits have been removed, you can review it now.

lib/dma.c
```diff
@@ -207,13 +215,7 @@ dma_controller_add_region(lm_ctx_t *lm_ctx, dma_controller_t *dma,
     region->size = size;
     region->page_size = page_size;
     region->offset = offset;
-
-    region->fd = dup(fd);       // dup the fd to get our own private copy
```
swapnili: This does not looks correct. In dma_controller_add_region() it holds reference on fd to call mmap() and in  dma_controller_remove_region() it munmap() and leaves the fd reference.
Now with this patch the fd reference is assumed to be held by caller and still the reference is left in   dma_controller_remove_region().
API wise this does not look correct. I think its better to keep these api's same and  the caller should take care of close(fd) if he has additional reference on the file.

kmod/muser.c
```diff
@@ -1446,113 +1446,118 @@ static int fd_install_dma_map(struct file *file)
...
+	/* TODO: These WARN_ON()s should really just detach mudev. */
 
-		if (mucmd->muser_cmd.type == MUSER_DMA_MMAP) {
-			ret = fd_install_dma_map(mucmd->muser_cmd.mmap.request.file);
-			if (ret < 0) {
-				return ret;
-			}
-			mudev->dma_map->fd = mucmd->muser_cmd.mmap.request.fd = ret;
-		}
+	if (mucmd->muser_cmd.type == MUSER_DMA_MMAP) {
```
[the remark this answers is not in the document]
tmakatos: I've thought of that, too. But then I realized that `maybe_install_fds` is a bit misleading and was planning to refactor it by taking the `if` out of it and changing it to `install_irq_fds`.

lib/dma.c
tmakatos: I agree with you, that's how I originally wanted to implement it but it turns out to be much more complicated in the kernel, so I've left it as a future optimization (https://github.com/nutanix/muser/issues/44). With this patch, the API changes as follows: `dma_controller_add_region`  operates with an fd as if it had opened it (it's installed by muser.ko) and it's responsible for closing it in the end.

kmod/muser.c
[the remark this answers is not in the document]
swapnili: Better to have one function installing fd's, instead of having it at multiple places.

lib/dma.c
swapnili: I do not see any reference to the above mentioned point in the issue (#44). Please update the issue accordingly, otherwise it will be forgotten.

tmakatos: > I do not see any reference to the above mentioned point in the issue (#44). Please update the issue accordingly, otherwise it will be forgotten.

I've update the description in #44.

PR #52, opened 2020-06-12
merge vfio-user into master

tmakatos:


lib/libmuser.c
```diff
@@ -1053,19 +1053,6 @@ lm_irq_trigger(lm_ctx_t *lm_ctx, uint32_t vector)
         return -1;
     }
 
-    if (vector == LM_DEV_INTX_IRQ_IDX && !lm_ctx->pci_config_space->hdr.cmd.id) {
```
franciozzy, line 1056: I think this logic is wrong, and that's why you were having issues.
It's saying: "If (vfio) vector is INTX and ID==0, then return -1". I think that's wrong.
I believe it should be: "if (vfio) vector is INTX and ID==1, then return -1".

Note that this "vfio vector" is a mechanism to indicate which IRQ type to use; it's not the IRQ VECTOR index. In VFIO parlance, that's the "subindex" indicated by "start" and "count" in VFIO_IRQ_SET.

tmakatos: Felipe is right, the original code was wrong and the right way to fix it is to remove the negation from `!lm_ctx->pci_config_space->hdr.cmd.id`.

jlevon (approved):
Yup, it's time :)

PR #79, opened 2020-11-12
Fix compilation warnings  

swapnili:


lib/muser_priv.h
```diff
@@ -35,6 +35,8 @@
 
 #include "muser.h"
 
+#define UNUSED __attribute__((unused))
```
tmakatos, line 38: Why not just drop this define?

samples/client.c
```diff
@@ -548,7 +548,8 @@ get_dirty_bitmaps(int sock, struct vfio_user_dma_region *dma_regions,
 {
     struct vfio_iommu_type1_dirty_bitmap dirty_bitmap = {0};
     struct vfio_iommu_type1_dirty_bitmap_get bitmaps[2];
-    int ret, i;
+    int ret;
+	size_t i;
```
tmakatos, line 552: identation

tmakatos (commented):
couple of minor comments, good to merge anyway

lib/muser_priv.h
swapnili: This is the define which is used for unused variables everywhere internally.

samples/client.c
swapnili: Will fix it.

lib/muser_priv.h
tmakatos: I know, why not use directly `__attribute__((unused))`?

swapnili: This way it looks better and readable to me. Also using __attribute__ all over needs extra indentation.

tmakatos (approved):

jlevon (approved):

PR #84, opened 2020-11-16
use object libraries to build libmuser

jlevon:
Using target_link_libraries for libmuser meant that unused symbols - including
all of the public API - were not brought into the target shared object. Use
CMake's "object library" feature instead, which doesn't use intermediate
archives and thus avoids this issue.


franciozzy (approved):
LGTM

tmakatos (approved):
I don't know much about this but I cloned the PR and can see the missing symbols in `nm libmuser.so`, so it looks right.

PR #85, opened 2020-11-16
Makefile: Remove unused KDIR

swapnili:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,17 +1,5 @@
 language: c
 dist: bionic
 compiler: gcc
-before_install:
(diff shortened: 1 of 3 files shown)


jlevon:
There is still a lingering reference in .travis.yml - presumably we don't need the linux checkout at all there?

swapnili:
Yes I saw that in .travis.yml, but was not sure if we need it. I thought it should be installing latest include/vfio.h, but that is not the case as it just builds the kernel. I will remove that.
Thanks

jlevon (approved):

PR #88, opened 2020-11-16
eliminate duplicate recv(2) code

tmakatos:
Instead of having each command handler do its own recv(2) to receive the
command, we do this once for all commands. This reduces the code a bit.

### DIFF
--- a/lib/muser_ctx.c
+++ b/lib/muser_ctx.c
@@ -831,17 +831,6 @@ dev_set_irqs_validate(lm_ctx_t *lm_ctx, struct vfio_irq_set *irq_set)
     return 0;
 }
 
-static int
-device_reset(lm_ctx_t *lm_ctx)
-{
-    lm_log(lm_ctx, LM_DBG, "Device reset called by client");
-    if (lm_ctx->reset != NULL) {
-        return lm_ctx->reset(lm_ctx->pvt);
-    }
-
-    return 0;
-}
-
 static long
 dev_set_irqs(lm_ctx_t *lm_ctx, struct vfio_irq_set *irq_set, void *data)
 {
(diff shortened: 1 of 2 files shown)


lib/muser_ctx.c
```diff
@@ -2181,6 +2037,25 @@ process_request(lm_ctx_t *lm_ctx)
...
     }
 
+    /*
+     * TODO from now on if an error occurs we still need to reply. Move this
+     * code into a separate function so that we don't have to use goto.
+     */
+
+    hdr.msg_size -= sizeof(hdr);
+    if (hdr.msg_size > 0) {
+        cmd_data = malloc(hdr.msg_size);
```
jlevon, line 2047: no need to fix this now, but you reminded me to file #89

lib/muser_ctx.c
```diff
@@ -2181,6 +2037,25 @@ process_request(lm_ctx_t *lm_ctx)
...
+        }
+        ret = recv(lm_ctx->conn_fd, cmd_data, hdr.msg_size, 0);
+        if (ret < 0) {
+            ret = -errno;
+            goto reply;
+        }
+    }
+
     /* FIXME in most of the following function we check that hdr.count is >=
      * than the command-specific struct and there is an additional recv(2) for
```
jlevon, line 2060: want to update the recv part of this?

jlevon: and does it even make sense try to move the size checks here themselves?

jlevon (approved):

lib/muser_ctx.c
tmakatos: > want to update the recv part of this?

Not sure I understand?

> and does it even make sense try to move the size checks here themselves?

Maybe, I'm tried to understand how we can further simplify code, I need to think about it a bit more and then come up with a suggestion.

jlevon: The FIXME is talking about the additional recv(2), but aren't these changes implementing exactly that?

tmakatos: Correct, I removed the FIXME comment in a new PR.

PR #91, opened 2020-11-16
don't assert on error and avoid having to send SIGUSR1 in client/server sample

tmakatos:


tmakatos:
This has been already merge via another PR.

PR #97, opened 2020-11-17
refactor socket code into lib/tran_sock.[ch]

jlevon:
While it's handy for the samples to reach in and re-use private libmuser code,
it's better if we can reduce the dependencies such that it only re-uses the
socket transport code.

### DIFF
--- a/lib/tran_sock.c
+++ b/lib/tran_sock.c
@@ -125,9 +125,9 @@ out:
 
 int
 _send_vfio_user_msg(int sock, uint16_t msg_id, bool is_reply,
-                   enum vfio_user_command cmd,
-                   struct iovec *iovecs, size_t nr_iovecs,
-                   int *fds, int count, int err)
+                    enum vfio_user_command cmd,
+                    struct iovec *iovecs, size_t nr_iovecs,
+                    int *fds, int count, int err)
 {
     int ret;
     struct vfio_user_header hdr = {.msg_id = msg_id};
@@ -395,7 +395,10 @@ recv_version(lm_ctx_t *lm_ctx, int sock, uint16_t *msg_idp,
     }
 #endif
 
-    // FIXME: incorrect, unspecified means "no migration support", handle this
+    /*
+     * FIXME: incorrect, if the client doesn't give a pgsize value, it means "no
+     * migration support", handle this
+     */
     lm_ctx->migration.pgsize = sysconf(_SC_PAGESIZE);
     lm_ctx->client_max_fds = 1;
 
--- a/lib/tran_sock.c
+++ b/lib/tran_sock.c
@@ -1,0 +1,572 @@
+/*
+ * Copyright (c) 2020 Nutanix Inc. All rights reserved.
+ *
+ * Authors: Thanos Makatos <thanos@nutanix.com>
+ *          Swapnil Ingle <swapnil.ingle@nutanix.com>
+ *          Felipe Franciosi <felipe@nutanix.com>
+ *
+ *  Redistribution and use in source and binary forms, with or without
+ *  modification, are permitted provided that the following conditions are met:
+ *      * Redistributions of source code must retain the above copyright
+ *        notice, this list of conditions and the following disclaimer.
+ *      * Redistributions in binary form must reproduce the above copyright
+ *        notice, this list of conditions and the following disclaimer in the
+ *        documentation and/or other materials provided with the distribution.
+ *      * Neither the name of Nutanix nor the names of its contributors may be
+ *        used to endorse or promote products derived from this software without
+ *        specific prior written permission.
+ *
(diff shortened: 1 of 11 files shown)


lib/tran_sock.c
```diff
@@ -0,0 +1,564 @@
...
+    ret = recv_vfio_user_msg(sock, hdr, is_reply, msg_id, NULL, NULL);
+
+    if (ret != 0) {
+        return ret;
+    }
+
+    assert(hdr->msg_size >= sizeof (*hdr));
+
+    len = hdr->msg_size - sizeof (*hdr);
+
```
swapnili, line 282: ditto

lib/tran_sock.c
```diff
@@ -0,0 +1,564 @@
...
+    if (len == 0) {
+        *datap = NULL;
+        *lenp = 0;
+        return 0;
+    }
+
+    data = calloc(1, len);
+
```
swapnili, line 290: ditto

lib/tran_sock.c
```diff
@@ -0,0 +1,564 @@
...
+    struct vfio_user_version *cversion;
+    struct vfio_user_header hdr;
+    size_t vlen;
+    int ret;
+
+    *versionp = NULL;
+
+    ret = recv_vfio_user_msg_alloc(sock, &hdr, false, msg_idp,
+                                   (void **)&cversion, &vlen);
+
```
swapnili, line 360: No need for empty line.

lib/tran_sock.c
```diff
@@ -0,0 +1,564 @@
...
+#if 0
+    if (cversion->major != LIB_MUSER_VFIO_USER_VERS_MJ) {
+        lm_log(lm_ctx, LM_ERR, "unsupported client major %hu (must be %hu)",
+               cversion->major, LIB_MUSER_VFIO_USER_VERS_MJ);
+        ret = -ENOTSUP;
+        goto out;
+    }
+#endif
+
+    // FIXME: incorrect, unspecified means "no migration support", handle this
```
swapnili, line 390: Can you please have elaborated description here. Now I see that we always specify pgsize and client_max_fds. Do you mean when we start parsing the JSON and client do not specifies it?

samples/client.c
```diff
@@ -41,9 +42,27 @@
...
+static char *irq_to_str[] = {
+    [LM_DEV_INTX_IRQ_IDX] = "INTx",
+    [LM_DEV_MSI_IRQ_IDX] = "MSI",
+    [LM_DEV_MSIX_IRQ_IDX] = "MSI-X",
+    [LM_DEV_ERR_IRQ_INDEX] = "ERR",
+    [LM_DEV_REQ_IRQ_INDEX] = "REQ"
+};
+
+void
+lm_log(UNUSED lm_ctx_t *lm_ctx, UNUSED lm_log_lvl_t lvl, const char *fmt, ...)
```
swapnili, line 58: Unused function?

lib/tran_sock.c
```diff
@@ -0,0 +1,564 @@
...
+    if (ret != 0) {
+        close(unix_sock);
+        return ret;
+    }
+    return unix_sock;
+}
+
+int
+_send_vfio_user_msg(int sock, uint16_t msg_id, bool is_reply,
+                   enum vfio_user_command cmd,
```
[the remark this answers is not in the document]
jlevon: aye

lib/tran_sock.c
```diff
@@ -0,0 +1,564 @@
...
+int
+recv_vfio_user_msg_alloc(int sock, struct vfio_user_header *hdr, bool is_reply,
+                         uint16_t *msg_id, void **datap, size_t *lenp)
+{
+    void *data;
+    size_t len;
+    int ret;
+
+    ret = recv_vfio_user_msg(sock, hdr, is_reply, msg_id, NULL, NULL);
+
```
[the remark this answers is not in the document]
jlevon: I personally prefer the visual look here, and I don't think we have a style that says otherwise, so I'm oging to leave these as is.

lib/tran_sock.c
jlevon: I did. Updating the comment.

samples/client.c
jlevon: It's called from tran_sock.c code, so we need it here.

swapnili: There is also another definition of lm_log() in lib/muser_ctx.c
As tran_sock.c is part of library, it should be using the functions defined in library.
sample/* are test examples, we should not be defining functions there and using it in library.

jlevon: tran_sock.c is used by both samples/client.c - which does not and should not link against libmuser - and libmuser.

In the first case, tran_sock.c's lm_log links with the one in samples/client.c
In the second case, tran_sock.c's lm_log links with the one in lib/muser_ctx.c

Make sense?

swapnili: Ok. Thanks got it.

swapnili (approved):

PR #98, opened 2020-11-17
enable assert() in release builds

jlevon:
Also fix release build failures, and build in release mode in travis.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -32,6 +32,16 @@ cmake_minimum_required (VERSION 2.6)
 project(muser)
 include(GNUInstallDirs)
 
+set(CMAKE_C_FLAGS_DEBUG "-O0 -ggdb")
+#
+# NB: this intentionally does not define NDEBUG: we want our assert()s even in a
+# release build.
+#
+set(CMAKE_C_FLAGS_RELEASE "-O3")
+
+set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -Wall -Werror -Wextra")
+
+
 # shared libraries
 add_subdirectory(lib)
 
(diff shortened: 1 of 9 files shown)


CMakeLists.txt
```diff
@@ -32,9 +32,6 @@ cmake_minimum_required (VERSION 2.6)
 project(muser)
 include(GNUInstallDirs)
 
-
-set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -Wall -Werror -Wextra")
```
swapnili, line 36: Why to remove this? We do want to catch the compiler warnings all over the code.

lib/CMakeLists.txt
```diff
@@ -30,6 +30,11 @@
 
 set(CMAKE_C_FLAGS "-Wall -Wextra -Werror -fPIC")
 set(CMAKE_C_FLAGS_DEBUG "-O0 -ggdb")
+#
+# NB: this intentionally does not define NDEBUG: we want our assert()s even in a
+# release build.
+#
+set(CMAKE_C_FLAGS_RELEASE "-O3")
```
swapnili, line 37: I think NDEBUG should be undefined from top level CMakeLists.txt. So that this rule is applicable all over the repo.

swapnili (changes requested):

CMakeLists.txt
jlevon: Yeah, this was a mistake; I've reworked how we set the flags in both places.

lib/CMakeLists.txt
jlevon: Thanks, yes!

swapnili (approved):
The travis build failed. Looks like you will need to rebase on master.

jlevon:
> The travis build failed. Looks like you will need to rebase on master.

Thanks, yes, I'll need to rebase the series.

swapnili (approved):

PR #100, opened 2020-11-18
live migration: support for resuming in the destination server plus samples

tmakatos:
This patch implements support for resuming on the destination server while live migrating according to the VFIO live migration protocol. I've also extended the client/server sample to show how it's done, in a rather simplistic way.

The server is now a device that can be programmed to trigger an interrupt at a specific time. This is done by writing the desired seconds since Epoch to BAR0. We try to make the sample non-trivial by programming the device to trigger an interrupt after two seconds from now, immediately migrating to the new server, and then receiving the interrupt from the new server.

The client/server sample is started as usual. We don't create the destination client, the current one is reused (this would be done by libvirt, there's isn't much value in demonstrating how to do this anyway). However we do read the migration state from the source server and write it to the destination server. The destination server is started by the client (this would also be done by libvirt).

Some assorted fixes/improves are also included.

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,3 +57,34 @@ To enable Python bindings set the PYTHON_BINDINGS environment variable to a
 non-empty string.
 
 Finally build your program and link it to libmuser.so.
+
+Example
+=======
+
+Directory samples/ contains a client/server implementation. The server
+implements a device that can be programmed to trigger interrupts (INTx) to the
+client. This is done by writing the desired time in seconds since Epoch. The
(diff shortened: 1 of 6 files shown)


jlevon (approved):
Not going to pretend I looked at every line, but overall looks good.
One thing: would be nice to have a comment somewhere describing what to run to test this.

tmakatos:
> Not going to pretend I looked at every line, but overall looks good.
> One thing: would be nice to have a comment somewhere describing what to run to test this.

Done!

PR #104, opened 2020-11-19
move live migration into separate file

tmakatos:


jlevon (approved):

jlevon:
Looks fine presuming you fix the release build :)

tmakatos:
> Looks fine presuming you fix the release build :)

Are you referring to the Travis build failure?

jlevon:
yes

tmakatos:
> yes

I don't get this error with GCC 8.3.0. I believe that this is a false positive: `negotiate` calls `recv_version` with `msg_id` uninitialized, `recv_version` in turns calls `recv_vfio_user_msg_alloc` with `is_reply=false`, which then calls `recv_vfio_user_msg`. Since `is_reply == false`, `msg_id` is only written to.

I can trivially silence this with:
```
@@ -467,7 +467,7 @@ static int
 negotiate(lm_ctx_t *lm_ctx, int sock)
 {
     struct vfio_user_version *client_version = NULL;
-    uint16_t msg_id;
+    uint16_t msg_id = 0;
     int ret;
```
But is this the right thing to do? Can we update GCC in the CI?

jlevon:
The -Wmaybe-uninitialized implementation in older GCC's is very buggy. However, it does catch real bugs sometimes. So we don't want to disable the option if we can avoid it, but we also want to be able to compile everywhere we can.

So, let's do the workaround of initing to zero even though a smarter compiler can figure it out.

tmakatos:
Fixed, waiting for Travis build.

PR #109, opened 2020-11-20
don't expose live migration as device region

tmakatos:


swapnili (approved):
I haven't gone through the whole change. But let's get this in as I need to rebase API changes on top of this.

PR #113, opened 2020-11-23
parse VFIO_USER_VERSION JSON stanza

jlevon:
This also fixes up devices w/o migration (like the gpio sample)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,8 @@
 language: c
 dist: bionic
 compiler: gcc
+before_install:
+  - sudo apt-get -y install libjson-c-dev
 script:
   - make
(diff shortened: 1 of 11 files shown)


tmakatos (commented):
I think we can avoid passing the whole `dev_info` struct and simply check for `migr->size != 0` intead of the existing (and erroneous) `migr != NULL`.

jlevon:
Do you mean in pci_config_setup()? We use dev_info there aside from checking ->migration

tmakatos:
> Do you mean in pci_config_setup()? We use dev_info there aside from checking ->migration

Ah yes, you're right.

tmakatos (approved):

PR #117, opened 2020-11-24
clean up headers

jlevon:
Introduce include/ dir for public headers, and clean up include guards etc.

### DIFF
--- a/include/muser.h
+++ b/include/muser.h
@@ -37,12 +37,16 @@
 #include <sys/uio.h>
 #include <unistd.h>
 
-#include "vfio_user.h"
 #include "pci.h"
-#include "caps/pm.h"
-#include "caps/px.h"
-#include "caps/msi.h"
-#include "caps/msix.h"
+#include "pci_caps/pm.h"
+#include "pci_caps/px.h"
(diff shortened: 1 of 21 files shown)


include/muser.h
```diff
@@ -37,12 +37,16 @@
 #include <sys/uio.h>
 #include <unistd.h>
 
-#include "vfio_user.h"
 #include "pci.h"
-#include "caps/pm.h"
-#include "caps/px.h"
-#include "caps/msi.h"
-#include "caps/msix.h"
+#include "pci_caps/pm.h"
```
tmakatos, line 41: Maybe we should put all PCI related stuff into a separate directory, e.g. `lib/pci/caps/pm.h`?

tmakatos (approved):

include/muser.h
jlevon: Well these headers are public, so they'd be under include/pci/caps/ 
I can do that if you like along with include/pci/pci.h
Do you want lib/pci/muser_pci.c too?

jlevon: I'm going to merge this as is now, we can look at this again later if we like

PR #118, opened 2020-11-24
handle_dma_map_or_unmap(): correct fd handling

jlevon:
This is a hack to make progress: if we get a DMA_MAP request without an fd,
ignore it. This means that a server's DMA_READ/WRITE to that region won't work,
and DMA_UNMAP won't either.


tmakatos:
superseded by #119

jlevon:
I actually don't think this is true - we still need at least the fd changes I had. Obviously the ignoring of the region was just a hack though.

tmakatos:
You're right, I'll take another look.

tmakatos (approved):

PR #119, opened 2020-11-24
ignore DMA map/unmap if device isn't using DMA controller

tmakatos:


test/test.c
```diff
@@ -0,0 +1,60 @@
...
+#include <stdarg.h>
+#include <stddef.h>
+#include <setjmp.h>
+#include <cmocka.h>
+#include <limits.h>
+#include <errno.h>
+#include <stdio.h>
+#include <assert.h>
+
+#include "../include/muser.h"
```
jlevon, line 40: nit, I used include_directories() at the top so you shouldn't need the relative paths for either here

lib/muser_ctx.c
```diff
@@ -508,6 +508,10 @@ handle_dma_map_or_unmap(lm_ctx_t *lm_ctx, uint32_t size, bool map,
         return -EINVAL;
     }
 
+    if (lm_ctx->dma == NULL) {
```
jlevon, line 511: you can remove :522 now

jlevon (changes requested):

tmakatos:
I've also tentatively introduced the cmocka unit test framework. I'd like to start doing better unit testing, ideally we should check that internal functions were (or weren't) with the correct arguments. I had tried to do this a year ago but got stuck (https://gcc.gnu.org/legacy-ml/gcc-help/2019-11/msg00015.html). In that email if we call `__wrap_foo` in `bar` then it works correctly, however it doesn't work at all in libmuser.

test/test.c
tmakatos: ok

lib/muser_ctx.c
tmakatos: ok

jlevon (approved):
Yeah, great to see a start on the testing too. I did have some vague thoughts on more unit testing and how we might approach it, maybe we can talk about that later this week

PR #125, opened 2020-11-25
PCI regions idx enum renamed to start with LM_PCI

swapnili:


include/muser.h
```diff
@@ -293,18 +293,18 @@ int lm_pci_setup_caps(lm_ctx_t *lm_ctx, lm_cap_t **caps, int nr_caps);
 typedef ssize_t (lm_region_access_cb_t) (void *pvt, char *buf, size_t count,
                                          loff_t offset, bool is_write);
 
-/* FIXME these are PCI regions */
+/* PCI regions */
 enum {
```
jlevon, line 297: I don't think we need the _DEV_ part do we?
Also out of context, I found the "REG" abbreviation to be easily confused with "register" - could we make it be "REGION" ?

jlevon (approved):

include/muser.h
swapnili: 1. I think the DEV part will be useful when we start supporting non-pci devices.
2. I agree, will change it to REGION

jlevon (approved):

Issue #101, opened 2020-11-18
add tests that round-trip across vfio-user protocol


jlevon:
Need to figure out if/how we can adapt the sample client/server code to be the basis of a test suite and start covering some of the basic interactions automatically.

tmakatos:
FYI I've tentatively included the cmocka unit test framework in #119.

tmakatos:
In https://github.com/nutanix/muser/pull/129, because functions within the same file can't be mocked using the `-Wl,--wrap` trick I have to essentially redefine the function with it's wrapped version within the source file. While this change is minimal, we do have to change the source to accomodate unit testing. We need to define whether the unit of a unit test is a function or a whole file. If it's the latter then we won't have to do such tricks but we limit which functions we can test. @jlevon @swapnili thoughts?

jlevon:
Updating this ticket a little bit. We have a reasonable level of coverage in terms of unit tests plus some other parts. #238 will add some basic vfio-user protocol usage. But it would be nice to have this expanded for fuller coverage of the protocol.

jlevon:
I'm marking this is basically fixed now we have the Python framework.

PR #129, opened 2020-11-26
allow DMA regions without file descriptor

### LINKED ISSUE #123: allow DMA regions without file descriptor in the DMA controller

tmakatos:
This patch allows DMA regions without a file descriptor, including unit tests.

Some code changes required to make unit testing work is discussed in #101.

fixes #123

### DIFF
--- a/lib/muser_ctx.c
+++ b/lib/muser_ctx.c
@@ -519,7 +519,7 @@ handle_dma_map_or_unmap(lm_ctx_t *lm_ctx, uint32_t size, bool map,
                         struct vfio_user_dma_region *dma_regions)
 {
     int nr_dma_regions;
-    int ret, i;
+    int ret, i, fdi;
 
     assert(lm_ctx != NULL);
     assert(fds != NULL);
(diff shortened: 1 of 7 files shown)


lib/muser_ctx.c
```diff
@@ -499,70 +499,78 @@ handle_device_get_info(lm_ctx_t *lm_ctx, uint32_t size,
...
-                                            fd,
+                                            fds[i],
                                             dma_regions[i].offset);
             if (ret < 0) {
                 lm_log(lm_ctx, LM_INF,
                        "failed to add DMA region %#lx-%#lx offset=%#lx fd=%d: %s",
                        dma_regions[i].addr,
                        dma_regions[i].addr + dma_regions[i].size - 1,
-                       dma_regions[i].offset, fd,
+                       dma_regions[i].offset, fds[i],
```
jlevon, line 566: This looks like it broke my fix. Consider two regions being passed in:
region[0] !mappable
region[1] mappable

When we get to region[1], i == 1, but the fd for it is in fd[0] (which you over-wrote with -1)

tmakatos: Ah yes, though we don't allow it because of the check we do earlier:
```
if (map && nr_fds > 0 && nr_dma_regions != nr_fds) {
        lm_log(lm_ctx, LM_ERR, "expected %d fds but got %d instead",
               nr_dma_regions, nr_fds);
        return -EINVAL;
    }
```
However this check is wrong since we should allow mappable and non-mappable regions in the same message. Let me fix this.

tmakatos: @jlevon fixed, please take another look.

jlevon (approved):

PR #133, opened 2020-11-27
refactor process_request

tmakatos:
This patch refactors process_request to simplify code in the following way:
1. Failures before exec_command do not require a response to be sent and all
   passed descriptors must be released.
2. Failures after exec_command require a response. File descriptors must be
   released depending on how many file descriptors where successfully consumed.

This refactoring makes it simpler to implement releasing file descriptors in
the 2nd case.

### DIFF
--- a/lib/libvfio-user.c
+++ b/lib/libvfio-user.c
@@ -842,20 +842,145 @@ get_next_command(vfu_ctx_t *vfu_ctx, struct vfio_user_header *hdr, int *fds,
 }
 
 static int
+exec_command(vfu_ctx_t *vfu_ctx, struct vfio_user_header *hdr, size_t size,
+             int *fds, int *nr_fds,
+             struct iovec *_iovecs, struct iovec **iovecs, size_t *nr_iovecs,
+             bool *free_iovec_data)
+
+{
+    int ret;
+    struct vfio_irq_info irq_info;
+    struct vfio_device_info dev_info;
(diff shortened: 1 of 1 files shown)


lib/libvfio-user.c
```diff
@@ -890,101 +871,143 @@ process_request(vfu_ctx_t *vfu_ctx)
...
+     * FIXME if migration device state is VFIO_DEVICE_STATE_STOP then only
+     * migration-related operations should execute. However, some operations
+     * are harmless (e.g. get region info). At the minimum we should fail
+     * accesses to device regions other than the migration region. I'd expect
+     * DMA unmap and get dirty pages to be required even in the stop-and-copy
+     * state.
+     */
+
+    nr_fds = vfu_ctx->client_max_fds;
+    fds = alloca(nr_fds * sizeof(int));
```
jlevon, line 1001: I know this is currently safe, but alloca() is the equivalent of strcpy() - you see it and you immediately wonder if this code is secure. I think we should ban it.

In this case, simply doing:

    int fds[MUSER_MAX_CLIENT_FDS] = { [ 0 ... MUSER_MAX_CLIENT_FDS - 1] = -1 };

seems fine, and is *much more clearly* secure. (We want the -1 init to make it clear there's no fd there)

I know you didn't add this, but we should clean it up, if not now.

jlevon (approved):

PR #134, opened 2020-11-30
Update API's to accept sparse mmap areas as array and its count

swapnili:


lib/private.h
```diff
@@ -67,6 +67,11 @@ typedef struct {
 
 struct migration;
 
+struct vfu_sparse_mmap_areas {
+    int nr_mmap_areas;
+    struct vfu_mmap_area areas[];
```
tmakatos, line 72: Shall we replace `struct vfu_mmap_area` with `struct iovec`. The latter exists exactly for that reason and we don't have to introduce yet another data type.

lib/libvfio-user.c
```diff
@@ -1393,23 +1393,29 @@ int vfu_pci_setup_caps(vfu_ctx_t *vfu_ctx, vfu_cap_t **caps, int nr_caps)
...
+    areas_sz  = nr_mmap_areas * sizeof(struct vfu_mmap_area);
+
+    smmap_areas = calloc(1, sizeof(struct vfu_sparse_mmap_areas) + areas_sz);
+    if (smmap_areas == NULL) {
         return -ENOMEM;
     }
 
-    memcpy(reg_info->mmap_areas, mmap_areas, size);
+    smmap_areas->nr_mmap_areas = nr_mmap_areas;
+    for (i = 0; i < nr_mmap_areas; i++) {
```
tmakatos, line 1414: Why not just `memcpy(smmap_areas->areas, mmap_areas, areas_sz)`?

tmakatos (changes requested):

lib/libvfio-user.c
```diff
@@ -1393,23 +1393,29 @@ int vfu_pci_setup_caps(vfu_ctx_t *vfu_ctx, vfu_cap_t **caps, int nr_caps)
...
-    int nr_mmap_areas;
-    size_t size;
+    struct vfu_sparse_mmap_areas *smmap_areas;
+    size_t areas_sz;
+    uint32_t i;
 
-    if (mmap_areas == NULL) {
-        return 0;
+    if (mmap_areas == NULL || nr_mmap_areas ==  0) {
+        return -EINVAL;
```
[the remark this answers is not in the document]
swapnili: Ack

lib/private.h
swapnili: iovec.iov_base is (void *) , I think instead of user passing address, it is better to just pass value the way we now have looks convenient. I am fine either way.

tmakatos: I don't think the type of `iovec.iov_base` is of a substantial significance here.

tmakatos (approved):

PR #135, opened 2020-11-30
don't leak passed file descriptors on failure

### LINKED ISSUE #120: incoming request fd's can be leaked
In error cases, we don't close any of the fd's we might have received in get_request_sock()

tmakatos:
Plus unit test.

fixes #120 

### DIFF
--- a/lib/irq.c
+++ b/lib/irq.c
@@ -154,15 +154,15 @@ irqs_set_data_bool(vfu_ctx_t *vfu_ctx, struct vfio_irq_set *irq_set, void *data)
 
 static int
 irqs_set_data_eventfd(vfu_ctx_t *vfu_ctx, struct vfio_irq_set *irq_set,
-                      void *data)
+                      int *data)
 {
-    int32_t *d32;
     int efd;
     __u32 i;
+    size_t j;
 
     assert(data != NULL);
-    for (i = irq_set->start, d32 = data; i < (irq_set->start + irq_set->count);
-         i++, d32++) {
+    for (i = irq_set->start, j = 0; i < (irq_set->start + irq_set->count);
+         i++, j++) {
         efd = vfu_ctx->irqs->efds[i];
         if (efd >= 0) {
             if (close(efd) == -1) {
@@ -171,8 +171,9 @@ irqs_set_data_eventfd(vfu_ctx_t *vfu_ctx, struct vfio_irq_set *irq_set,
 
             vfu_ctx->irqs->efds[i] = -1;
         }
-        if (*d32 >= 0) {
-            vfu_ctx->irqs->efds[i] = *d32;
+        if (data[j] >= 0) {
+            vfu_ctx->irqs->efds[i] = data[j];
+            consume_fd(data, j);
         }
         vfu_log(vfu_ctx, VFU_DBG, "event fd[%d]=%d", i, vfu_ctx->irqs->efds[i]);
     }
--- a/lib/irq.c
+++ b/lib/irq.c
@@ -363,7 +363,6 @@ handle_device_set_irqs(vfu_ctx_t *vfu_ctx, uint32_t size,
             data = irq_set + 1;
             break;
         default:
-            // FIXME?
             vfu_log(vfu_ctx, VFU_ERR, "invalid IRQ type %d",
                     irq_set->flags & VFIO_IRQ_SET_DATA_TYPE_MASK);
             return -EINVAL;
--- a/lib/irq.c
+++ b/lib/irq.c
@@ -171,14 +171,12 @@ irqs_set_data_eventfd(vfu_ctx_t *vfu_ctx, struct vfio_irq_set *irq_set,
 
             vfu_ctx->irqs->efds[i] = -1;
         }
-        if (data[j] >= 0) { /* TODO IIUC this will always be >= 0? */
-            vfu_ctx->irqs->efds[i] = data[j];
-            /*
-             * We've already checked in handle_device_set_irqs that
-             * nr_fds == irq_set->count.
-             */
-            consume_fd(data, irq_set->count, j);
-        }
+        assert(data[j] >= 0);
+        /*
+         * We've already checked in handle_device_set_irqs that
+         * nr_fds == irq_set->count.
+         */
+        vfu_ctx->irqs->efds[i] = consume_fd(data, irq_set->count, j);
         vfu_log(vfu_ctx, VFU_DBG, "event fd[%d]=%d", i, vfu_ctx->irqs->efds[i]);
     }
 
--- a/lib/irq.c
+++ b/lib/irq.c
@@ -171,9 +171,13 @@ irqs_set_data_eventfd(vfu_ctx_t *vfu_ctx, struct vfio_irq_set *irq_set,
 
             vfu_ctx->irqs->efds[i] = -1;
         }
-        if (data[j] >= 0) {
+        if (data[j] >= 0) { /* TODO IIUC this will always be >= 0? */
             vfu_ctx->irqs->efds[i] = data[j];
-            consume_fd(data, j);
+            /*
+             * We've already checked in handle_device_set_irqs that
+             * nr_fds == irq_set->count.
+             */
+            consume_fd(data, irq_set->count, j);
         }
         vfu_log(vfu_ctx, VFU_DBG, "event fd[%d]=%d", i, vfu_ctx->irqs->efds[i]);
     }
--- a/lib/irq.c
+++ b/lib/irq.c
@@ -337,7 +337,7 @@ handle_device_get_irq_info(vfu_ctx_t *vfu_ctx, uint32_t size,
 
 int
 handle_device_set_irqs(vfu_ctx_t *vfu_ctx, uint32_t size,
-                       int *fds, int nr_fds, struct vfio_irq_set *irq_set)
+                       int *fds, size_t nr_fds, struct vfio_irq_set *irq_set)
 {
     void *data = NULL;
 
@@ -345,13 +345,17 @@ handle_device_set_irqs(vfu_ctx_t *vfu_ctx, uint32_t size,
     assert(irq_set != NULL);
 
     if (size < sizeof *irq_set || size != irq_set->argsz) {
+        vfu_log(vfu_ctx, VFU_ERR, "bad size %d", size);
         return -EINVAL;
     }
 
     switch (irq_set->flags & VFIO_IRQ_SET_DATA_TYPE_MASK) {
         case VFIO_IRQ_SET_DATA_EVENTFD:
             data = fds;
-            if (nr_fds != (int)irq_set->count) {
+            if (nr_fds != irq_set->count) {
+                vfu_log(vfu_ctx, VFU_ERR,
+                        "bad number of FDs, expected=%u, actual=%d", nr_fds,
+                        irq_set->count);
                 return -EINVAL;
             }
             break;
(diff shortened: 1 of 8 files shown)


lib/irq.c
```diff
@@ -360,6 +364,8 @@ handle_device_set_irqs(vfu_ctx_t *vfu_ctx, uint32_t size,
             break;
         default:
             // FIXME?
```
jlevon, line 366: you fixed this FIXME, can remove

lib/libvfio-user.c
```diff
@@ -1004,15 +1020,19 @@ process_request(vfu_ctx_t *vfu_ctx)
...
-    ret = exec_command(vfu_ctx, &hdr, ret, fds, &nr_fds, _iovecs, &iovecs,
+    ret = exec_command(vfu_ctx, &hdr, ret, fds, &_nr_fds, _iovecs, &iovecs,
                        &nr_iovecs, &free_iovec_data);
 
     /*
      * TODO: In case of error during command handling set errno respectively
      * in the reply message.
      */
     if (ret < 0) {
+        for (i = _nr_fds; i < nr_fds; i++) {
```
jlevon, line 1033: this needs to happen on success too.

lib/libvfio-user.c
```diff
@@ -939,6 +950,9 @@ exec_command(vfu_ctx_t *vfu_ctx, struct vfio_user_header *hdr, size_t size,
         case VFIO_USER_DEVICE_SET_IRQS:
             ret = handle_device_set_irqs(vfu_ctx, hdr->msg_size, fds, *nr_fds,
                                          cmd_data);
+            if (ret < 0) {
```
jlevon, line 953: I don't understand why this is here rather than general

jlevon (changes requested):
I think there's a simpler way to do this:

int consume_fd(int *fds, size_t nr_fds, size_t index)
{
   int fd;

   if (index >= nr_fds) {
       // log
       return -1;
   }

   fd = *fds[index];
   *fdp = -1;
   return fd;
}

Use that everywhere. Now it's really clear when we're using the fds.

Then after the exec_command(), walk through any remaining fds in the array that aren't -1, closing them and complaining in the log as needed.

lib/libvfio-user.c
```diff
@@ -506,24 +505,25 @@ handle_device_get_info(vfu_ctx_t *vfu_ctx, uint32_t size,
...
  * @dma_regions: memory that contains the DMA regions to be mapped/unmapped
  *
- * @returns 0 on success, -errno on failure.
+ * @returns 0 on success, -errno on failure. @nr_fds receives the number of
+ * mappable DMA regions that were successfully added.
  */
 int
 handle_dma_map_or_unmap(vfu_ctx_t *vfu_ctx, uint32_t size, bool map,
-                        int *fds, int nr_fds,
+                        int *fds, size_t *nr_fds,
```
swapnili, line 517: IMO let's not abuse nr_fds to return back with successfully added DMA regions, as nr_fds are suppose to be the size of fds array. Maybe have separate argument to return added regions.
In-fact for such situations returning >0 is useful.

lib/irq.c
tmakatos: > you fixed this FIXME, can remove

done

lib/libvfio-user.c
tmakatos: right

tmakatos: > this needs to happen on success too.

right

lib/libvfio-user.c
tmakatos: > I don't understand why this is here rather than general

It would be wrong to blindly do this. `handle_dma_map_or_unmap` can return `<0` but it might have consumed file descriptors.

jlevon: Sure, but this wouldn't be specific to SET_IRQS either. I think the suggestion I made is much easier.

tmakatos:
I've adopted @jlevon 's suggestion and it does result in simpler code, however I made `consume_fd` much simpler. @jlevon @swapnili please take another look.

lib/libvfio-user.c
tmakatos: This is now obsolete.

lib/libvfio-user.c
tmakatos: Now obsolete becasue of `consume_fd`.

lib/irq.c
```diff
@@ -171,8 +171,9 @@ irqs_set_data_eventfd(vfu_ctx_t *vfu_ctx, struct vfio_irq_set *irq_set,
 
             vfu_ctx->irqs->efds[i] = -1;
         }
-        if (*d32 >= 0) {
-            vfu_ctx->irqs->efds[i] = *d32;
+        if (data[j] >= 0) {
+            vfu_ctx->irqs->efds[i] = data[j];
```
jlevon, line 175: I actually think this is still much cleaner to return the fd from consume_fd() - you don't need to mention the array and index twice - but I suppose it's not a big deal.

lib/libvfio-user.c
```diff
@@ -501,27 +500,33 @@ handle_device_get_info(vfu_ctx_t *vfu_ctx, uint32_t size,
...
-         regions, irrespectively if @nr_fds is 0.
- * @nr_fds: size of above array. It must be either 0 or exactly match
- *          the number of DMA regions in @dma_regions.
+ * @fds: array of file descriptors.
+ * @nr_fds: size of above array.
  * @dma_regions: memory that contains the DMA regions to be mapped/unmapped
  *
- * @returns 0 on success, -errno on failure.
+ * @returns 0 on success, -errno on failure. @nr_fds receives the number of
+ * mappable DMA regions that were successfully added.
```
jlevon, line 520: this last sentence is stale again now

lib/libvfio-user.c
```diff
@@ -536,15 +541,19 @@ handle_dma_map_or_unmap(vfu_ctx_t *vfu_ctx, uint32_t size, bool map,
...
         if (map) {
             int fd = -1;
             if (dma_regions[i].flags == VFIO_USER_F_DMA_REGION_MAPPABLE) {
                 if (fdi == nr_fds) {
-                    return -EINVAL;
+                    ret = -EINVAL;
+                    break;
                 }
-                fd = fds[fdi++];
+                fd = fds[fdi];
```
jlevon, line 556: this I really don't like though! The whole point of consume_fd() was that it happens at the point we deference the fd array.

If dma_controller_add_region() fails, then we should explicitly close(fd) that we just took ownership of. That's much more obvious pattern than what's here now.

lib/libvfio-user.c
```diff
@@ -1006,13 +1021,20 @@ process_request(vfu_ctx_t *vfu_ctx)
         return ret;
     }
 
-    ret = exec_command(vfu_ctx, &hdr, ret, fds, &nr_fds, _iovecs, &iovecs,
+    ret = exec_command(vfu_ctx, &hdr, ret, fds, nr_fds, _iovecs, &iovecs,
                        &nr_iovecs, &free_iovec_data);
 
+    for (i = 0; i < nr_fds; i++) {
+        if (fds[i] != -1) {
```
jlevon, line 1028: do we want to log this?

jlevon (changes requested):

lib/irq.c
tmakatos: I missed the point of `consume_fd`, it makes even more sense now, will fix.

lib/libvfio-user.c
tmakatos: right

lib/libvfio-user.c
tmakatos: right, will fix

lib/libvfio-user.c
tmakatos: sure

tmakatos:
@jlevon I've rebased on master nad addressed your comment via additional commits, please take another look.

lib/libvfio-user.c
```diff
@@ -553,19 +575,22 @@ handle_dma_map_or_unmap(vfu_ctx_t *vfu_ctx, uint32_t size, bool map,
                                             fd,
                                             dma_regions[i].offset);
             if (ret < 0) {
+                if (fd != -1) {
```
jlevon, line 578: Sorry, I'm not sure why we're doing this. What's the point of restore_fd()?
We now own the fd here, so we can and should just close() it.

lib/irq.c
```diff
@@ -171,8 +171,13 @@ irqs_set_data_eventfd(vfu_ctx_t *vfu_ctx, struct vfio_irq_set *irq_set,
 
             vfu_ctx->irqs->efds[i] = -1;
         }
-        if (*d32 >= 0) {
-            vfu_ctx->irqs->efds[i] = *d32;
+        if (data[j] >= 0) { /* TODO IIUC this will always be >= 0? */
+            vfu_ctx->irqs->efds[i] = data[j];
```
jlevon, line 175: you can now do:

vfu_ctx->irqs->efds[i] = consume_fd(data, irq_set->count, j);

As for the TODO: why not add an assert in consume_fd() that fd is >= 0?

lib/libvfio-user.c
```diff
@@ -1006,13 +1034,21 @@ process_request(vfu_ctx_t *vfu_ctx)
         return ret;
     }
 
-    ret = exec_command(vfu_ctx, &hdr, ret, fds, &nr_fds, _iovecs, &iovecs,
+    ret = exec_command(vfu_ctx, &hdr, ret, fds, nr_fds, _iovecs, &iovecs,
                        &nr_iovecs, &free_iovec_data);
 
+    for (i = 0; i < nr_fds; i++) {
+        if (fds[i] != -1) {
+            vfu_log(vfu_ctx, VFU_INF, "closing fd=%d", fds[i]);
```
jlevon, line 1042: sorry, I should have been a bit clearer: this is an error, but one causable by a malicious client. So it should be debug-only, but still clarify it's not normal, e.g.

vfu_log(vfu_ctx, VFU_DBG, "closing unexpected fd %d (index %zu) from cmd %u", fds[i], i, hdr.cmd);

jlevon (changes requested):

lib/libvfio-user.c
tmakatos: Right, the `restore_fd` logic is unnecessary.

lib/irq.c
tmakatos: That was my intention, I forgot to do that.

Also, replaced the TODO with an assert.

lib/libvfio-user.c
tmakatos: It's not necessarily caused by a malicious client. While _additional_ FDs are malicious and will be closed, if a failure occurs we will be closing a legitimate FD. Currently this can only happen if we run out of DMA regions.

tmakatos:
Address your comments (except for the one regarding logging an message when closing FDs).

lib/libvfio-user.c
tmakatos: Another case where the FD is not unexpected is setting up interrupts where the client passed the wrong flags.

jlevon: That's true enough. We still probably want to survey at some point our logging for when it *can* be malicious though

jlevon (approved):
Thanks, and sorry for all the back and forth!

lib/libvfio-user.c
tmakatos: In this particular case can tell which FDs are legitimate and which aren't, I've created https://github.com/nutanix/libvfio-user/issues/141 for that.

PR #139, opened 2020-11-30
fix copyright

tmakatos:

### DIFF
--- a/include/vfio-user.h
+++ b/include/vfio-user.h
@@ -1,34 +1,38 @@
-/*-
- *   BSD LICENSE
+/*
+ * Copyright (c) 2020 Nutanix Inc. All rights reserved.
  *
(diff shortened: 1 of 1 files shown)


include/vfio-user.h
```diff
@@ -1,34 +1,29 @@
-/*-
- *   BSD LICENSE
- *
- *   Copyright (c) Intel Corporation.
- *   All rights reserved.
+/*
+ * Copyright (c) 2020 Intel Corporation. All rights reserved.
  *
```
swapnili, line 3: Should we also add Nutanix here? As we also made changes to this file and we will be maintaining it.

tmakatos: Good point, updated.

tmakatos:
@swapnili please take another look.

swapnili (approved):

PR #144, opened 2020-12-02
introduce device and PCI types

tmakatos:


lib/libvfio-user.c
```diff
@@ -1288,10 +1295,16 @@ vfu_create_ctx(vfu_trans_t trans, const char *path, int flags, void *pvt)
         return NULL;
     }
 
+    if (dev_type != VFU_DEV_TYPE_PCI) {
+        errno = EINVAL;
```
swapnili, line 1299: Should this be ENOTSUP?

include/libvfio-user.h
```diff
@@ -261,18 +268,32 @@ vfu_setup_log(vfu_ctx_t *vfu_ctx, vfu_log_fn_t *log, vfu_log_lvl_t log_lvl);
...
+ * @revision: PCI/PCI-X/PCIe revision
+ *
+ * @returns 0 on success, -1 on failure and sets errno.
  */
 int
 vfu_pci_setup_config_hdr(vfu_ctx_t *vfu_ctx, vfu_pci_hdr_id_t id,
                          vfu_pci_hdr_ss_t ss, vfu_pci_hdr_cc_t cc,
-                         bool extended);
+                         vfu_pci_type_t pci_type,
+                         int revision __attribute__((unused)));
```
swapnili, line 296: Consider using UNUSED macro, good to be consistent across the code.

lib/libvfio-user.c
```diff
@@ -1455,9 +1480,10 @@ vfu_setup_region(vfu_ctx_t *vfu_ctx, int region_idx, size_t size,
     switch(region_idx) {
     case VFU_PCI_DEV_BAR0_REGION_IDX ... VFU_PCI_DEV_VGA_REGION_IDX:
         // Validate the config region provided.
-        if (region_idx == VFU_PCI_DEV_CFG_REGION_IDX &&
-            !is_valid_pci_config_space_region(flags, size)) {
+        if (region_idx == VFU_PCI_DEV_CFG_REGION_IDX) {
+            if (flags == VFU_REG_FLAG_RW) {
```
jlevon, line 1484: what is this really testing for? Not using & for flags looks odd at least?

jlevon (changes requested):

lib/libvfio-user.c
tmakatos: sorry that shoud be `!=`

tmakatos: > what is this really testing for? Not using & for flags looks odd at least?

sorry that shoud be !=

tmakatos:
@jlevon fixed, please take another look

lib/libvfio-user.c
tmakatos: No, `ENOTSUP` should be for thing we _could_ support but we don't. The user might set `dev_type` to something completely bogus.

jlevon (approved):

lib/libvfio-user.c
swapnili: But today we only support PCI and isn't the idea of adding device_type is to say we will support non-pci in future.

tmakatos: When we add another device type we'll document it, so if the user supplies a wrong value then it's an invalid argument.

PR #148, opened 2020-12-02
emit warning when receiving extraneous file descriptors

### LINKED ISSUE #141: emit warning for extraneous passed file desriptors
Check discussion in https://github.com/nutanix/libvfio-user/pull/135

tmakatos:
fixes #141

### DIFF
--- a/lib/libvfio-user.c
+++ b/lib/libvfio-user.c
@@ -1033,13 +1033,9 @@ process_request(vfu_ctx_t *vfu_ctx)
 
     for (i = 0; i < nr_fds; i++) {
         if (fds[i] != -1) {
-            if (ret < 0) {
-                vfu_log(vfu_ctx, LOG_WARNING,
-                        "closing unexpected fd %d (index %zu) from cmd %u",
-                        fds[i], i, hdr.cmd);
-            } else {
-                vfu_log(vfu_ctx, LOG_DEBUG, "closing fd %d", fds[i]);
-            }
+            vfu_log(vfu_ctx, LOG_DEBUG,
+                    "closing unexpected fd %d (index %zu) from cmd %u",
+                    fds[i], i, hdr.cmd);
             close(fds[i]);
         }
     }


lib/libvfio-user.c
```diff
@@ -1033,7 +1033,13 @@ process_request(vfu_ctx_t *vfu_ctx)
 
     for (i = 0; i < nr_fds; i++) {
         if (fds[i] != -1) {
-            vfu_log(vfu_ctx, LOG_INFO, "closing fd=%d", fds[i]);
+            if (ret < 0) {
```
jlevon, line 1036: I'm not convinced this is right. A bad client could still use this to spam the log with WARNINGs.

Also I'm not clear on when ret == 0 but we still have leftover fd's?

tmakatos: > I'm not convinced this is right. A bad client could still use this to spam the log with WARNINGs.

So we should log with `LOG_DEBUG` or not log at all?

> 
> Also I'm not clear on when ret == 0 but we still have leftover fd's?
Right, this should be the other way round: if `ret == 0` then any FD left is an extraneous FD.

jlevon: LOG_DEBUG - it's useful for dev, and I think we are OK to presume that it's OK for bad clients to spam DEBUG messages

tmakatos: Actually, we can still have unexpected file descriptors even if `ret < 0`. However it can be tricky to differentiate, maybe we should not try to differentiate and emit a simpler warning?

tmakatos: @jlevon I commented slightly earlier than you did however my comment appears after yours for some reason. Could you take another look and see whether we really need to differentiate?

jlevon: I don't think we need to differentiate, but the message should make it clear that it's unexpected to need to close fd still (even if it was due to partial failure, it's useful in LOG_DEBUG to know there fd's left over)

tmakatos:
@jlevon fixed, please take another look.

jlevon (approved):

PR #149, opened 2020-12-02
replace struct vfu_mmap_area with struct iovec

### LINKED ISSUE #147: replace struct vfu_mmap_area with struct iovec
IMO we should try to resuse existing data types rather when possible rather than introducing new ones. `struct vfu_mmap_area` is one of them, we can replace it with `struct iovec`.

@jlevon @swapnili thoughts?

tmakatos:
fixes #147

### DIFF
--- a/include/libvfio-user.h
+++ b/include/libvfio-user.h
@@ -69,11 +69,6 @@ typedef struct {
 
 typedef struct vfu_ctx vfu_ctx_t;
 
-struct vfu_mmap_area {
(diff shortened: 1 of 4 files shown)


jlevon (approved):

tmakatos:
I've accidentally included the fix for https://github.com/nutanix/libvfio-user/issues/141, which apparently has issues. I'll see whether we can fix soon it via it's proper PR, otherwise I'll revert it.

PR #155, opened 2020-12-03
draft

jlevon:
Here's roughly where I got to.

I realised my plan yesterday had a potentially fatal flaw - we can't add callbacks for caps if we don't have the ctx to add them to. So I reworked it instead to be used like this:

```
// allocs, inits, and stores the whole cfg space, returning a pointer to it
vfu_pci_hdr_t *space = vfu_pci_init(ctx, ...);

// use helpers
vfu_pci_set_id(ctx, PCI_ID_NVME_FOO,...);
vfu_pci_set_class(ctx, STORAGE_NVME_CLASS, NVME_SUBCLASS, ...);
vfu_pci_add_capability(ctx, PCI_CAP_ID_EXP, false, pxcap, sizeof (*pxcap));

...

vfu_ctx_drive();

I don't see the need to support anything other than Type 0 header - thoughts?

Thanos - you were right, there's no need at all for a 'size' parameter.

There are several FIXME comments below I'd be interested in your opinions on. In general, we should take a look at what spdk needs to do - currently it has a callback for all of extended config space!

### DIFF
--- a/include/libvfio-user.h
+++ b/include/libvfio-user.h
@@ -305,7 +305,7 @@ vfu_pci_set_class(vfu_ctx_t *vfu_ctx, uint8_t base, uint8_t sub, uint8_t pi);
  */
 int
 vfu_pci_add_capability(vfu_ctx_t *vfu_ctx, uint16_t cap_id, bool extended,
-                       void *data, size_t size);
+                       void *data);
 
 /**
  * Callback function that gets called when a capability is accessed.
@@ -334,6 +334,9 @@ typedef ssize_t (vfu_cap_access_cb_t) (void *pvt, uint8_t id,
  * and PCI_EXT_CAP_ID_VNDR), this can be used to add a capability where reads
  * and writes are delegated to a callback.
  *
+ * Note the size of the capability is derived either from the ID, or from an
+ * embedded length field for PCI_(EXT_)CAP_ID_VNDR.
+ *
  * @vfu_ctx: the libvfio-user context
  * @cap: capability ID to add
  * @extended: true if an extended capability (PCI_EXT_CAP_ID_*)
@@ -342,7 +345,7 @@ typedef ssize_t (vfu_cap_access_cb_t) (void *pvt, uint8_t id,
  */
 int
 vfu_pci_add_capability_cb(vfu_ctx_t *vfu_ctx, uint16_t cap_id, bool extended,
-                          size_t size, vfu_cap_access_cb_t cb, void *pvt);
+                          vfu_cap_access_cb_t cb, void *pvt);
 
 #define VFU_REGION_FLAG_READ    (1 << 0)
 #define VFU_REGION_FLAG_WRITE   (1 << 1)
--- a/include/libvfio-user.h
+++ b/include/libvfio-user.h
@@ -100,9 +100,18 @@ vfu_mmap(vfu_ctx_t * vfu_ctx, off_t offset, size_t length);
 
 /*
  * Returns a pointer to the standard part of the PCI configuration space.
+ * FIXME: spdk uses this to fiddle with the BARs
  */
 vfu_pci_config_space_t *
 vfu_pci_get_config_space(vfu_ctx_t *vfu_ctx);
+
+/**
+ * Returns the non-standard part of the PCI configuration space.
+ * @vfu_ctx: the libvfio-user context
+ * FIXME: do we need this?
+ */
+uint8_t *
+vfu_get_pci_non_std_config_space(vfu_ctx_t *vfu_ctx);
 
 #define VFU_DMA_REGIONS  0x10
 
@@ -115,38 +124,10 @@ vfu_pci_get_config_space(vfu_ctx_t *vfu_ctx);
  */
 typedef void (vfu_log_fn_t) (void *pvt, int level, const char *msg);
 
-/**
- * Callback function that gets called when a capability is accessed. The
- * callback is not called when the ID and next fields are accessed, these are
- * handled by the library.
- *
- * @pvt: private pointer
- * @id: capability ID being accessed
- * @buf: pointer to data being read or written
- * @count: number of bytes being read or written
- * @offset: offset within the capability
- * @is_write: whether the capability is read or written
- *
- * @returns the number of bytes read or written
- */
-typedef ssize_t (vfu_cap_access_t) (void *pvt, uint8_t id,
-                                    char *buf, size_t count,
-                                    loff_t offset, bool is_write);
-
-/* FIXME does it have to be packed as well? */
-typedef union {
-    struct msicap msi;
-    struct msixcap msix;
-    struct pmcap pm;
-    struct pxcap px;
-} vfu_cap_t;
-
 typedef enum {
     VFU_TRANS_SOCK,
     VFU_TRANS_MAX
 } vfu_trans_t;
-
-#define VFU_MAX_CAPS (PCI_CFG_SPACE_SIZE - PCI_STD_HEADER_SIZEOF) / PCI_CAP_SIZEOF
 
 /*
  * FIXME the names of migration callback functions are probably far too long,
@@ -257,9 +238,6 @@ vfu_create_ctx(vfu_trans_t trans, const char *path,
 int
 vfu_setup_log(vfu_ctx_t *vfu_ctx, vfu_log_fn_t *log, int level);
 
-//TODO: Check other PCI header registers suitable to be filled by device.
-//      Or should we pass whole vfu_pci_hdr_t to be filled by user.
-
 typedef enum {
     VFU_PCI_TYPE_CONVENTIONAL,
     VFU_PCI_TYPE_PCI_X_1,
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by tmakatos** [comment]
I got confused by the name of the PR thinking that's it's still in progress, I'll take a look tomorrow.

File: include/libvfio-user.h [hunk 1]
```diff
@@ -268,33 +246,106 @@ typedef enum {
...
- * @caps: array of (vfu_cap_t *)
- * *nr_caps: number of elements in @caps
+ * @cap: capability ID to add
+ * @extended: true if an extended capability (PCI_EXT_CAP_ID_*)
+ * @cb: callback
+ * @pvt: private data for callback
  */
 int
-vfu_pci_setup_caps(vfu_ctx_t *vfu_ctx, vfu_cap_t **caps, int nr_caps);
+vfu_pci_add_capability_cb(vfu_ctx_t *vfu_ctx, uint16_t cap_id, bool extended,
```
tmakatos, line 347: For now there is no need for a callback. We don't know PCI/PCIe well enough to understand whether we will ever need a callback. If we do need one, it's best to introduce a new function that sets a callback for a particular capability. If not, then we won't unnecessarily burden the API with parameters that are always unused.

Also, there's no need for `pvt`, we can pass `vfu_ctx->pvt` to the callback.

include/libvfio-user.h
```diff
@@ -268,33 +246,106 @@ typedef enum {
...
+ * @pvt: private pointer
+ * @id: capability ID being accessed
+ * @buf: pointer to data being read or written
+ * @count: number of bytes being read or written
+ * @offset: offset within the capability (including the header)
+ * @is_write: whether the capability is read or written
+ *
+ * @returns the number of bytes read or written
+ *
+ * FIXME: cb would have to identify each cap by "pvt" - is that OK?
```
tmakatos, line 322: `pvt` must be `vfu_ctx->pvt`

include/libvfio-user.h
```diff
@@ -268,33 +246,106 @@ typedef enum {
...
+ * FIXME: this can already be handled by region callback - do we need this
+ * too?
```
tmakatos, line 324: The region callback (`vfu_region_access_cb_t`) doesn't have the capability ID, so we need `vfu_cap_access_cb_t`. Alternatively, we can add a region parameter to `vfu_region_access_cb_t` so when a region is accessed this parameter will be the region and when a capability is accessed it will be tha capability ID.

include/libvfio-user.h
```diff
@@ -268,33 +246,106 @@ typedef enum {
...
+typedef ssize_t (vfu_cap_access_cb_t) (void *pvt, uint8_t id,
+                                       char *buf, size_t count,
+                                       loff_t offset, bool is_write);
 
-//TODO: Support variable size capabilities.
 /**
- * Setup PCI capabilities.
+ * Add a PCI capability to PCI config space with a callback.
+ *
+ * For capabilities not handled by the library, (in particular PCI_CAP_ID_VNDR
```
tmakatos, line 333: Not sure about this yet, see my comment further down.

include/libvfio-user.h
```diff
@@ -268,33 +246,106 @@ typedef enum {
...
+ * and writes are delegated to a callback.
+ *
+ * Note the size of the capability is derived either from the ID, or from an
+ * embedded length field for PCI_(EXT_)CAP_ID_VNDR.
+ *
  * @vfu_ctx: the libvfio-user context
- * @caps: array of (vfu_cap_t *)
- * *nr_caps: number of elements in @caps
+ * @cap: capability ID to add
+ * @extended: true if an extended capability (PCI_EXT_CAP_ID_*)
```
tmakatos, line 342: s/extended/express

include/libvfio-user.h
tmakatos, line 347: s/extended/express

include/libvfio-user.h
```diff
@@ -268,33 +246,106 @@ typedef enum {
...
+ * and PCI_EXT_CAP_ID_VNDR), this can be used to add a capability where reads
+ * and writes are delegated to a callback.
+ *
+ * Note the size of the capability is derived either from the ID, or from an
+ * embedded length field for PCI_(EXT_)CAP_ID_VNDR.
+ *
  * @vfu_ctx: the libvfio-user context
- * @caps: array of (vfu_cap_t *)
- * *nr_caps: number of elements in @caps
+ * @cap: capability ID to add
```
tmakatos, line 341: I would add "(PCI_CAP_ID_ or PCI_EXT_CAP_ID_ in <linux/pci_regs.h>)".

include/libvfio-user.h
```diff
@@ -268,33 +246,106 @@ typedef enum {
...
+ * Certain standard capabilities are handled entirely within the library:
+ *
+ * PCI_CAP_ID_EXP (pxcap)
+ * PCI_CAP_ID_MSIX (msixcap)
+ * PCI_CAP_ID_PM (pmcap)
+ *
+ * However, they must still be explicitly initialized and added here.
  *
- * @returns 0 on success, -1 on failure and sets errno.
+ * FIXME: are we really making the user init these too? why? Is there a better
```
tmakatos, line 289: AFAIK there's no other way around it, the capabilities must be configured by the user. We can introduce APIs to do this but I'm not sure how helpful that would be, capabilities are simple structs with many fields.

include/libvfio-user.h
```diff
@@ -268,33 +246,106 @@ typedef enum {
...
+ * API?
+ *
+ * The contents of @data is copied in. It must start with either a struct
+ * pci_cap or a struct pci_extcap, with the ID field set (the 'next' field is
+ * ignored).  If the capability is not one of the handled ones above (e.g.
+ * PCI_CAP_ID_VNDR), the library will allow all reads and disallow all writes
+ * for the capability from the client.
```
tmakatos, line 296: I don't know enough about capabilities to tell whether it's OK to allow reads to go unnoticed by the device, what if some have register semantics? I think it safer to simply state that currently we only handle a small set of capabilities and that we can accept contributions for more capabilities.

include/libvfio-user.h
```diff
@@ -620,6 +664,10 @@ vfu_ctx_try_attach(vfu_ctx_t *vfu_ctx);
 /*
  * FIXME need to make sure that there can be at most one capability with a given
  * ID, otherwise this function will return the first one with this ID.
+ *
+ * FIXME do we want this? It's only used right now to get MSI-X enabled by
```
tmakatos, line 668: There can be multipe capabilities with the same ID, the vendor-specific capability is one of them, not sure about others.

Not sure what's best TBH. E.g. in the SPDK case it's used only to access `msix->mxc.mxe`. We can introduce function that access particular bits of capabilities, e.g. `vfu_pci_cap_msix_is_enabled(vfu_ctx_t *vfu_ctx)` for the SPDK case, however each capabilities has many individual fields and there are a couple of dozens capabilities, so we might easily end up with a hundend functions just for capabilities. Since we don't yet know enough we might allow direct access to the capabilities for now and if it turns out that an API is better then we can deprecate `vfu_ctx_get_cap`.

include/libvfio-user.h
```diff
@@ -100,10 +100,19 @@ vfu_mmap(vfu_ctx_t * vfu_ctx, off_t offset, size_t length);
...
 vfu_pci_config_space_t *
 vfu_pci_get_config_space(vfu_ctx_t *vfu_ctx);
 
+/**
+ * Returns the non-standard part of the PCI configuration space.
+ * @vfu_ctx: the libvfio-user context
+ * FIXME: do we need this?
+ */
+uint8_t *
+vfu_get_pci_non_std_config_space(vfu_ctx_t *vfu_ctx);
```
tmakatos, line 114: This function doesn't need to be public.

include/libvfio-user.h
```diff
@@ -268,33 +246,106 @@ typedef enum {
...
  *
  * @vfu_ctx: the libvfio-user context
- * @id: Device and vendor ID
- * @ss: Subsystem vendor and device ID
- * @cc: Class code
  * @pci_type: PCI type (convention PCI, PCI-X mode 1, PCI-X mode2, PCI-Express)
  * @revision: PCI/PCI-X/PCIe revision
+ */
+vfu_pci_hdr_t *
+vfu_pci_init(vfu_ctx_t *vfu_ctx, vfu_pci_type_t pci_type, int revision UNUSED);
```
tmakatos, line 261: We should add the header type as a parameter and state that only type 0 is currently supported. This way if someone implements other header types in the future we won't have to change the API.

include/libvfio-user.h
```diff
@@ -268,33 +246,106 @@ typedef enum {
...
+
+/*
+ * FIXME
+ */
+void
+vfu_pci_set_id(vfu_ctx_t *vfu_ctx, uint16_t vid, uint16_t did,
```
tmakatos, line 267: We can configure these via a single function, we should look into the PCI spec and see what can be configured by the device.

tmakatos (commented):
I've added some first comments.

include/libvfio-user.h
```diff
@@ -100,10 +100,19 @@ vfu_mmap(vfu_ctx_t * vfu_ctx, off_t offset, size_t length);
 
 /*
  * Returns a pointer to the standard part of the PCI configuration space.
+ * FIXME: spdk uses this to fiddle with the BARs
```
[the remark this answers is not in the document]
jlevon: Agreed we should remove it - but we can't until we have something fixed up for spdk's use. Will need to take a look at that.

include/libvfio-user.h
jlevon: OK, let's leave the callback out of it for now.
(The pvt is needed - it's the only way an API user could identify *which* capability is being accessed.)

include/libvfio-user.h
jlevon: as per above

include/libvfio-user.h
jlevon: Well, there's nothing stopping a user parsing the cap list to figure this out, that's what I meant. It's somewhat weird to have a cb for caps and the config space region when one is container in the other.

include/libvfio-user.h
jlevon: What do you mean? I think "extended" is the right word here.

include/libvfio-user.h
jlevon: good idea

include/libvfio-user.h
jlevon: We have to do *something* if we're not doing a cb.

include/libvfio-user.h
jlevon: Yeah, I think you're probably right. But we could potentially make this be a helper against the config space pointer, depends what we do above.

include/libvfio-user.h
jlevon: I'm not convinced it will ever make sense to have the other types, but sure.

include/libvfio-user.h
jlevon: I've split them up because it looks way clearer to me than one function with loads of params and little context as to what is what.

include/libvfio-user.h
tmakatos: > (The pvt is needed - it's the only way an API user could identify _which_ capability is being accessed.)

You're right.

include/libvfio-user.h
tmakatos: > We have to do _something_ if we're not doing a cb.

I think we should not allow capabilities that are not in the list of capabilities we support then `vfu_pci_add_capability` should fail with `ENOTSUP`. This way we won't have to do anything.

include/libvfio-user.h
tmakatos: I see what you mean. I think we shouldn't have a callbacl for the PCI config space region at all, everything should be handled by the library. Then having callbacks specific for PCI capabilites (assuming we'll need them) is necessary.

include/libvfio-user.h
tmakatos: You're right.

include/libvfio-user.h
jlevon: that won't help with vendor caps: we know we have to support them, but we can't implement them.

include/libvfio-user.h
jlevon: This comes down to "need more understanding of typical devices", but there's free space in config space that could be completely arbitrary device area, isn't there?

tmakatos: I wasn't clear enough, I meant that there shouldn't be a callaback for the PCI config space _header_. I don't know whether the remaining space must be used exclusively for capabilities or the device and device driver can use this space as they please. I wouldn't think that would be the case though, if they have to do that they would use the vendor-specific capability. To make further progress on this we need to undrstand more about PCI, I've started compiling a list of questions here: https://github.com/nutanix/libvfio-user/issues/180.

include/libvfio-user.h
tmakatos: For now we can allow the host to directly read/write to a vendor-specific capability. If the device needs to handle accesses, then we can introduce a function that sets a callback for a vendor-specific capability. I implemented [support for the vendor-specific capability](https://github.com/nutanix/libvfio-user/pull/177/commits/4da7d7dd441b0f13c0ba423e74f3bfe0265a9e4e) and IMO this is the best way to solve the problem for now.

include/libvfio-user.h
```diff
@@ -268,33 +246,106 @@ typedef enum {
...
+ * @cap: capability ID to add
+ * @extended: true if an extended capability (PCI_EXT_CAP_ID_*)
+ * @data: capability data, not including the capability header
  */
 int
-vfu_pci_setup_config_hdr(vfu_ctx_t *vfu_ctx, vfu_pci_hdr_id_t id,
-                         vfu_pci_hdr_ss_t ss, vfu_pci_hdr_cc_t cc,
-                         vfu_pci_type_t pci_type,
-                         int revision __attribute__((unused)));
+vfu_pci_add_capability(vfu_ctx_t *vfu_ctx, uint16_t cap_id, bool extended,
```
swapnili, line 307: Do we need cap size here? Or user need to check specifications for size? 
I remember sometime back @tmakatos said there are also variable size capabilities.

include/libvfio-user.h
swapnili: It is not clear to me why user cannot identify capability accessed by using cap_id? We do pass cap_id in the callback.

include/libvfio-user.h
tmakatos: No need for size, capabilities can vary in size but their size can be unambiguously determined by specific fields in their contents.

include/libvfio-user.h
tmakatos: There can be multiple capabilities with the same ID, and the offset parameter we pass is relative to the capability, not the entire PCI config space.

include/libvfio-user.h
tmakatos: > Instead why not have a struct(e.g vfu_pci_std_hdr) around the standard header bits which we want user to specify. It should be unlikely that this struct will change.
> Having multiple functions is little inconvenient to users from my POV.

A struct similar to `lm_pci_info_t`? The one we had and removed? Not sure it's worth to have a struct for just 3 parameters IMO.

include/libvfio-user.h
swapnili: I thought cap_id's are unique and controlled by pcisig, check https://pcisig.com/sites/default/files/files/PCI_Code-ID_r_1_11__v24_Jan_2019.pdf
Am I missing some point here?

tmakatos: You can have multiple instances of the same capability, e.g. there can be multiple vendor-specific capabilities, each one with different size.

include/libvfio-user.h
jlevon: My thinking was generally that we'd handle what we knew how to, and punt the rest to the config space cb if there is one.

include/libvfio-user.h
jlevon: I'm not sure about the "write" part of that, hence my suggestion above. We could add a "READONLY" flag maybe?

jlevon:
We can't use cap_id because there can be more than one capability per ID for e.g. vendor caps

jlevon:
> Instead why not have a struct(e.g vfu_pci_std_hdr) around the standard header bits which we want user to specify. It should be unlikely that this struct will change.

That's what we have already. I don't agree at all that this is a readable approach:

1449     vfu_pci_hdr_id_t id = {                                                      
1450         .vid = 0x4e58,                                                           
1451         .did = 0x0001                                                            
1452     };                                                                           
1453                                                                                  
1454     vfu_pci_hdr_ss_t ss = {                                                      
1455         .vid = 0x0,                                                              
1456         .sid = 0x0,                                                              
1457     };                                                                           
1458                                                                                  
1459     vfu_pci_hdr_cc_t cc = {                                                      
1460         /* controller uses the NVM Express programming interface */              
1461         .pi = 0x02,                                                              
1462         /* non-volatile memory controller */                                     
1463         .scc = 0x08,                                                             
1464         /* mass storage controller */                                            
1465         .bcc = 0x01                                                              
1466     };                                                                           

When we can have easily readable functions that make it really clear what they do.

swapnili:
> When we can have easily readable functions that make it really clear what they do.

Ok it is good, as long as we do not end up with numerous functions.

jlevon:
The discussion points here are either resolved or should be moved over to the other PR series I have

PR #164, opened 2020-12-04
Fix pci type

tmakatos:


jlevon (approved):
(I reviewed the "Fix pci type" part.)

PR #166, opened 2020-12-04
add lspci sample

tmakatos:


README.md
```diff
@@ -92,12 +92,45 @@ Finally build your program and link with `libvfio-user.so`.
...
+The above sample implements a very simple PCI device that supports the Power
+Management PCI capability. The sample can be trivially modified to change the
+PCI configuration space header and add more PCI capabilities.
+
+
+Client/Server Implementation
+----------------------------
+
+[Client](./samples/client.c)/[server](./samples/server.c) implements a basic
+client/server model were basic tasks are performed.
```
jlevon, line 127: nit, s/were/where/

samples/lspci.c
```diff
@@ -0,0 +1,78 @@
...
+#include <assert.h>
+
+#include "libvfio-user.h"
+
+int main(void)
+{
+    int i, j;
+    char *buf;
+    const int bytes_per_line = 0x10;
+	vfu_pci_hdr_id_t id = { 0 };
```
jlevon, line 45: nit, mix of tabs and spaces?

jlevon (approved):
Looks good modulo nits!

include/libvfio-user.h
```diff
@@ -617,6 +617,18 @@ vfu_get_pci_non_std_config_space(vfu_ctx_t *vfu_ctx);
...
+ * Finalizes the device making it ready for vfu_ctx_drive or vfu_ctx_try_attach.
+ * This function is optional as it is automatically called by vfu_ctx_drive or
+ * vfu_ctx_try_attach. Calling it multiple times is idempotent.
+ *
+ * @vfu_ctx: the libvfio-user context
+ *
+ * @returns: 0 on success, -1 on error. Sets errno.
+ */
+int
+vfu_realize_ctx(vfu_ctx_t *vfu_ctx);
```
swapnili, line 630: I do not quite understand the value of making vfu_realize_ctx() API public. Anyways this is optional and ctx realisation is taken care automatically internally.

jlevon: It's nice to separate out the part that accept()s a client from ongoing operations in the later drive().

It would be nice if this documented the blocking for a client behaviour here though.

swapnili:
We do have vfu_ctx_try_attach() to accept() clients, though today its for non-blocking ctx.
We can make vfu_ctx_try_attach() also for blocking ctx, instead of having separate API just for NB accept().

jlevon:
It's also useful to separate out "setup" errors that we can only really detect during attach from others though I think. Not sure.

swapnili:
I am not sure if it is worth to have public function to get the errors separately, in any case those errors will be returned by drive() or vfu_ctx_try_attach(). My point is to be more careful while adding public functions, otherwise we will end up having pile of those.

tmakatos:
> It's also useful to separate out "setup" errors that we can only really detect during attach from others though I think. Not sure.

This is the main reason I added it. Just because we've had itnitally implemented setting up the context in `drive()` doesn't necessarily mean that it was the right thing to do. `drive()` should mean just that: attach and drive.

Issue #76, opened 2020-11-11
replace travis


jlevon:
travis is essentially did: we probably want to replace it with github actions or something

PR #171, opened 2020-12-07
replace travis with github actions

jlevon:
Travis is closing down their free tier; stop using it.


jlevon:
Fixes #76.

We should probably be building inside a docker container, so we can do centos builds etc. but this should be enough to get us unstuck for now.

swapnili (approved):

Issue #73, opened 2020-11-10
compile libmuser as a static library


changpe1:
Existing code will compile libmuser as a shared library default, we need to export LD_LIBRARY_PATH when link libmuser library, this isn't work for SPDK unit test binaries, so I would like to have a static library option, e.g: make static or make SHARED=no.

PR #173, opened 2020-12-07
build a static libvfio-user

jlevon:
Note that we intentionally don't install it: this is really for use when
building as a sub-module.


jlevon:
Fixes #73.

tmakatos (approved):
Same here, don't know much about CMake, I just tested that it works.

PR #174, opened 2020-12-07
 add shared library versioning

jlevon:
Add basic "soname" versioning (but don't yet start using it meaningfully).


tmakatos (approved):
I don't much about CMake, it looks sensible. I downloaded and tested the PR and it works.

PR #175, opened 2020-12-08
Misc fixes for vfu_ctx_try_attach() and vfu_realize_ctx()

swapnili:


include/libvfio-user.h
```diff
@@ -607,20 +606,22 @@ uint8_t *
 vfu_get_pci_non_std_config_space(vfu_ctx_t *vfu_ctx);
 
 /*
- * Attempts to attach to the transport. LIBVFIO_USER_FLAG_ATTACH_NB must be set
- * when creating the context. Returns 0 on success and -1 on error. If errno is
+ * Attempts to attach to the transport. Attach is mandatory before
+ * vfu_ctx_drive() or vfu_ctx_poll() and is non blocking if context is created
+ * with LIBVFIO_USER_FLAG_ATTACH_NB flag.
+ * Returns clients file descriptor on success and -1 on error. If errno is
```
tmakatos, line 622: s/clients/client's

lib/private.h
```diff
@@ -129,7 +129,7 @@ struct vfu_ctx {
 
     uint32_t                irq_count[VFU_DEV_NUM_IRQS];
     vfu_irqs_t              *irqs;
-    bool                    ready;
+    bool                    realized;
```
tmakatos, line 132: IMO we don't need to store this piece of information, if we decide to do so then why not call `vfu_ctx_realize` automatically if it hasn't been done so.

tmakatos (changes requested):
It's nice that we get rid of `vfu_ctx_try_attach`.

Could you add unit tests to validate that the behavior is correct?

lib/private.h
swapnili: The only use of this bool is to return EINVAL from vfu_drive_ctx() and vfu_ctx_poll() if the device is not realized(). I think instead of calling realized() everytime its better to enforce user and fail if device not realized(). 

Or we can get rid of this bool and process_context() may fail weirdly is device is not realized() by user.

include/libvfio-user.h
swapnili: Ack. Sure will add unit tests.

lib/private.h
tmakatos: drive() and poll() will also fail weirldy if the user hasn't set up regions or IRQs, I don't think we can protect against misuse. We should document very well the API and provide samples to avoid such problems.

swapnili: Yes but I see that realized() is the step which is after the device setup i.e regions(), IRQ etc. So realized() can always return errors and user can pre-do setup accordingly.
But this is debatable and we can just document and let process_context() behave with un-realized() device.

tmakatos: Realize will not necessarily return errors if the user didn't e.g. configure a region they were plan to use, right?

swapnili: Yes we cannot predict what user wants :) and my point was not even in that direction. We can only do some sanity checks which requires the setup step to be completed. 
Ok lets take discussion back to the options,
1. Remove the bool and assume user realized() the device and just continue with drive() or poll().
2. Keep bool and always realized() in drive() and poll()
Let me know which works for you.

tmakatos: I'm fine with either options, @jlevon what do you think?

jlevon: I think what Swapnil has here already is fine personally.

tmakatos (approved):

lib/private.h
tmakatos: OK

swapnili:
@tmakatos Fixed the typo and also added some unit tests more can be added but maybe for later.

test/unit-tests.c
```diff
@@ -325,6 +325,37 @@ test_process_command_free_passed_fds(void **state __attribute__((unused)))
...
+
+    assert_int_equal(0, vfu_realize_ctx(&vfu_ctx));
+    assert_true(vfu_ctx.realized);
+    cfg_reg = &vfu_ctx.reg_info[VFU_PCI_DEV_CFG_REGION_IDX];
+    assert_int_equal(VFU_REGION_FLAG_RW, cfg_reg->flags);
+    assert_int_equal(PCI_CFG_SPACE_SIZE, cfg_reg->size);
+    assert_non_null(vfu_ctx.pci.config_space);
+    assert_non_null(vfu_ctx.irqs);
+    assert_null(vfu_ctx.pci.caps);
+}
```
tmakatos, line 346: This test is a good start. It can be improved by checking a few extra things, I suppose this can be left for future work.

test/unit-tests.c
```diff
@@ -325,6 +325,37 @@ test_process_command_free_passed_fds(void **state __attribute__((unused)))
...
+
+static void
+test_attach_ctx(void **state __attribute__((unused)))
+{
+    vfu_ctx_t vfu_ctx = {
+        .trans = &sock_transport_ops,
+        .fd = 111
+    };
+
+    assert_int_equal(-1, vfu_attach_ctx(&vfu_ctx));
```
tmakatos, line 356: We must check that `vfu_attach_ctx` calls the transport attach, so we need to set `vfu_ctx->trans->attach` to some dummy function that returns a known value and then ensure that the dummy function is called with the expected arguments and that it returns a known value.

tmakatos (approved):
The tests are a good start, could be extended to be more elabore, this can be done separately.

swapnili:
Sure will add them in separate patch.

PR #178, opened 2020-12-08
[draft] Drop vfu_ctx_drive() and use vfu_ctx_poll()

swapnili:
Updated vfu_ctx_poll() to also handle blocking ctx.
Instead of having seperate functions for blocking and
non-blocking ctx, better to have one.
This way user can call same set of functions for both cases.

### DIFF
--- a/include/libvfio-user.h
+++ b/include/libvfio-user.h
@@ -438,10 +438,10 @@ vfu_destroy_ctx(vfu_ctx_t *vfu_ctx);
 /**
  * Polls the vfu_ctx and processes the command recieved from client.
  * - Blocking vfu_ctx:
- *   Blocks until new request is recieved from client and continues processing
+ *   Blocks until new request is received from client and continues processing
  *   the requests. Exits only in case of error or if the client disconnects.
  * - Non-blocking vfu_ctx(LIBVFIO_USER_FLAG_ATTACH_NB):
- *   Processes one request from client if its available, otherwise it
+ *   Processes one request from client if it's available, otherwise it
  *   immediatelly returns and the caller is responsible for periodically
  *   calling again.
  *
--- a/include/libvfio-user.h
+++ b/include/libvfio-user.h
@@ -446,7 +446,7 @@ vfu_destroy_ctx(vfu_ctx_t *vfu_ctx);
  * @returns 0 on success, -errno on failure.
  */
 int
-vfu_ctx_poll(vfu_ctx_t *vfu_ctx);
+vfu_run_ctx(vfu_ctx_t *vfu_ctx);
 
 /**
  * Triggers an interrupt.
@@ -603,7 +603,7 @@ vfu_realize_ctx(vfu_ctx_t *vfu_ctx);
 
 /*
  * Attempts to attach to the transport. Attach is mandatory before
- * vfu_ctx_poll() and is non blocking if context is created
+ * vfu_run_ctx() and is non blocking if context is created
  * with LIBVFIO_USER_FLAG_ATTACH_NB flag.
  * Returns client's file descriptor on success and -1 on error. If errno is
  * set to EAGAIN or EWOULDBLOCK then the transport is not ready to attach to and
--- a/include/libvfio-user.h
+++ b/include/libvfio-user.h
@@ -437,9 +437,13 @@ vfu_destroy_ctx(vfu_ctx_t *vfu_ctx);
 
 /**
  * Polls the vfu_ctx and processes the command recieved from client.
- * Non-blocking poll if vfu_ctx is created with LIBVFIO_USER_FLAG_ATTACH_NB,
- * otherwise blocking. With non-blocking poll application can periodically poll
- * the context directly from one of its own threads.
+ * - Blocking vfu_ctx:
+ *   Blocks until new request is recieved from client and continues processing
+ *   the requests. Exits only in case of error or if the client disconnects.
+ * - Non-blocking vfu_ctx(LIBVFIO_USER_FLAG_ATTACH_NB):
+ *   Processes one request from client if its available, otherwise it
+ *   immediatelly returns and the caller is responsible for periodically
+ *   calling again.
  *
  * @vfu_ctx: The libvfio-user context to poll
  *
--- a/include/libvfio-user.h
+++ b/include/libvfio-user.h
@@ -436,24 +436,10 @@ void
 vfu_destroy_ctx(vfu_ctx_t *vfu_ctx);
 
 /**
- * Once the vfu_ctx is configured vfu_ctx_drive() drives it. This function waits
- * for commands coming from the client, and processes them in a loop.
- *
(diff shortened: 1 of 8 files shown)


lib/libvfio-user.c
```diff
@@ -1162,34 +1162,18 @@ vfu_realize_ctx(vfu_ctx_t *vfu_ctx)
...
 
 int
-vfu_ctx_drive(vfu_ctx_t *vfu_ctx)
+vfu_ctx_poll(vfu_ctx_t *vfu_ctx)
 {
     int err;
+    int blocking = vfu_ctx->flags & LIBVFIO_USER_FLAG_ATTACH_NB ? 0 : 1;
 
-    if (vfu_ctx == NULL || !vfu_ctx->realized) {
+    if (!vfu_ctx->realized) {
```
tmakatos, line 1170: I think we should have at least `assert(vfu_ctx != NULL)`.

lib/libvfio-user.c
```diff
@@ -1162,34 +1162,18 @@ vfu_realize_ctx(vfu_ctx_t *vfu_ctx)
 }
 
 int
-vfu_ctx_drive(vfu_ctx_t *vfu_ctx)
+vfu_ctx_poll(vfu_ctx_t *vfu_ctx)
 {
     int err;
+    int blocking = vfu_ctx->flags & LIBVFIO_USER_FLAG_ATTACH_NB ? 0 : 1;
```
tmakatos, line 1168: `bool blocking = !(vfu_ctx->flags & LIBVFIO_USER_FLAG_ATTACH_NB` is cleaner IMO.

tmakatos (approved):
minor nits

tmakatos:
Also, some unit tests would be good.

lib/libvfio-user.c
swapnili: Ack

lib/libvfio-user.c
swapnili: Ack

swapnili:
Sure, also added some UT's

include/libvfio-user.h
```diff
@@ -436,24 +436,10 @@ void
...
- * a thread and making a blocking call to vfu_ctx_drive(). Instead, the
- * application can periodically poll the context directly from one of
- * its own threads.
- *
- * This is only allowed when LIBVFIO_USER_FLAG_ATTACH_NB is specified during
- * creation.
+ * Polls the vfu_ctx and processes the command recieved from client.
+ * Non-blocking poll if vfu_ctx is created with LIBVFIO_USER_FLAG_ATTACH_NB,
+ * otherwise blocking. With non-blocking poll application can periodically poll
+ * the context directly from one of its own threads.
```
tmakatos, line 442: I just realized that the name of the function can be misleading: in blocking mode the function keeps processing requests until failure or the client closes the connection. In this case the semantics of the function are very different from the standard select(2) and poll(2), which tell you that some file descriptor is reading and it's up to the caller to consume the file descriptor data. I think a more suitable name would be `process_request` or something similar. I should have realized that earlier.

tmakatos (changes requested):

include/libvfio-user.h
swapnili: Agreed this needs to be renamed suitably. I am fine with vfu_process_request(). @jlevon any better suggestions?

jlevon: "process_request" implies it does only one request.
How about vfu_ctx_run() ?

swapnili: Yes that also sounds good. Maybe we can call it as vfu_run_ctx(), to be in sync with vfu_*_ctx.

swapnili: I missed the likes, no notifications for emoji's :)
Renamed vfu_ctx_poll() to vfu_run_ctx(), @tmakatos have another look

lib/libvfio-user.c
```diff
@@ -1162,34 +1164,20 @@ vfu_realize_ctx(vfu_ctx_t *vfu_ctx)
...
 
 int
-vfu_ctx_drive(vfu_ctx_t *vfu_ctx)
+vfu_run_ctx(vfu_ctx_t *vfu_ctx)
 {
     int err;
+    bool blocking = !(vfu_ctx->flags & LIBVFIO_USER_FLAG_ATTACH_NB);
 
-    if (vfu_ctx == NULL || !vfu_ctx->realized) {
+    assert(vfu_ctx != NULL);
```
tmakatos, line 1172: the assert must be before the assignment

include/libvfio-user.h
```diff
@@ -436,31 +436,17 @@ void
...
-
-/**
- * Polls, without blocking, an vfu_ctx. This is an alternative to using
- * a thread and making a blocking call to vfu_ctx_drive(). Instead, the
- * application can periodically poll the context directly from one of
- * its own threads.
- *
- * This is only allowed when LIBVFIO_USER_FLAG_ATTACH_NB is specified during
- * creation.
+ * Polls the vfu_ctx and processes the command recieved from client.
```
tmakatos, line 439: I think the documentation isn't clear enough. We must clearly state that in blocking mode `vfu_ctx_run` will continue processing commands indefinitely, blocking until new requests are available, and only exits in case of error or if the client disconnects. In non-blocking mode the function will process one request if it's available, otherwise it immediatelly returns and the caller is responsible for periodically calling this function.

tmakatos (changes requested):

lib/libvfio-user.c
swapnili: Thanks. I overlooked it.

include/libvfio-user.h
swapnili: Sure

include/libvfio-user.h
```diff
@@ -437,9 +437,13 @@ vfu_destroy_ctx(vfu_ctx_t *vfu_ctx);
 
 /**
  * Polls the vfu_ctx and processes the command recieved from client.
- * Non-blocking poll if vfu_ctx is created with LIBVFIO_USER_FLAG_ATTACH_NB,
- * otherwise blocking. With non-blocking poll application can periodically poll
- * the context directly from one of its own threads.
+ * - Blocking vfu_ctx:
+ *   Blocks until new request is recieved from client and continues processing
```
tmakatos, line 441: s/recieved/received

include/libvfio-user.h
```diff
@@ -437,9 +437,13 @@ vfu_destroy_ctx(vfu_ctx_t *vfu_ctx);
...
+ *   the requests. Exits only in case of error or if the client disconnects.
+ * - Non-blocking vfu_ctx(LIBVFIO_USER_FLAG_ATTACH_NB):
+ *   Processes one request from client if its available, otherwise it
```
tmakatos, line 444: s/its/it's

tmakatos (approved):
Good to go, minor nits.

tmakatos (approved):

PR #184, opened 2020-12-11
allow over-riding "cmake"

jlevon:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -41,6 +41,8 @@ ifeq ($(VERBOSE),)
     MAKEFLAGS += -s
 endif
 
+CMAKE = cmake
+
 BUILD_DIR_BASE = $(CURDIR)/build
 BUILD_DIR = $(BUILD_DIR_BASE)/$(BUILD_TYPE)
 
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -41,6 +41,8 @@ ifeq ($(VERBOSE),)
     MAKEFLAGS += -s
 endif
 
+CMAKE = cmake
```
swapnili, line 44: Don't we need to check if $CMAKE is already defined?

jlevon: can you explain what you mean a little further?

swapnili: If we want to take CMAKE from environment then don't we need to do following,
ifndef CMAKE
 CMAKE = cmake
endif

I am not sure what is the meaning of over-ride cmake here means. Maybe I totally missed the point.
Can you please elaborate. Thanks

jlevon: This is a make variable, so you can over-ride via the usual mechanism:

```
$ make CMAKE=fewfwe
/bin/sh: line 1: fewfwe: command not found
make: *** [/home/john.levon/src/libvfio-user/build/dbg/Makefile] Error 127
```

swapnili: Ok I thought we wanted to use the env provided CMAKE value.
e.g: make CMAKE=/mybin/cmake

swapnili (approved):

PR #187, opened 2020-12-11
return region capabilities a la VFIO

tmakatos:
This patch returns region capabilities the same way VFIO does: if argsz
is not large enough then it returns only region info and sets argsz to
what it should be in order to fit the capabilities, the client then
retries with a large enough argsz. The protocol specification has been
updated as well.

Plus unit tests.

### DIFF
--- a/samples/client.c
+++ b/samples/client.c
@@ -329,7 +329,7 @@ get_device_info(int sock, struct vfio_device_info *dev_info)
         errx(EXIT_FAILURE, "failed to get device info: %s", strerror(-ret));
     }
 
-    if (dev_info->num_regions != 2) {
+    if (dev_info->num_regions != 10) {
         errx(EXIT_FAILURE, "bad number of device regions %d",
              dev_info->num_regions);
     }
--- a/samples/client.c
+++ b/samples/client.c
@@ -260,43 +260,68 @@ get_region_vfio_caps(int sock, size_t cap_sz)
     return migr;
 }
 
+static void
+do_get_device_region_info(int sock, struct vfio_region_info *region_info)
+{
+    int ret = vfu_msg(sock, 0,
+                  VFIO_USER_DEVICE_GET_REGION_INFO,
+                  region_info, region_info->argsz,
+                  NULL,
+                  region_info, region_info->argsz);
+    if (ret < 0) {
+        errx(EXIT_FAILURE, "failed to get device region info: %s",
+                strerror(-ret));
+    }
+}
+
+static bool
+get_device_region_info(int sock, uint32_t index)
+{
+    struct vfio_region_info *region_info;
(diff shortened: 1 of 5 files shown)


tmakatos:
I forgot to fix the client/server sample, I'll do that in a separate commit in this PR. The existing commit in this PR can be reviewed now, it shouldn't be affected by the client/server fix.

tmakatos:
We also don't pass FDs according to the protocol specification (https://github.com/nutanix/libvfio-user/issues/188).

samples/client.c
```diff
@@ -256,47 +243,70 @@ get_region_vfio_caps(int sock, size_t cap_sz)
...
+        errx(EXIT_FAILURE, "failed to get device region info: %s",
+                strerror(-ret));
+    }
+}
+
+static bool
+get_device_region_info(int sock, uint32_t index)
+{
+    struct vfio_region_info *region_info;
+    size_t cap_sz UNUSED;
```
jlevon, line 267: nit, why is this UNUSED?

samples/server.c
```diff
@@ -434,13 +436,17 @@ int main(int argc, char *argv[])
...
     }
 
+    /*
+     * Setup BAR1 to be 3 pages in size where only the first and the last pages
+     * are mappable.
+     */
     struct iovec mmap_areas[] = {
-        { .iov_base  = (void*)0x400, .iov_len = 0x400 },
-        { .iov_base  = (void*)0x2000, .iov_len = 0x400 }
+        { .iov_base  = (void*)0, .iov_len = 0x1000},
```
jlevon, line 444: nits: NULL? space before the trailing }

jlevon (changes requested):
A couple of nits only, but could you also please check in the latest spec into docs/ or something?

samples/client.c
tmakatos: > nit, why is this UNUSED?

accidental, will fix

samples/server.c
tmakatos: fixed

tmakatos:
@jlevon fixed client/server samples, please take a look at the new commits

tmakatos:
> A couple of nits only, but could you also please check in the latest spec into docs/ or something?

It's in https://github.com/tmakatos/qemu/tree/vfio-user/v8.

jlevon (approved):

jlevon:
I know it's in the qemu branch, I'm suggesting it could easily live here as the official place, at least until it's merged. It's also useful as we change the code for the spec, to see the spec change at the same time.

tmakatos:
> I know it's in the qemu branch, I'm suggesting it could easily live here as the official place, at least until it's merged. It's also useful as we change the code for the spec, to see the spec change at the same time.

Sure will add it.

Issue #112, opened 2020-11-23
avoid dynamically allocating memory


tmakatos:
We've added `max_msg_size` in v7 (not sent yet), see https://libmuser.slack.com/archives/C01AFGCSPTR/p1605195853082400 for more details.

jlevon:
I think we can close it. The only performance-relevant path where this could really matter would be ioregionfd/ioeventfd handling, and that doesn't exist right now.

PR #191, opened 2020-12-14
don't return memory allocated on the stack

### LINKED ISSUE #189: commit a3ba81765daeffaaa2e9e59d49d49ae4438b1b59 broke client/server sample
```
$ build/dbg/samples/client foo
client: failed to read from region -559038737 0-0x3: Invalid argument
devinfo: flags 0, num_regions 0, num_irqs 0
client: could not find migration region
```

tmakatos:

fixes #189

### DIFF
--- a/lib/libvfio-user.c
+++ b/lib/libvfio-user.c
@@ -690,7 +690,7 @@ handle_region_access(vfu_ctx_t *vfu_ctx, uint32_t size, uint16_t cmd,
     if (cmd == VFIO_USER_REGION_READ) {
         *len += region_access->count;
     }
-    *data = malloc(*len);
+    *data = calloc(1, *len);
     if (*data == NULL) {
         return -ENOMEM;
     }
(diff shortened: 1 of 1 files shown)


lib/libvfio-user.c
```diff
@@ -878,8 +878,8 @@ exec_command(vfu_ctx_t *vfu_ctx, struct vfio_user_header *hdr, size_t size,
              bool *free_iovec_data)
```
swapnili, line 878: This function can be simplified. 
e.g Looks like there is no need to pass _iovecs,.

swapnili (approved):

jlevon (approved):
This looks OK, but the iovec handling is getting a bit complicated here.
We might want to look at cleaning it up later.
(One approach is to keep a record of every allocation in the vfu_ctx, then just walk through and free that list at the end. But we still have the iovec pointer handling stuff...)

tmakatos:
I agree it's getting a bit complicated, I've deliberately left it like that since we're planning to simplify it once v7 is approved (https://github.com/nutanix/libvfio-user/issues/112).

PR #197, opened 2020-12-14
clean up the API header file

jlevon:
This is almost entirely re-ordering: first the basic lifecycle things, then
vfu_setup_*() group, then handlers and helpers, and finally PCI handling.

### DIFF
--- a/include/libvfio-user.h
+++ b/include/libvfio-user.h
@@ -57,6 +57,8 @@ extern "C" {
 #define LIB_VFIO_USER_MAJOR 0
 #define LIB_VFIO_USER_MINOR 1
 
+#define VFU_DMA_REGIONS  0x10
+
 // FIXME: too common a name?
 typedef uint64_t dma_addr_t;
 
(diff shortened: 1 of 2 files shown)


lib/cap.c
```diff
@@ -40,6 +40,8 @@
 #include "libvfio-user.h"
 #include "cap.h"
 
+#define VFU_MAX_CAPS 48
```
tmakatos, line 43: Why  change the original way we calculated it? `(PCI_CFG_SPACE_SIZE - PCI_STD_HEADER_SIZEOF) / PCI_CAP_SIZEOF`

tmakatos (approved):
I haven't looked at it very carefully, looks good.

lib/cap.c
jlevon: It needs to go away anyway as it's not right, but yeah, I should restore that.

include/libvfio-user.h
```diff
@@ -57,6 +57,8 @@ extern "C" {
 #define LIB_VFIO_USER_MAJOR 0
 #define LIB_VFIO_USER_MINOR 1
 
+#define VFU_DMA_REGIONS  0x10
```
swapnili, line 60: I know it is not related to this patch, but from where this limit came. Maybe @tmakatos knows.

swapnili (approved):

include/libvfio-user.h
tmakatos: It's what QEMU typically uses in our tests, we should definitely find a better way of determining the number of DMA regions.

jlevon: #202

PR #209, opened 2020-12-16
enable ERR and REQ IRQs by default

tmakatos:
VFIO in QEMU seems to expect this


jlevon (changes requested):
Shouldn't we document this in the header?

tmakatos:
> Shouldn't we document this in the header?

we should

tmakatos:
@jlevon fixed, please take another look

jlevon (approved):

PR #210, opened 2020-12-16
fix clang build

jlevon:
Also add clang to pull request build checks.

### DIFF
--- a/samples/client.c
+++ b/samples/client.c
@@ -155,7 +155,7 @@ recv_version(int sock, int *server_max_fds, size_t *pgsize)
     }
 
     if (sversion->major != LIB_VFIO_USER_MAJOR) {
-        errx(EXIT_FAILURE, "unsupported server major %hu (must be %hu)",
+        errx(EXIT_FAILURE, "unsupported server major %hu (must be %u)",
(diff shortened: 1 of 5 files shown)


samples/client.c
```diff
@@ -437,21 +437,25 @@ access_region(int sock, int region, bool is_write, uint64_t offset,
...
-        recv_data_len = sizeof(recv_data.region_access);
+        recv_data_len = sizeof(*recv_data);
     } else {
         op = VFIO_USER_REGION_READ;
         nr_send_iovecs = 2;
-        recv_data_len = sizeof(recv_data);
+        recv_data_len = sizeof (*recv_data) + data_len;
+    }
+
+    recv_data = calloc(1, recv_data_len);
```
tmakatos, line 455: I think we can get away with using `alloca` here since this is a sample, we avoid the `free` later as well.

tmakatos (approved):
Optional suggestion made, otherwise good to go.

samples/client.c
jlevon: I'm not a big fan of alloca(). I know it doesn't matter for the samples as you say, but since I've already written it, I think I'll leave it :)

PR #212, opened 2020-12-16
add copy of vfio-user protocol specification (v8)

tmakatos:


jlevon (approved):
thanks!

PR #213, opened 2020-12-17
Reorg vfu_create_ctx() and also create dma_controller

swapnili:
dma_controller_add_region() adds dma_region only if
vfu_ctx->dma is created.

vfu_setup_device_dma_cb() should only set callbacks.

### DIFF
--- a/lib/libvfio-user.c
+++ b/lib/libvfio-user.c
@@ -1334,7 +1334,7 @@ vfu_create_ctx(vfu_trans_t trans, const char *path, int flags, void *pvt,
 
     vfu_ctx->uuid = strdup(path);
     if (vfu_ctx->uuid == NULL) {
-        err = errno;
+        err = -errno;
         goto err_out;
     }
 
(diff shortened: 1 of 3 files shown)


lib/libvfio-user.c
```diff
@@ -1334,7 +1335,7 @@ vfu_create_ctx(vfu_trans_t trans, const char *path, int flags, void *pvt,
     vfu_ctx->uuid = strdup(path);
     if (vfu_ctx->uuid == NULL) {
         err = errno;
```
tmakatos, line 1337: `-errno` (since you're touching this code)

tmakatos (changes requested):
According to the API documentation the user must explicitly ask for a DMA controlling by registering a callabck. Not registering a callback means that the device isn't going to perform DMA to the guest, so DMA map/unmap commands are simply ack'ed. Why are we changing this behavior?

swapnili:
Yes you are right there is no need to keep the maps if user do not want them.
From the documentation of vfu_setup_device_dma_cb() it was not clear, I will add some explanation there.
I think if vfu_setup_device_dma_cb() is called we should unconditionally dma_controller_create()?

lib/libvfio-user.c
swapnili: Ack

tmakatos:
> I think if vfu_setup_device_dma_cb() is called we should unconditionally dma_controller_create()?

Exactly.

tmakatos (approved):

PR #233, opened 2021-01-13
Misc fixes for DMA_MAP region prot

swapnili:
1. Validate prot passed in vfu_addr_to_sg()
2. Let user know region prot via vfu_unmap_dma_cb_t


lib/libvfio-user.c
```diff
@@ -1399,13 +1400,18 @@ inline int
...
 
     if (unlikely(vfu_ctx->unmap_dma == NULL)) {
         errno = EINVAL;
         return -1;
     }
-    return dma_addr_to_sg(vfu_ctx->dma, dma_addr, len, sg, max_sg, prot);
+
+    ret = dma_addr_to_sg(vfu_ctx->dma, dma_addr, len, sg, max_sg, prot);
+
+    return ret < 0 ? ERROR(-ret) : ret;
```
tmakatos, line 1414: Again this is not correct according to `dma_addr_to_sg`'s return semantics.

test/unit-tests.c
```diff
@@ -101,6 +101,21 @@ test_dma_map_without_fd(void **state __attribute__((unused)))
     assert_int_equal(0, handle_dma_map_or_unmap(&vfu_ctx, size, true, &fd, 0, &r));
 }
 
+static void dma_map_cb(UNUSED vfu_ctx_t *vfu_ctx, uint64_t iova, uint64_t len,
+                uint32_t prot)
+{
+    if (iova == 0xcafebabe) {
+        assert(len == 0x1000);
+        assert(prot == (PROT_READ|PROT_WRITE));
```
tmakatos, line 109: nit: not need for parenthesis plus space between `|` operator

test/unit-tests.c
```diff
@@ -101,6 +101,21 @@ test_dma_map_without_fd(void **state __attribute__((unused)))
     assert_int_equal(0, handle_dma_map_or_unmap(&vfu_ctx, size, true, &fd, 0, &r));
 }
 
+static void dma_map_cb(UNUSED vfu_ctx_t *vfu_ctx, uint64_t iova, uint64_t len,
+                uint32_t prot)
+{
+    if (iova == 0xcafebabe) {
+        assert(len == 0x1000);
```
tmakatos, line 108: Better use `assert_int_equal`?

tmakatos (changes requested):
Looks good apart from minor nits plus `dma_addr_to_sg` return semantics.

lib/dma.h
```diff
@@ -184,11 +184,18 @@ _dma_mark_dirty(const dma_controller_t *dma, const dma_memory_region_t *region,
...
 
-static inline void
+static inline int
 dma_init_sg(const dma_controller_t *dma, dma_sg_t *sg, dma_addr_t dma_addr,
             uint32_t len, int prot, int region_index)
 {
     const dma_memory_region_t *const region = &dma->regions[region_index];
+    uint32_t reg_prot = region->prot;
+
+    if ((reg_prot == PROT_NONE) ||
```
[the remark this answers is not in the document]
swapnili: PROT_NONE means 'Pages may not be accessed'. If we return back the sg_list and user tries to touch it, then it will be a segfault. I think user is calling this API in order to access the region so instead better to return failure.

lib/dma.h
```diff
@@ -184,11 +184,18 @@ _dma_mark_dirty(const dma_controller_t *dma, const dma_memory_region_t *region,
...
+        ((prot & PROT_WRITE) && !(region->prot & PROT_WRITE))) {
```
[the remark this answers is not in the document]
swapnili: IIUC PROT_WRITE implies PROT_READ there is nothing like write-only protection for mmap()

lib/dma.h
```diff
@@ -225,7 +234,11 @@ dma_addr_to_sg(const dma_controller_t *dma,
     if (likely(max_sg > 0 && len > 0 &&
                dma_addr >= region->dma_addr && dma_addr + len <= region_end &&
                region_hint < dma->nregions)) {
-        dma_init_sg(dma, sg, dma_addr, len, prot, region_hint);
+        ret = dma_init_sg(dma, sg, dma_addr, len, prot, region_hint);
+        if (ret < 0) {
+            return ret;
```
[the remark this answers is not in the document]
swapnili: Ack

test/unit-tests.c
swapnili: w/o parenthesis compiler warnings (-Werror=parentheses). But anyways now with assert_int_equal() it doesn't matter.

test/unit-tests.c
swapnili: Ok

lib/dma.h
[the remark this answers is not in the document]
tmakatos: Correct, however this function doesn't touch pages, it just translates the DMA address.

lib/dma.h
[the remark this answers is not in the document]
tmakatos: ok

include/libvfio-user.h
```diff
@@ -454,11 +455,12 @@ vfu_irq_message(vfu_ctx_t *vfu_ctx, uint32_t subindex);
  * @len: size of memory to be mapped
  * @sg: array that receives the scatter/gather entries to be mapped
  * @max_sg: maximum number of elements in above array
- * @prot: protection as define in <sys/mman.h>
+ * @prot: protection as defined in <sys/mman.h>
  *
  * @returns the number of scatter/gather entries created on success, and on
  * failure:
- *  -1:         if the GPA address span is invalid, or
+ *  -1:         if the GPA address span is invalid or
```
tmakatos, line 462: We need to clear `errno`  if the `if` is taken in L407, so that the caller can figure out what `-1` exactly means. We should also document it.

lib/dma.h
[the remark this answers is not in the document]
tmakatos: This sounds like a usage error. Plus, this function is about translating DMA addresses to local addresses, `prot` signifies the intention, it doesn't touch any memory.

lib/dma.h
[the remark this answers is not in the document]
tmakatos: I guess you're right, that's probably an  x86 limitation.

lib/dma.h
```diff
@@ -206,7 +216,9 @@ dma_init_sg(const dma_controller_t *dma, dma_sg_t *sg, dma_addr_t dma_addr,
  * Returns:
  * - On success, number of scatter gather entries created.
  * - On failure:
- *     -1 if the dma address span is invalid
+ *     -1 if
+ *          - the dma address span is invalid
```
tmakatos, line 220: s/dma/DMA

lib/dma.h
```diff
@@ -206,7 +216,9 @@ dma_init_sg(const dma_controller_t *dma, dma_sg_t *sg, dma_addr_t dma_addr,
...
+ *          - prot violation (errno=EACCES)
```
tmakatos, line 221: s/prot/protection

test/unit-tests.c
```diff
@@ -105,12 +105,12 @@ static void dma_map_cb(UNUSED vfu_ctx_t *vfu_ctx, uint64_t iova, uint64_t len,
                 uint32_t prot)
 {
     if (iova == 0xcafebabe) {
-        assert(len == 0x1000);
-        assert(prot == (PROT_READ|PROT_WRITE));
+        assert_int_equal(len, 0x1000);
```
tmakatos, line 108: nit: constants are usually the 1st argument to the assert functions

tmakatos (changes requested):
minor comments, looks good otherwise

test/unit-tests.c
```diff
@@ -806,12 +806,14 @@ test_dma_addr_to_sg(void **state __attribute__((unused)))
     assert_true(sg.mappable);
 
     r->prot = PROT_NONE;
-    assert_int_equal(-EACCES,
+    assert_int_equal(-1,
```
tmakatos, line 809: We should also have a unit test where we call `dma_addr_to_sg` which fails because a region is not found (must return -1  but `errno` must be 0).

include/libvfio-user.h
tmakatos: > Do you mean the if at lib/dma.c:407?
Yes.
> If yes then that line wont be reached in case of EACESS as the loop will break at L387
I'm referring to a scenario where the `if` is taken in L407 or L411 (the `while` loop won't have broken in L387). In this case we return `-1` but the contents of `errno` are undefined (who knows what syscall got executed earlier), we need to clear to `0`.

tmakatos:
Looks good, just one comment about clearing `errno` in `_dma_addr_sg_split`.

include/libvfio-user.h
swapnili: Ok clearing errno, but I do not see anything which will set errno.

tmakatos: `errno` might contain whatever the previous syscall did, e.g.
```
open(...); /* failed or some reason with errno == ENOENT */
dma_init_sg(...) /* fails because the region wasn't found, no errno still contains ENOENT */

tmakatos (approved):

PR #235, opened 2021-01-18
[DRAFT RFC] ioeventfd/ioregionfd support

jlevon:
Thoughts before sharing more widely?


jlevon:
I thought about using GET_REGION_INFO with a CAP extension, but decided it against it for two reasons:

1 - unlike the others, there's none of this in real vfio
2 - with the above, there's no way for the server to know if ioeventfd/ioregionfd is actually used/requested by the client (it would just have to close those fds). An explicit call is much clearer for this

docs/vfio-user.rst
```diff
@@ -1127,6 +1140,111 @@ client must write data on the same order and transction size as read.
...
+|            | +-----+-----------------------------+ |
++------------+--------+------------------------------+
+| index      | 24     | 4                            |
++------------+--------+------------------------------+
+| nr_subregions     | 28     | number of subregions in array |
++------------+--------+------------------------------+
+| subregions | 32     | array of subregions          |
++------------+--------+------------------------------+
+
+* *index* is the index of memory region being queried, it is the only field
```
tmakatos, line 1197: nit: the order of each element in the list should match the order in the definition above

jlevon: done, thanks

docs/vfio-user.rst
```diff
@@ -1127,6 +1140,111 @@ client must write data on the same order and transction size as read.
...
+
+struct vfio_user_subregion_info {
+    // offset within region (not GPA)
+    u64 offset;
+    // size of sub-region (perhaps zero depending on flags below)
+    u64 size;
+    // index into ancillary data fd array
+    int fd_index;
+    // ioeventfd | ioregionfd
+    u32 subregion_type;
```
tmakatos, line 1225: Won't the type be the same for all array entries in a response?

docs/vfio-user.rst
```diff
@@ -1127,6 +1140,111 @@ client must write data on the same order and transction size as read.
...
+
+    // FIXME: if ioregionfd subsumes ioeventfd, we could potentially just have
+    // the one struct; ioregionfd-less systems could fall back to ioeventfd
+    // interfaces
+    union {
```
tmakatos, line 1230: I don't understand most of these fields, could you elaborate?

docs/vfio-user.rst
```diff
@@ -1127,6 +1140,111 @@ client must write data on the same order and transction size as read.
...
++------------+--------+------------------------------+
+| nr_subregions     | 28     | number of subregions in array |
++------------+--------+------------------------------+
+| subregions | 32     | array of subregions          |
++------------+--------+------------------------------+
+
+* *index* is the index of memory region being queried, it is the only field
+  that is required to be set in the command message.
+* *argsz* is the size of the VFIO region IOFD info structure plus the
+  size of the subregion array (such that each array entry is defined by this
```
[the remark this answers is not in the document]
jlevon: yeah, clarified in last push

tmakatos:
> I thought about using GET_REGION_INFO with a CAP extension, but decided it against it for two reasons:
> 
> 1 - unlike the others, there's none of this in real vfio

We could introduce a new ` VFIO_REGION_TYPE_VFIO_USER` type and then freely introduce new subtypes, assuming it would be accepted upstream. This way if we want to add more attributes to regions we won't have to introduce new commands.

> 2 - with the above, there's no way for the server to know if ioeventfd/ioregionfd is actually used/requested by the client (it would just have to close those fds). An explicit call is much clearer for this

We have the same problem with sparse maps (the client will receive FDs for sparse regions and might not necessarily mmap them), although in the server it's just one FD.

docs/vfio-user.rst
```diff
@@ -1127,6 +1140,111 @@ client must write data on the same order and transction size as read.
...
+| index      | 24     | 4                            |
++------------+--------+------------------------------+
+| nr_subregions     | 28     | number of subregions in array |
++------------+--------+------------------------------+
+| subregions | 32     | array of subregions          |
++------------+--------+------------------------------+
+
+* *index* is the index of memory region being queried, it is the only field
+  that is required to be set in the command message.
+* *argsz* is the size of the VFIO region IOFD info structure plus the
```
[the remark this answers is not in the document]
jlevon: I missed out copying that stanza from the other one, thanks.

docs/vfio-user.rst
```diff
@@ -1127,6 +1140,111 @@ client must write data on the same order and transction size as read.
...
+  size of the subregion array (such that each array entry is defined by this
+  size, minus the offset, divided by nr_subregions)
+* nr_subregions is the number of entries in the subregion array
+
+The reply message will additionally include at least one file descriptor in the
```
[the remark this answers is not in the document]
jlevon: No - it's possible to share FDs between eventfds, or FDs between ioregionfds.

docs/vfio-user.rst
```diff
@@ -1127,6 +1140,111 @@ client must write data on the same order and transction size as read.
...
+------------------------
+
+https://www.spinics.net/lists/kvm/msg208139.html
+https://patchwork.kernel.org/project/kvm/patch/1251028605-31977-23-git-send-email-avi@redhat.com/
+
+FIXME rewrite in table format
+
+struct vfio_user_subregion_info {
+    // offset within region (not GPA)
+    u64 offset;
```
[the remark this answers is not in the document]
jlevon: That just seems like it would be confusing?

docs/vfio-user.rst
```diff
@@ -1127,6 +1140,111 @@ client must write data on the same order and transction size as read.
...
+    // size of sub-region (perhaps zero depending on flags below)
+    u64 size;
+    // index into ancillary data fd array
+    int fd_index;
```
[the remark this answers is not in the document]
jlevon: It's about sharing, not ordering.

docs/vfio-user.rst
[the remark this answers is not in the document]
tmakatos: Fair enough.

docs/vfio-user.rst
jlevon: Nope - it's totally possible we'd watch an ioeventfd for one offset, but an ioregion elsewhere

docs/vfio-user.rst
jlevon: I updated the links above so it's clearer where to look for that info

docs/vfio-user.rst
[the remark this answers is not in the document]
tmakatos: I think each time an FD is sent to another process a new FD is created at the other end, e.g. if the server sends FD 10 twice within the same message then the client receives two different FDs, but pointing to the same file description (as if returned by `dup(2)`. I might be wrong here.

docs/vfio-user.rst
tmakatos: That's very cool, so we can mix and match FD types within a region? If so, it's definitely worth explaining earlier.

jlevon:
> > I thought about using GET_REGION_INFO with a CAP extension, but decided it against it for two reasons:
> > 1 - unlike the others, there's none of this in real vfio
> 
> We could introduce a new ` VFIO_REGION_TYPE_VFIO_USER` type and then freely introduce new subtypes, assuming it would be accepted upstream. This way if we want to add more attributes to regions we won't have to introduce new commands.

Yep, possible, would like to see other's opinions too. The fd + ioeventfd + ioregionfd felt like it was getting pretty overloaded. It's already hard to read I think as defined in vfio.

> > 2 - with the above, there's no way for the server to know if ioeventfd/ioregionfd is actually used/requested by the client (it would just have to close those fds). An explicit call is much clearer for this
> 
> We have the same problem with sparse maps (the client will receive FDs for sparse regions and might not necessarily mmap them), although in the server it's just one FD.

That is true, but I think somewhat less of a problem.

docs/vfio-user.rst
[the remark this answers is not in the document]
tmakatos: OK now I understand, we only pass each FD once and use `fd_index`, thanks.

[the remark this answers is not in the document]
jlevon: I think you're suggesting that if (say) we want to share an eventfd for 8 different eventfd offsets, we'd send that single fd 8 times in the ancillary data, so we don't need the index?

Maybe but I'm not sure that's actually simpler than indexing?

docs/vfio-user.rst
jlevon: I might have missed something that means you couldn't do that, but AFAIK yes. Will clarify

docs/vfio-user.rst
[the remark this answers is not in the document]
tmakatos: Indexing is definitely simpler, I get it now 👍

docs/vfio-user.rst
tmakatos: It's not explicitly stated that we can't, so technicially the description is correct as it is. However, I think clarifying it makes me think less hard.

tmakatos (approved):

docs/vfio-user.rst
jlevon: I expanded the text, please take a look

tmakatos (approved):
looks good

docs/vfio-user.rst
```diff
@@ -1127,6 +1140,133 @@ client must write data on the same order and transction size as read.
...
+This command message is sent by the client to the server to query for
+sub-regions of the given region that should be accessed via file descriptors, as
+opposed to the client mmap()ing in the (sub-)region, or using the
+VFIO_USER_REGION_READ/WRITE round-trip.
+
+This message is optional, as these mechanisms are performance features.
+
+In its reponse, the server specifies a set of sub-regions and the requested file
+descriptor notification mechanism to use for that sub-region.  Each sub-region
+in the reponse message may choose to use a different method, as defined below.
```
swapnili, line 1174: response

swapnili (approved):

docs/vfio-user.rst
jlevon: done

docs/vfio-user.rst
```diff
@@ -1127,6 +1140,133 @@ client must write data on the same order and transction size as read.
...
++--------------+------------------------+
+
+This command message is sent by the client to the server to query for
+sub-regions of the given region that should be accessed via file descriptors, as
+opposed to the client mmap()ing in the (sub-)region, or using the
+VFIO_USER_REGION_READ/WRITE round-trip.
+
+This message is optional, as these mechanisms are performance features.
+
+In its reponse, the server specifies a set of sub-regions and the requested file
```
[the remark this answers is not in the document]
jlevon: done

jlevon:
closed in favour of #437

Issue #240, opened 2021-01-22
libvfio-user fails to compile in release mode in Fedora 33


tmakatos:
See https://github.com/spdk/spdk/issues/1754 for mode details.

PR #241, opened 2021-01-22
fix compilation on newer GCC versions

jlevon:
A newer GCC caught that in test_dma_controller_add_region_no_fd(), we were not
allocating enough stack space for the region at the end of the dma_controller_t.

Also add Fedora Rawhide to github CI for this.


jlevon:
Fixes #240

tmakatos (approved):

PR #250, opened 2021-01-22
use names for migration states when logging

tmakatos:


lib/migration.c
```diff
@@ -145,8 +165,9 @@ handle_device_state(vfu_ctx_t *vfu_ctx, struct migration *migr,
     if (!_migr_state_transition_is_valid(migr->info.device_state,
                                               *device_state)) {
         /* TODO print descriptive device state names instead of raw value */
```
jlevon, line 167: nit, stale comment

lib/migration.c
```diff
@@ -56,24 +56,44 @@ struct migration {
     } iter;
 };
 
+struct migr_state_data {
+    __u32 state;
+    char *name;
```
jlevon, line 61: nit, const char *

lib/migration.c
```diff
@@ -56,24 +56,44 @@ struct migration {
     } iter;
 };
 
+struct migr_state_data {
+    __u32 state;
```
jlevon, line 60: Not for this, but why are we using __ or indeed u* at all? We're in userspace without any namespacing issues; we can/should use the formal uint32_t types

lib/migration.c
```diff
@@ -120,7 +140,7 @@ init_migration(const vfu_migration_t * const vfu_migr, int *err)
 static bool
 _migr_state_transition_is_valid(__u32 from, __u32 to)
 {
-    return migr_states[from] & (1 << to);
+    return migr_states[from].state & (1 << to);
```
jlevon, line 143: I just noticed this now: don't we want an equality comparison here?
It's not valid to migrate from VFIO_DEVICE_STATE_STOP to
VFIO_DEVICE_STATE_STOP | VFIO_DEVICE_STATE_RUNNING
but a simple bitwise and allows it.

jlevon (changes requested):

lib/migration.c
tmakatos: ok

lib/migration.c
tmakatos: ok

lib/migration.c
tmakatos: No, it's correct the way it is. The destination state sets only one bit in the `state` bit array. I've implemented a unit test, will push tomorrow.

jlevon: Oh, I mis-read the existing code, I missed that you were shifting 1 by an OR'd value

test/unit-tests.c
```diff
@@ -1102,6 +1102,72 @@ test_vfu_setup_device_dma_cb(void **state __attribute__((unused)))
...
+    assert_true(f(4, 4));
+    assert_false(f(4, 5));
+    assert_true(f(4, 6));
+    assert_false(f(4, 7));
+
+    /*
+     * Transitioning to any other state from the remaining 3 states
+     * (101b - invalid, 110b - error, 111b - invalid)  is invalid.
+     * Transitioning from the error state to the stopped state is possible but
+     * that requires a devices reset, so we don't consider it a valid state
```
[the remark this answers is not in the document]
tmakatos: fixed

jlevon (approved):

PR #263, opened 2021-01-22
introduce function for transitioning vfio-user migration iteration states

tmakatos:
This makes it easier to tell where we change the migration iteration state. In
the future we might also perform additional validations.


lib/migration.c
```diff
@@ -214,11 +222,8 @@ handle_pending_bytes(vfu_ctx_t *vfu_ctx, struct migration *migr,
              * FIXME what happens if data haven't been consumed in the previous
              * iteration? Ask on LKML.
              */
-            if (*pending_bytes == 0) {
-                migr->iter.state = VFIO_USER_MIGR_ITER_STATE_FINISHED;
-            } else {
-                migr->iter.state = VFIO_USER_MIGR_ITER_STATE_STARTED;
-            }
+            migr_state_transition(migr,
```
jlevon, line 225: I think I preferred the if/else to the very long line here

tmakatos: ok

jlevon (approved):

PR #272, opened 2021-01-27
allow device to specify data_offset when resuming

tmakatos:
Handling data_offset and data_size internally is wrong: we can't simply
assume that the migration data should be appending to the migration
region, devices might have their own requirements.

This also requires a way for the  device to return the data_offset, we
do this by making the prepare_data callback applicable in resume state.

Also, allow migration read/write callabcks to return errors.


tmakatos:
BTW @jlevon you already reviewed this on GerritHub.

jlevon (approved):

PR #274, opened 2021-01-27
run client/server sample in CI

tmakatos:


tmakatos:
The number of iterations in the stop-and-copy phase is sometimes 2 and sometimes 3, sounds like a race condition, need to figure out what's wrong.

test/test-client-server.sh
```diff
@@ -0,0 +1,12 @@
...
+set -e
+
+sock="/tmp/vfio-user.sock"
+rm -f ${sock}*
+../samples/server -v ${sock} &
+while [ ! -S ${sock} ]; do
+	sleep 0.1
+done
+../samples/client ${sock} &
+wait $!
```
jlevon, line 12: shouldn't this just be:

wait

tmakatos: The problem with a simple `wait` is that it return the exit code of the last child that exited, so any other children that failed won't cause `wait` to fail, so the CI will succeed. In my tests it was usually the client that was failing. I'll try to find a way to make the whole script fail even if one child fails.

tmakatos: fixed, take another look

jlevon: why do we even need the wait $! ? Can't we run the client synchronously?

tmakatos: That is a very good point!

tmakatos: Fixed, please take another look.

jlevon (approved):

PR #289, opened 2021-01-29
define full PCI Express capability struct

jlevon:
Define the full-size capability as defined in the specification.
Previously, we were defining the structure as in the form used by PCI Express
Integrated Endpoints. It's reasonable to assume, however, that a vfio-user
device is a normal PCI Express Endpoint connected over a Link.

We'll go further, and define the whole structure, including the slot registers
at the end that are usually only used for Ports.

The presumption here is that it can't hurt to use the larger size: the only way
a client could care is if it presumed the next capability was at a particular
offset from this one, and we must hope nothing is that silly.

This also corrects a buffer overflow: cap_size() in fact disagreed with the
original size of our struct pxcap (found via clang's address sanitizer).

### DIFF
--- a/include/pci_caps/px.h
+++ b/include/pci_caps/px.h
@@ -107,6 +107,36 @@ struct pxls {
 } __attribute__((packed));
 _Static_assert(sizeof(struct pxls) == 0x2, "bad PXLS size");
 
+struct pxscap {
+    unsigned int stuff:32;
+} __attribute__((packed));
+_Static_assert(sizeof(struct pxscap) == 0x4, "bad PXSCAP size");
+
+struct pxsc {
+    unsigned int stuff:16;
+} __attribute__((packed));
+_Static_assert(sizeof(struct pxsc) == 0x2, "bad PXSC size");
+
+struct pxss {
+    unsigned int stuff:16;
+} __attribute__((packed));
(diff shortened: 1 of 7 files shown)


include/pci_caps/px.h
```diff
@@ -131,25 +161,106 @@ struct pxdc2 {
...
+    /* Link Control 2 */
+    struct pxlc2 pxlc2;
+    /* Link Status 2 */
+    struct pxls2 pxls2;
+    /* Slot Capabilities 2 */
+    struct pxscap2 pxscap2;
+    /* Slot Control 2 */
+    struct pxsc2 pxsc2;
+    /* Slot Status 2 */
+    struct pxss2 pxss2;
```
tmakatos, line 261: I quickly went through the order and size of each register as a sanity check. Regarding `pxss2`, the spec says:

> 7.8.23. Slot Status 2 Register (Offset 3Ah)
5 This section is a placeholder. There are no capabilities that require this register.
This register must be treated by software as RsvdZ.

Could you point where it states the size?

tmakatos (approved):
Looks good.

include/pci_caps/px.h
jlevon: I think the diagram is the only clue that it's 2 bytes.

PR #298, opened 2021-02-02
fix leaks in unit tests

jlevon:
These aren't interesting, but it lets us run ASAN.

### DIFF
--- a/test/unit-tests.c
+++ b/test/unit-tests.c
@@ -412,6 +412,9 @@ test_realize_ctx(void **state __attribute__((unused)))
     assert_non_null(vfu_ctx.irqs);
     assert_int_equal(0, vfu_ctx.pci.nr_caps);
     assert_int_equal(0, vfu_ctx.pci.nr_ext_caps);
+
+    free(vfu_ctx.irqs);
+    free(vfu_ctx.pci.config_space);
 }
 
 static int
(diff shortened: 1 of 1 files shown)


test/unit-tests.c
```diff
@@ -560,13 +570,20 @@ test_vfu_ctx_create(void **state __attribute__((unused)))
...
+
     assert_int_equal(1, vfu_ctx->irq_count[VFU_DEV_ERR_IRQ]);
     assert_int_equal(1, vfu_ctx->irq_count[VFU_DEV_REQ_IRQ]);
     assert_int_equal(0, vfu_pci_init(vfu_ctx, VFU_PCI_TYPE_CONVENTIONAL,
                         PCI_HEADER_TYPE_NORMAL, 0));
     assert_int_equal(PCI_STD_HEADER_SIZEOF,
                      vfu_pci_add_capability(vfu_ctx, 0, 0, &pm));
     assert_int_equal(0, vfu_realize_ctx(vfu_ctx));
+
+    patch(close);
```
swapnili, line 582: I do not see close() being called in this UT, apart from dma_controller_destroy()(but we do not have DMA regions)
Who can call close() here?

On other note, I think close(vfu_ctx->fd) is missing in vfu_destroy_ctx()

jlevon: This is the failure:

```
1: [ RUN      ] test_vfu_ctx_create
1: [  FAILED  ] test_vfu_ctx_create
1: [  ERROR   ] --- No entries for symbol __wrap_close.
1: /home/john.levon/src/libvfio-user/test/mocks.c:127: error: Could not get value to check parameter fd of function __wrap_close
1: There were no previously declared parameter values for this test.
```

I presume this is from tran->detach() (close_sock())

We can get rid of the vfu_ctx->fd copy I think

swapnili (approved):

PR #299, opened 2021-02-03
close listening socket in vfu_destroy_ctx()

jlevon:
We were forgetting to close vfu_ctx->fd, add a tran callback for this. While
we're there, clean up the tran callbacks somewhat.


swapnili (approved):
Thanks for fixing this.

PR #305, opened 2021-02-03
expose migration region

tmakatos:
This PR depends on some other pending PRs so I'll leave it as a draft, however it would be good to get some initial feedback.

Also, need to implement unit tests.

### DIFF
--- a/include/libvfio-user.h
+++ b/include/libvfio-user.h
@@ -461,19 +461,20 @@ typedef struct {
 
 /**
  * vfu_setup_device_migration provides an abstraction over the migration
- * protocol: the user specifies a set of callbacks which are called in respone
- * to client accesses of the migration region; the migration read/write
- * callbacks are not called. Offsets in callbacks are relative to @data_offset.
+ * protocol: the user specifies a set of callbacks which are called in response
+ * to client accesses of the migration region; the migration region read/write
+ * callbacks are not called after this function call. Offsets in callbacks are
+ * relative to @data_offset.
  * 
  * @vfu_ctx: the libvfio-user context
  * @callbacks: migration callbacks
  * @data_offset: offset in the migration region where data begins.
  *
- * @returns 0 on success, -1 on error, Sets errno.
+ * @returns 0 on success, -1 on error, sets errno.
  */
 int
 vfu_setup_device_migration_callbacks(vfu_ctx_t *vfu_ctx,
-                                     const vfu_migration_callbacks_t * callbacks,
+                                     const vfu_migration_callbacks_t *callbacks,
                                      uint64_t data_offset);
 
 /**
@@ -589,16 +590,16 @@ vfu_dma_write(vfu_ctx_t *vfu_ctx, dma_sg_t *sg, void *data);
 /*
  * Supported PCI regions.
  *
- * Note to developers: in VFIO, each region starts at a terabyte offset
- * (VFIO_PCI_INDEX_TO_OFFSET) and because Linux supports up to 128 TB of
- * virtual memory, there can be up to 128 device regions. PCI regions are fixed
- * and in retrospect this choice has proven to be problematic because devices
- * might contain potentially unused regions. New regions can now be positioned
- * anywhere by using the VFIO_REGION_INFO_CAP_TYPE capability.
- * In vfio-user we don't have this problem because the region index is just an
+ * Note: in VFIO, each region starts at a terabyte offset
+ * (VFIO_PCI_INDEX_TO_OFFSET) and because Linux supports up to 128 TB of user
+ * space virtual memory, there can be up to 128 device regions. PCI regions are
+ * fixed and in retrospect this choice has proven to be problematic because
+ * devices might contain potentially unused regions. New regions can now be
+ * positioned anywhere by using the VFIO_REGION_INFO_CAP_TYPE capability.  In
+ * vfio-user we don't have this problem because the region index is just an
  * identifier: the VMM memory maps a file descriptor that is passed to it and
- * the offset that it memory maps is relative to the beginning of the region,
- * therefore it is something reasonably small. Thus, additional regions can
+ * the mapping offset is derived from the mmap_areas offset value, rather than a
+ * static mapping from region index to offset. Thus, additional regions can
  * have static indexes in vfio-user.
  */
 enum {
--- a/include/libvfio-user.h
+++ b/include/libvfio-user.h
@@ -590,6 +590,18 @@ vfu_dma_write(vfu_ctx_t *vfu_ctx, dma_sg_t *sg, void *data);
 
 /*
  * Supported PCI regions.
+ *
+ * Note to developers: in VFIO, each region starts at a terabyte offset
+ * (VFIO_PCI_INDEX_TO_OFFSET) and because Linux supports up to 128 TB of
+ * virtual memory, there can be up to 128 device regions. PCI regions are fixed
+ * and in retrospect this choice has proven to be problematic because devices
+ * might contain potentially unused regions. New regions can now be positioned
+ * anywhere by using the VFIO_REGION_INFO_CAP_TYPE capability.
+ * In vfio-user we don't have this problem because the region index is just an
+ * identifier: the VMM memory maps a file descriptor that is passed to it and
+ * the offset that it memory maps is relative to the beginning of the region,
+ * therefore it is something reasonably small. Thus, additional regions can
+ * have static indexes in vfio-user.
  */
 enum {
     VFU_PCI_DEV_BAR0_REGION_IDX,
--- a/include/libvfio-user.h
+++ b/include/libvfio-user.h
@@ -250,20 +250,18 @@ typedef ssize_t (vfu_region_access_cb_t)(vfu_ctx_t *vfu_ctx, char *buf,
  * the device. The migration region must contain at the beginning the migration
  * registers (struct vfio_device_migration_info defined in <linux/vfio.h>) and
  * the remaining part of the region can be arbitrarily used by the device
- * implementation. Therefore, the size of the migration region must be at least
- * sizeof(struct vfio_device_migration_info). libvfio-user offers two ways for
- * the migration region to be used:
+ * implementation. The region provided must have at least
+ * vfu_get_migr_register_area_size() bytes available at the start of the region
+ * (this size is guaranteed to be page-aligned). If mmap_areas is given, it
+ * must _not_ include this part of the region.
+ *
+ * libvfio-user offers two ways for the migration region to be used:
  *  1. natively: the device implementation must handle accesses to the
  *      migration registers and migration data via the region callbacks. The
  *      semantics of these registers are explained in <linux/vfio.h>.
  *  2. via the vfu_migration_t callbacks: the device implementation registers
- *      a set of callbacks by calling vfu_setup_device_migration. The device
- *      does not have access to the migration registers, however the migration
- *      region provided must still reserve space for them (see function
- *      vfu_get_migr_regs_size). The region's read/write callbacks are never
- *      called.
- * The migration region can be partially memory mapped, except the migration
- * registers.
+ *      a set of callbacks by calling vfu_setup_device_migration. The region's
+ *      read/write callbacks are never called.
  *
  * @vfu_ctx: the libvfio-user context
  * @region_idx: region index
@@ -287,11 +285,11 @@ vfu_setup_region(vfu_ctx_t *vfu_ctx, int region_idx, size_t size,
                  int fd);
 
 /*
- * Returns the size of the migration registers, which is guaranteed to be page
- * aligned.
+ * Returns the size of the area needed to hold the migration registers at the
+ * beginning of the migration region; guaranteed to be page aligned.
  */
 size_t
-vfu_get_migr_regs_size(void);
+vfu_get_migr_register_area_size(void);
 
 /*
  * Callback function that is called when the guest resets the device.
--- a/include/libvfio-user.h
+++ b/include/libvfio-user.h
@@ -227,7 +227,12 @@ typedef ssize_t (vfu_region_access_cb_t)(vfu_ctx_t *vfu_ctx, char *buf,
  * given callback.  However, the callback can still be invoked, even on a
  * mappable area, if the client chooses to call VFIO_USER_REGION_READ/WRITE.
  *
- * A VFU_PCI_DEV_CFG_REGION_IDX region, corresponding to PCI config space, has
+ * The following regions are special and are explained below:
+ *  - VFU_PCI_DEV_CFG_REGION_IDX,
+ *  - VFU_PCI_DEV_MIGR_REGION_IDX, and
+ *  - VFU_GENERIC_DEV_MIGR_REG_IDX.
+ *
+ * Region VFU_PCI_DEV_CFG_REGION_IDX, corresponding to PCI config space, has
  * special handling:
  *
  *  - the @size argument is ignored: the region size is always the size defined
@@ -239,6 +244,26 @@ typedef ssize_t (vfu_region_access_cb_t)(vfu_ctx_t *vfu_ctx, char *buf,
  *  - if no callback is provided, reads to other areas are a simple memcpy(),
  *    and writes are an error
  *  - otherwise, the callback is expected to handle the access
+ *
+ * Regions VFU_PCI_DEV_MIGR_REGION_IDX and VFU_GENERIC_DEV_MIGR_REG_IDX,
+ * corresponding to the migration region, enable live migration support for
+ * the device. The migration region must contain at the beginning the migration
+ * registers (struct vfio_device_migration_info defined in <linux/vfio.h>) and
+ * the remaining part of the region can be arbitrarily used by the device
+ * implementation. Therefore, the size of the migration region must be at least
+ * sizeof(struct vfio_device_migration_info). libvfio-user offers two ways for
+ * the migration region to be used:
+ *  1. natively: the device implementation must handle accesses to the
+ *      migration registers and migration data via the region callbacks. The
+ *      semantics of these registers are explained in <linux/vfio.h>.
+ *  2. via the vfu_migration_t callbacks: the device implementation registers
+ *      a set of callbacks by calling vfu_setup_device_migration. The device
+ *      does not have access to the migration registers, however the migration
+ *      region provided must still reserve space for them (see function
+ *      vfu_get_migr_regs_size). The region's read/write callbacks are never
+ *      called.
+ * The migration region can be partially memory mapped, except the migration
+ * registers.
  *
  * @vfu_ctx: the libvfio-user context
  * @region_idx: region index
@@ -260,6 +285,13 @@ vfu_setup_region(vfu_ctx_t *vfu_ctx, int region_idx, size_t size,
                  vfu_region_access_cb_t *region_access, int flags,
                  struct iovec *mmap_areas, uint32_t nr_mmap_areas,
                  int fd);
+
+/*
+ * Returns the size of the migration registers, which is guaranteed to be page
+ * aligned.
+ */
+size_t
+vfu_get_migr_regs_size(void);
 
 /*
  * Callback function that is called when the guest resets the device.
@@ -348,6 +380,8 @@ typedef enum {
 } vfu_migr_state_t;
 
 
+#define VFU_MIGR_CALLBACKS_VERS 1
+
 /*
  * Callbacks during the pre-copy and stop-and-copy phases.
  *
@@ -362,6 +396,11 @@ typedef enum {
  * there are no more migration data to be consumed in this iteration).
  */
 typedef struct {
+
+    /*
+     * Set it to VFU_MIGR_CALLBACKS_VERS.
+     */
+    int version;
 
     /* migration state transition callback */
     /* TODO rename to vfu_migration_state_transition_callback */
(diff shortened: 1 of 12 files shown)


include/libvfio-user.h
```diff
@@ -240,6 +245,25 @@ typedef ssize_t (vfu_region_access_cb_t)(vfu_ctx_t *vfu_ctx, char *buf,
  *    and writes are an error
  *  - otherwise, the callback is expected to handle the access
  *
+ * Regions VFU_PCI_DEV_MIGR_REGION_IDX and VFU_GENERIC_DEV_MIGR_REG_IDX,
+ * corresponding to the migration region, enable live migration support for
+ * the device. The migration region must contain at the beginning the migration
+ * registers (struct vfio_device_migration_info defined in <linux/vfio.h>) and
+ * the remaining part of the region can be arbitrarily used by the device
+ * implementation. Therefore, the size of the migration region must be at least
+ * sizeof(struct vfio_device_migration_info). libvfio-user offers two ways for
```
tmakatos, line 254: Add function that returns page aligned space required for migration registers.

include/libvfio-user.h
```diff
@@ -422,21 +446,26 @@ typedef struct {
...
-
-//TODO: Re-visit once migration support is done.
 /**
- * Enable support for device migration.
+ * vfu_setup_device_migration provides an abstraction over the migration
+ * protocol: the user specifies a set of callbacks which are called in respone
+ * to client accesses of the migration region; the migration read/write
+ * callbacks are not called. Offsets in callbacks are relative to @data_offset.
+ * 
+ * FIXME we shouldn't be using vfu_migration_callbacks_t for the same reason
```
tmakatos, line 455: Add single function that sets migration callback based on enum.

tmakatos (commented):
Added comments based on meeting.

tmakatos:
I think configuring migration has started to become a bit cumbersome: the user needs to create the migration region (`vfu_setup_region`) making sure the size is correct (by calling the newly-added `vfu_get_migr_regs_size`) and then configure it via `vfu_setup_device_migration_callbacks`. I think it still makes sense to add another function that combines all these (allocating the migration region memory if it's to be sparse mmaped) for the benefit of simpler implementations.

jlevon:
I don't agree that we want a new function that literally *is* adding a region, plus other stuff. It's *more* confusing to me to have it look like it's something new/different when it isn't.

tmakatos:
> I don't agree that we want a new function that literally _is_ adding a region, plus other stuff. It's _more_ confusing to me to have it look like it's something new/different when it isn't.

Fair enough.

tmakatos:
All the commits will be squashed since they're very relevant, I've put them separately to make reviewing easier.

include/libvfio-user.h
```diff
@@ -240,6 +245,26 @@ typedef ssize_t (vfu_region_access_cb_t)(vfu_ctx_t *vfu_ctx, char *buf,
  *    and writes are an error
  *  - otherwise, the callback is expected to handle the access
  *
+ * Regions VFU_PCI_DEV_MIGR_REGION_IDX and VFU_GENERIC_DEV_MIGR_REG_IDX,
+ * corresponding to the migration region, enable live migration support for
+ * the device. The migration region must contain at the beginning the migration
+ * registers (struct vfio_device_migration_info defined in <linux/vfio.h>) and
+ * the remaining part of the region can be arbitrarily used by the device
+ * implementation. Therefore, the size of the migration region must be at least
+ * sizeof(struct vfio_device_migration_info). libvfio-user offers two ways for
```
jlevon, line 254: Can we clarify this, specifically say something like:

The region provided must have at least vfu_get_migr_regs_size() bytes available at the start of the region (this size is guaranteed to be page-aligned). If mmap_areas is given, it must *not* include this part of the region.

jlevon, line 254: New para here? it's a bit hard to follow all in one

include/libvfio-user.h
```diff
@@ -240,6 +245,26 @@ typedef ssize_t (vfu_region_access_cb_t)(vfu_ctx_t *vfu_ctx, char *buf,
...
+ * the migration region to be used:
+ *  1. natively: the device implementation must handle accesses to the
+ *      migration registers and migration data via the region callbacks. The
+ *      semantics of these registers are explained in <linux/vfio.h>.
+ *  2. via the vfu_migration_t callbacks: the device implementation registers
+ *      a set of callbacks by calling vfu_setup_device_migration. The device
```
jlevon, line 260: I think you can remove this "The device ... see function vfu_get_migr_regs_size)". if you have the above. Right now it sounds like if you choose 1) then you can do what you want with mmap, but that's not really true is it?

include/libvfio-user.h
```diff
@@ -240,6 +245,26 @@ typedef ssize_t (vfu_region_access_cb_t)(vfu_ctx_t *vfu_ctx, char *buf,
...
+ *      does not have access to the migration registers, however the migration
+ *      region provided must still reserve space for them (see function
+ *      vfu_get_migr_regs_size). The region's read/write callbacks are never
+ *      called.
+ * The migration region can be partially memory mapped, except the migration
```
jlevon, line 265: Think you can remove this too if you do the above

include/libvfio-user.h
```diff
@@ -261,6 +286,13 @@ vfu_setup_region(vfu_ctx_t *vfu_ctx, int region_idx, size_t size,
                  struct iovec *mmap_areas, uint32_t nr_mmap_areas,
                  int fd);
 
+/*
+ * Returns the size of the migration registers, which is guaranteed to be page
```
jlevon, line 290: This makes it sound like the registers themselves take up a page. Maybe generalize a bit?

Returns the size of the area needed to hold the migration registers at the beginning of the migration region; guaranteed to be page aligned.

and call it "vfu_get_migr_register_area_size" or somesuch? What do you think?

include/libvfio-user.h
```diff
@@ -422,21 +461,22 @@ typedef struct {
...
-    vfu_migration_callbacks_t   callbacks;
-    struct iovec                *mmap_areas;
-    uint32_t                    nr_mmap_areas;
-} vfu_migration_t;
-
-//TODO: Re-visit once migration support is done.
 /**
- * Enable support for device migration.
+ * vfu_setup_device_migration provides an abstraction over the migration
+ * protocol: the user specifies a set of callbacks which are called in respone
```
jlevon, line 466: nit, response

include/libvfio-user.h
```diff
@@ -422,21 +461,22 @@ typedef struct {
...
+ * to client accesses of the migration region; the migration read/write
```
jlevon, line 467: I think you mean the "region read/write callbacks are not called"; maybe add "after this function call".

include/libvfio-user.h
```diff
@@ -422,21 +461,22 @@ typedef struct {
...
- * @migration: information required to migrate device
+ * @callbacks: migration callbacks
+ * @data_offset: offset in the migration region where data begins.
+ *
+ * @returns 0 on success, -1 on error, Sets errno.
  */
 int
-vfu_setup_device_migration(vfu_ctx_t *vfu_ctx, vfu_migration_t *migration);
+vfu_setup_device_migration_callbacks(vfu_ctx_t *vfu_ctx,
+                                     const vfu_migration_callbacks_t * callbacks,
```
jlevon, line 478: nit, spurious space after *

include/libvfio-user.h
```diff
@@ -422,21 +461,22 @@ typedef struct {
...
+ * protocol: the user specifies a set of callbacks which are called in respone
+ * to client accesses of the migration region; the migration read/write
+ * callbacks are not called. Offsets in callbacks are relative to @data_offset.
+ * 
  * @vfu_ctx: the libvfio-user context
- * @migration: information required to migrate device
+ * @callbacks: migration callbacks
+ * @data_offset: offset in the migration region where data begins.
+ *
+ * @returns 0 on success, -1 on error, Sets errno.
```
jlevon, line 474: nit, s/Sets/sets/

include/libvfio-user.h
```diff
@@ -550,6 +590,18 @@ vfu_dma_write(vfu_ctx_t *vfu_ctx, dma_sg_t *sg, void *data);
 
 /*
  * Supported PCI regions.
+ *
+ * Note to developers: in VFIO, each region starts at a terabyte offset
```
jlevon, line 594: I don't think you need "Note to developers": everyone's a developer :)

include/libvfio-user.h
```diff
@@ -550,6 +590,18 @@ vfu_dma_write(vfu_ctx_t *vfu_ctx, dma_sg_t *sg, void *data);
...
+ * (VFIO_PCI_INDEX_TO_OFFSET) and because Linux supports up to 128 TB of
+ * virtual memory, there can be up to 128 device regions. PCI regions are fixed
```
jlevon, line 596: *userspace* virtual memory ?

include/libvfio-user.h
```diff
@@ -550,6 +590,18 @@ vfu_dma_write(vfu_ctx_t *vfu_ctx, dma_sg_t *sg, void *data);
...
+ * and in retrospect this choice has proven to be problematic because devices
+ * might contain potentially unused regions. New regions can now be positioned
+ * anywhere by using the VFIO_REGION_INFO_CAP_TYPE capability.
+ * In vfio-user we don't have this problem because the region index is just an
+ * identifier: the VMM memory maps a file descriptor that is passed to it and
+ * the offset that it memory maps is relative to the beginning of the region,
```
jlevon, line 602: I think the crucial point to make here is that the mapping offset is derived from mmap_areas offset value, rather than a static mapping from region index to offset.

lib/libvfio-user.c
```diff
@@ -1234,6 +1239,38 @@ copyin_mmap_areas(vfu_reg_info_t *reg_info,
...
+     * utility functions?
+     */
+    return (off1 < (off2 + size2) && off2 < (off1 + size1));
+}
+
+static bool
+maps_over_migr_regs(struct iovec *iov)
+{
+    return ranges_intersect(0, sizeof(struct vfio_device_migration_info),
+                            (size_t)iov->iov_base, (size_t)iov->iov_len);
```
jlevon, line 1258: nit, no need to cast iov_len

lib/libvfio-user.c
```diff
@@ -1266,6 +1303,12 @@ vfu_setup_region(vfu_ctx_t *vfu_ctx, int region_idx, size_t size,
         return ERROR(EINVAL);
     }
 
+    if (region_idx == VFU_PCI_DEV_MIGR_REGION_IDX &&
+        size < sizeof(struct vfio_device_migration_info)) {
+        vfu_log(vfu_ctx, LOG_ERR, "ivalid migration region size %d", size);
```
jlevon, line 1308: nit, "invalid"

lib/libvfio-user.c
```diff
@@ -1266,6 +1303,12 @@ vfu_setup_region(vfu_ctx_t *vfu_ctx, int region_idx, size_t size,
         return ERROR(EINVAL);
     }
 
+    if (region_idx == VFU_PCI_DEV_MIGR_REGION_IDX &&
+        size < sizeof(struct vfio_device_migration_info)) {
```
jlevon, line 1307: This should be requiring the page aligned size (i.e. what your utility function returns), not the struct size. Otherwise it's just more confusing I think.

lib/libvfio-user.c
```diff
@@ -1344,41 +1406,39 @@ vfu_setup_device_nr_irqs(vfu_ctx_t *vfu_ctx, enum vfu_dev_irq_type type,
...
+vfu_get_migr_regs_size(void)
+{
+    return ROUND_UP(sizeof(struct vfio_device_migration_info),
+                    sysconf(_SC_PAGE_SIZE));
+}
+
 int
-vfu_setup_device_migration(vfu_ctx_t *vfu_ctx, vfu_migration_t *migration)
+vfu_setup_device_migration_callbacks(vfu_ctx_t *vfu_ctx,
+                                     const vfu_migration_callbacks_t * callbacks,
```
jlevon, line 1418: nit, spurious space after *

lib/migration.c
```diff
@@ -51,9 +51,14 @@ enum migr_iter_state {
 };
 
 struct migration {
+    /*
+     * TODO if the user provides an FD then should mmap it an use the migration
```
jlevon, line 55: nit, "and use"

lib/migration.h
```diff
@@ -44,7 +44,8 @@
 #include "libvfio-user.h"
 
 struct migration *
-init_migration(const vfu_migration_t * const vfu_migr, int *err);
+init_migration(const vfu_migration_callbacks_t * callbacks,
```
jlevon, line 47: nit, spurious space after *

samples/client.c
```diff
@@ -217,12 +217,13 @@ get_region_vfio_caps(struct vfio_info_cap_header *header,
     while (true) {
         switch (header->id) {
             case VFIO_REGION_INFO_CAP_SPARSE_MMAP:
-                sparse = (struct vfio_region_info_cap_sparse_mmap*)header;
+                *sparse = (struct vfio_region_info_cap_sparse_mmap*)header;
```
jlevon, line 220: missing space before *

samples/server.c
```diff
@@ -501,21 +523,41 @@ int main(int argc, char *argv[])
...
+        err(EXIT_FAILURE, "failed to truncate migration file");
+    }
+    server_data.bar1 = mmap(NULL, server_data.bar1_size, PROT_READ | PROT_WRITE,
+                            MAP_SHARED, fileno(bar1_fp), 0);
+    if (server_data.bar1 == MAP_FAILED) {
+        err(EXIT_FAILURE, "failed to mmap migration file");
+    }
+    struct iovec migr_mmap_areas[] = {
+        [0] = {
+            .iov_base  = (void*)migr_regs_size,
```
jlevon, line 549: nit, missing space before *

test/unit-tests.c
```diff
@@ -1202,11 +1203,157 @@ test_migration_state_transitions(void **state __attribute__ ((unused)))
...
+static struct test_setup_migr_reg_dat {
+    vfu_ctx_t *v;
+    size_t rs; /* migration registers size */
+    size_t ds; /* migration data size */
+    size_t s; /* migration region size*/
+    const vfu_migration_callbacks_t c;
+} migr_reg_data = {
+    .c = {
+        .version = VFU_MIGR_CALLBACKS_VERS,
+        .transition = (void*)0x1,
```
jlevon, line 1221: nit, here and below, void* -> void *

test/unit-tests.c
```diff
@@ -1202,11 +1203,157 @@ test_migration_state_transitions(void **state __attribute__ ((unused)))
...
+        .get_pending_bytes = (void*)0x2,
+        .prepare_data = (void*)0x3,
+        .read_data = (void*)0x4,
+        .write_data = (void*)0x5,
+        .data_written = (void*)0x6
+    }
+};
+
+static int
+setup_test_setup_migration_region(void **state) {
```
jlevon, line 1231: nit, { on newline

test/unit-tests.c
```diff
@@ -1202,11 +1203,157 @@ test_migration_state_transitions(void **state __attribute__ ((unused)))
...
+    p->ds = sysconf(_SC_PAGE_SIZE);
+    p->s = p->rs + p->ds;
+    *state = p;
+    return setup(state);
+}
+
+static vfu_ctx_t *
+get_vfu_ctx(void **state)
+{
+    return (*((struct test_setup_migr_reg_dat**)(state)))->v;
```
jlevon, line 1248: nit, s/**/ **/

test/unit-tests.c
```diff
@@ -1202,11 +1203,157 @@ test_migration_state_transitions(void **state __attribute__ ((unused)))
...
+}
+
+static int
+teardown_test_setup_migration_retion(void **state)
```
jlevon, line 1252: nit, "region" ?

test/unit-tests.c
```diff
@@ -1202,11 +1203,157 @@ test_migration_state_transitions(void **state __attribute__ ((unused)))
...
+test_setup_migration_region_size_ok(void **state)
+{
+    vfu_ctx_t *v = get_vfu_ctx(state);
+    int r = vfu_setup_region(v, VFU_PCI_DEV_MIGR_REGION_IDX,
+        sizeof(struct vfio_device_migration_info), NULL,
+        VFU_REGION_FLAG_READ | VFU_REGION_FLAG_WRITE, NULL, 0, -1);
+    assert_int_equal(0, r);
+}
+
+static void
```
jlevon, line 1280: I'm wondering if we should handle this ourselves. Not sure

test/unit-tests.c
```diff
@@ -1202,11 +1203,157 @@ test_migration_state_transitions(void **state __attribute__ ((unused)))
...
+    assert_int_equal(EINVAL, errno);
+}
+
+static void
+test_setup_migration_region_sparsely_mappable_valid(void **state)
+{
+    struct test_setup_migr_reg_dat *p = *state;
+    struct iovec mmap_areas[] = {
+        [0] = {
+            .iov_base = (void*)p->rs,
```
jlevon, line 1313: nit, void *

jlevon (changes requested):

include/libvfio-user.h
tmakatos: Sure, I was hoping for a better wording :).

tmakatos: ok

include/libvfio-user.h
tmakatos: ok

include/libvfio-user.h
tmakatos: ok

include/libvfio-user.h
tmakatos: ok

test/unit-tests.c
tmakatos: ok

test/unit-tests.c
tmakatos: ok

test/unit-tests.c
tmakatos: ok

tmakatos: 'region' isn't that accurate, since the pointer points to `struct test_setup_migr_reg_dat`, which is 'migration region test data'. Honestly I think it'f fine as 'state' since the cast is right below.

test/unit-tests.c
tmakatos: Do you mean making the migration region mappable without telling the user?

test/unit-tests.c
tmakatos: ok

test/unit-tests.c
jlevon: It's the "retion" part I was talking about

test/unit-tests.c
jlevon: So for non-migration regions, if an fd is provided, but no mmap areas, we map the whole region.
I'm saying that for the migration region, if an fd is provided , but no mmap areas, could we map the whole region, excluding the first page?

test/unit-tests.c
tmakatos: ah yes

test/unit-tests.c
tmakatos: We can't assume that the space after the migration registers is simply padding, the user might setup the migration region without calling `vfu_get_migr_register_area_size` and putting data right after the migration registers.

tmakatos:
@jlevon fixed typo and replied to your comment regarding implicitly making the migration region mappable, please take another look.

test/unit-tests.c
jlevon: I'm not sure why we need to support that (but maybe we do?). Isn't it simpler to have a simple "page size" requirement?

tmakatos: We don't necessarily need to support that, but seems that we're making it more complicated (a) by introducing a requirement, and (b) by changing the semantics of the how the function works if mmap areas is NULL but only if this is the migration region. I think it will be harder to get it right this way.

jlevon: ok, but I think we should always insist on a page-sized buffer, regardless.

tmakatos: > ok, but I think we should always insist on a page-sized buffer, regardless.

We sure can, but if we're introducing a restriction we should explain why?

I think if we want to make the API easier we can explain that the simplest way to make the migration region mappable is for the client to use `vfu_get_migr_register_area_size` in order to allocate space for the migration registers and then

jlevon: Right, I think we're agreeing: we should say that regardless of whether you're using the callbacks, and regardless of if you're providing an fd, you need to reserve a non-mappable area of size vfu_get_migr_register_area_size().

tmakatos: I re-read the PR and it's currently implemented the way you suggest:
```
The migration region must contain at the beginning the migration
registers (struct vfio_device_migration_info defined in <linux/vfio.h>) and
the remaining part of the region can be arbitrarily used by the device
implementation. The region provided must have at least
vfu_get_migr_register_area_size() bytes available at the start of the region
(this size is guaranteed to be page-aligned).
```
So going back to your original comment in this discussion, we're deciding _not_ to special case the `mmap_areas == NULL` case for the migration region, right?

jlevon: We won't special case.

But it's not implemented as you suggest, yet:

```
   return ranges_intersect(0, sizeof(struct vfio_device_migration_info),
                            (size_t)iov->iov_base, (size_t)iov->iov_len);
```

```
    if (region_idx == VFU_PCI_DEV_MIGR_REGION_IDX &&
        size < sizeof(struct vfio_device_migration_info)) {
```

At least both of these are using the struct size not the vfu_get_migr_register_area_size()

tmakatos: Right, this needs fixing.

tmakatos:
@jlevon fixed, please take another look.

jlevon:
I'm not seeing changes for my review comments

tmakatos:
> I'm not seeing changes for my review comments

I somehow missed some of them, let me take a look.

include/libvfio-user.h
tmakatos: Thanks, I know the way I wrote it was a bit convoluted and was hoping in a suggestion :).

tmakatos:
@jlevon addressed your comments.

lib/irq.c
```diff
@@ -427,10 +428,10 @@ vfu_irq_message(vfu_ctx_t *vfu_ctx, uint32_t subindex)
     }
 
     irq_info.subindex = subindex;
-    ret = tran_sock_msg(vfu_ctx->conn_fd, msg_id,
-                        VFIO_USER_VM_INTERRUPT,
-                        &irq_info, sizeof irq_info,
-                        NULL, NULL, 0);
+    ret = vfu_ctx->trans->send_msg(vfu_ctx, msg_id,
```
jlevon, line 431: I compared against master, and I'm so confused as to why github thinks there's a change here.

jlevon (approved):
All looks good, thanks. I'm a bit worried about why my changes are showing up in the diffs though, wtf?

lib/irq.c
tmakatos: Not sure either, I also diff'ed against master and don't see your changes. I'll merge and if it breaks things I'll revert the merge.

tmakatos:
@jlevon your changes don't appear in the commit on master, we're good.

PR #316, opened 2021-02-09
remove recv_blocking()

jlevon:
It has the incorrect implication that other socket I/O is necessarily
non-blocking. Replace with an explicit recv(..., MSG_WAITALL).


jlevon:
If this seems familiar, it's because I accidentally merged it into another PR, then somehow dropped it.
Swapnil: you approved this last time I think.

swapnili (approved):

PR #320, opened 2021-02-09
move exec_command socket handling into the transport

jlevon:
Also clean up some code surrounding this. In particular, don't play games with
modifying the message header.

### DIFF
--- a/lib/tran_sock.c
+++ b/lib/tran_sock.c
@@ -524,7 +524,7 @@ recv_version(vfu_ctx_t *vfu_ctx, int sock, uint16_t *msg_idp,
     }
 
     if (hdr.cmd != VFIO_USER_VERSION) {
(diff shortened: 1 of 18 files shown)


lib/tran_sock.c
```diff
@@ -712,6 +712,39 @@ tran_sock_get_request(vfu_ctx_t *vfu_ctx, struct vfio_user_header *hdr,
...
 
+static int
+tran_sock_recv_body(vfu_ctx_t *vfu_ctx, const struct vfio_user_header *hdr,
+                    void **datap)
+{
+    size_t body_size = hdr->msg_size - sizeof (*hdr);
+    void *data;
+    int ret;
+
+    // FIXME: should check max-msg-size
```
tmakatos, line 723: an assert for now would be enough

tmakatos (approved):

lib/tran_sock.c
jlevon: As you noticed, I fix this up later

PR #325, opened 2021-02-10
API error return converged to one func

swapnili:
Use ERROR_INT() or ERROR_PTR() to return errors from API's.
This way if we want to change the behaviour later, we will just need
to update these funcitons.
Also fixed some error return cases and comments.

### DIFF
--- a/lib/irq.c
+++ b/lib/irq.c
@@ -376,31 +376,29 @@ handle_device_set_irqs(vfu_ctx_t *vfu_ctx, uint32_t size,
     return dev_set_irqs(vfu_ctx, irq_set, data);
 }
 
-static int
+static bool
 validate_irq_subindex(vfu_ctx_t *vfu_ctx, uint32_t subindex)
 {
     if (vfu_ctx == NULL) {
-        return -1;
+        return false;
     }
 
     if ((subindex >= vfu_ctx->irqs->max_ivs)) {
         vfu_log(vfu_ctx, LOG_ERR, "bad IRQ %d, max=%d\n", subindex,
                vfu_ctx->irqs->max_ivs);
-        return -1;
+        return false;
     }
 
-    return 0;
+    return true;
 }
 
 int
 vfu_irq_trigger(vfu_ctx_t *vfu_ctx, uint32_t subindex)
 {
-    int ret;
     eventfd_t val = 1;
 
-    ret = validate_irq_subindex(vfu_ctx, subindex);
-    if (ret < 0) {
-        return ERROR_INT(-ret);
+    if (validate_irq_subindex(vfu_ctx, subindex) == false) {
+        return ERROR_INT(EINVAL);
     }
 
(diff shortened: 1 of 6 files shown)


lib/irq.c
```diff
@@ -403,14 +400,12 @@ vfu_irq_trigger(vfu_ctx_t *vfu_ctx, uint32_t subindex)
 
     ret = validate_irq_subindex(vfu_ctx, subindex);
     if (ret < 0) {
-        return ret;
+        return ERROR_INT(-ret);
```
jlevon, line 403: should be ERROR_INT(EINVAL)

lib/irq.c
```diff
@@ -380,15 +380,12 @@ static int
 validate_irq_subindex(vfu_ctx_t *vfu_ctx, uint32_t subindex)
```
jlevon, line 380: now this doesn't fiddle with errno, can you make it a boolean return like other validate functions?

jlevon (changes requested):
Thanks for this, nice cleanup.

lib/irq.c
swapnili: Thanks. Fixed it.

lib/irq.c
swapnili: Sure

lib/irq.c
```diff
@@ -376,41 +376,34 @@ handle_device_set_irqs(vfu_ctx_t *vfu_ctx, uint32_t size,
...
 int
 vfu_irq_trigger(vfu_ctx_t *vfu_ctx, uint32_t subindex)
 {
-    int ret;
     eventfd_t val = 1;
 
-    ret = validate_irq_subindex(vfu_ctx, subindex);
-    if (ret < 0) {
-        return ret;
+    if (validate_irq_subindex(vfu_ctx, subindex) == false) {
```
jlevon, line 400: nit, sorry, but don't ever do "== <boolean>", this should just be:

  if (!validate_irq_subindex()) {

lib/irq.c
```diff
@@ -423,8 +416,8 @@ vfu_irq_message(vfu_ctx_t *vfu_ctx, uint32_t subindex)
     struct vfio_user_irq_info irq_info;
 
     ret = validate_irq_subindex(vfu_ctx, subindex);
-    if (ret < 0) {
-        return -1;
+    if (ret == false) {
```
jlevon, line 419: ditto:

if (!validate_irq_subindex()) {

Here, we also shouldn't set ret, as that's an int

jlevon (changes requested):
w

lib/irq.c
swapnili: Yes I also do not like doing "==", but its the frodo coding_style which says so at-least for checking NULL.
Changed it to !

lib/irq.c
swapnili: Done

jlevon (approved):

PR #326, opened 2021-02-10
fix pre-push build order

jlevon:
Re-order so we don't end up with an ASAN-enabled build at the end; purely for
convenience.


swapnili:
Can you please also write about the issue with old build order?

jlevon:
hmm, not sure why the body of the commit didn't end up there, fixed.

swapnili (approved):

PR #333, opened 2021-02-11
gpio: print errno when vfu_run_ctx fails

tmakatos:


jlevon (changes requested):
err() already does this

tmakatos:
> err() already does this

Right, not sure how I missed this.

PR #337, opened 2021-02-11
fix VFIO_USER_DEVICE_GET_INFO request size check

jlevon:
The spec says "only the message header is needed in the command
message", we were needlessly requiring the device info struct to be provided
too.

### DIFF
--- a/lib/libvfio-user.c
+++ b/lib/libvfio-user.c
@@ -433,7 +433,7 @@ handle_device_get_info(vfu_ctx_t *vfu_ctx, uint32_t size,
     assert(vfu_ctx != NULL);
     assert(dev_info != NULL);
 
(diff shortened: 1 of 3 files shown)


tmakatos:
I think the plan was to include `struct vfio_device_info` in the request to mimic VFIO. Maybe there's value in keeping it this way so that the client can set `argsz` and the server can tell which exact version of the struct to respond with? Or will we be covered by the verson negotiation?

jlevon:
It's how (say) get region info works, certainly. It's a bit weird since vfio of course has to provide a buffer - it's ioctl not socket - but providing an unused struct, (except argsz I guess), is very weird over a socket

jlevon:
After input by JJ, going to fix the spec instead.

PR #339, opened 2021-02-11
rename fedora workflow

jlevon:
We're no longer running CI on fedora-rawhide, so rename it.


tmakatos (approved):
Looks good. Do we need to specify the version? Won't the latest stable vesion be used if we just specify "fedora"?

jlevon:
Well, right now, we're installing fedora 33 and using dnf to update to a hard-coded "34".

If we wanted to fix this more I think we'd figure out how to fix rawhide again.

Issue #342, opened 2021-02-12
 make CFLAGS="-Wmissing-declarations" fails


tmakatos:
```
 make CFLAGS="-Wmissing-declarations"
...
[ 15%] Building C object lib/CMakeFiles/dma.dir/dma.c.o
/root/src/libvfio-user/lib/dma.c:94:1: error: no previous declaration for ‘_dma_controller_do_remove_region’ [-Werror=missing-declarations]
 _dma_controller_do_remove_region(dma_controller_t *dma,
 ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/root/src/libvfio-user/lib/dma.c:420:9: error: no previous declaration for ‘_get_bitmap_size’ [-Werror=missing-declarations]
 ssize_t _get_bitmap_size(size_t region_size, size_t pgsize)
```

jlevon:
oh, definitely worth adding!

jlevon:
Looks like this wasn't actually fixed

jlevon:
fixed by #399

PR #346, opened 2021-02-16
unit test exec_command and friends w.r.t. migration device state

tmakatos:


tmakatos:
TODO fill in how to repro bug with `make test CC=gcc BUILD_TYPE=rel`

swapnili:
> Using `make test CC=gcc BUILD_TYPE=rel`
> commit [c592301](https://github.com/nutanix/libvfio-user/commit/c5923013fc309a985d8b19c1f95a63ccf7134639) works, while commit [0e034d6](https://github.com/nutanix/libvfio-user/commit/0e034d6385cb1eac9eef7750dece96415936203f) doesn't.
> @jlevon @swapnili any ideas why this breaks?!

Looks like you fixed the compilation error and not the unit-test is failing https://github.com/nutanix/libvfio-user/pull/346/checks?check_run_id=1912354953

tmakatos:
> > Using `make test CC=gcc BUILD_TYPE=rel`
> > commit [c592301](https://github.com/nutanix/libvfio-user/commit/c5923013fc309a985d8b19c1f95a63ccf7134639) works, while commit [0e034d6](https://github.com/nutanix/libvfio-user/commit/0e034d6385cb1eac9eef7750dece96415936203f) doesn't.
> > @jlevon @swapnili any ideas why this breaks?!
> 
> Looks like you fixed the compilation error and not the unit-test is failing https://github.com/nutanix/libvfio-user/pull/346/checks?check_run_id=1912354953

I wasn't clear, let me clarify: if compiled with `make CC=gcc BUILD_TYPE=rel`, commit 0e034d6385cb1eac9eef7750dece96415936203f fails the `test_cmd_allowed_when_stopped_and_copying` unit test:
```
[ RUN      ] test_cmd_allowed_when_stopped_and_copying
[  ERROR   ] --- r
[   LINE   ] --- /root/src/libvfio-user/test/unit-tests.c:1436: error: Failure!
```
This doesn't happen if it's compiled with `BUILD_TYPE=dbg`. The previous commit, c5923013fc309a985d8b19c1f95a63ccf7134639, works correctly. Looking at the "bad" commit (0e034d6385cb1eac9eef7750dece96415936203f), I don't understand why this change would break the unit test, I've tried to debug this and in unit test `test_cmd_allowed_when_stopped_and_copying` and for `i=0` `cmd_allowed_when_stopped_and_copying` returns `1` while the expression inside it is correctly evaluated to `0`. This doesn't make any sense.

tmakatos:
I managed to create a minimal repro and it turns out the problem is not providing a function declaration for the `__real_` functions, since they get resolved anyway. My suspicion is that we automatically get a function declaretion that default to return type `int` (4 bytes) whereas we return `bool` (1 byte). Will confirm that this is the case for libvfio-user and will also provide the minimal repro.

jlevon:
cf https://github.com/nutanix/libvfio-user/issues/342 right?

tmakatos:
> cf #342 right?

"cf" ?

tmakatos:
> Will confirm that this is the case for libvfio-user and will also provide the minimal repro.

Indeed I get the same behavior in the unit tests, I've sent an email to the GCC mailing list to clarify, I think this results in legitimate undefined behavior, so I guess we should fix #342.

tmakatos:
#349 should fix the bug.

swapnili (approved):

PR #355, opened 2021-02-22
don't use wrong buf offset for responding with PCI caps

tmakatos:

### DIFF
--- a/lib/pci.c
+++ b/lib/pci.c
@@ -393,6 +393,7 @@ pci_config_space_access(vfu_ctx_t *vfu_ctx, char *buf, size_t count,
 
         offset += ret;
         count -= ret;
+        buf += ret;
     }
 
     return offset - start;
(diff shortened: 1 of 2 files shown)


jlevon (approved):
d'oh, thanks

tmakatos:
> d'oh, thanks

This unit test fails: https://github.com/nutanix/libvfio-user/pull/355/checks?check_run_id=1953750278

Need to figure out why.

jlevon:
Actually, I think this isn't right after all.

The bug is probably in pci_config_space_access() - in the loop, we are not moving "buf" along too, ie.. need:

393    buf += ret;

right?

jlevon (changes requested):
as per comment

tmakatos:
> Actually, I think this isn't right after all.
> 
> The bug is probably in pci_config_space_access() - in the loop, we are not moving "buf" along too, ie.. need:
> 
> 393 buf += ret;
> 
> right?

Maybe, this does allow the controller to progress further but I'm still getting errors, will continue debugging to see whether they're related.

jlevon (approved):
Thank you

Issue #347, opened 2021-02-17
unit tests fail with valgrind


tmakatos:
Testing commit e4e0ebde2c25f3011d20b6a2f1b1b8f510b09d9e:
```
# valgrind build/dbg/test/unit-tests
...
[ RUN      ] test_pci_ext_caps
==30751== Conditional jump or move depends on uninitialised value(s)
==30751==    at 0x11C29B: pci_config_space_ptr (pci.h:57)
==30751==    by 0x11DED5: vfu_pci_find_next_ext_capability (pci_caps.c:660)
==30751==    by 0x11DFFD: vfu_pci_find_next_capability (pci_caps.c:694)
==30751==    by 0x110B1E: test_pci_ext_caps (unit-tests.c:958)
==30751==    by 0x485E0D8: ??? (in /usr/lib/x86_64-linux-gnu/libcmocka.so.0.5.1)
==30751==    by 0x485EA48: _cmocka_run_group_tests (in /usr/lib/x86_64-linux-gnu/libcmocka.so.0.5.1)
==30751==    by 0x112484: main (unit-tests.c:1445)
==30751==
==30751== Conditional jump or move depends on uninitialised value(s)
==30751==    at 0x11DF10: vfu_pci_find_next_ext_capability (pci_caps.c:666)
==30751==    by 0x11DFFD: vfu_pci_find_next_capability (pci_caps.c:694)
==30751==    by 0x110B1E: test_pci_ext_caps (unit-tests.c:958)
==30751==    by 0x485E0D8: ??? (in /usr/lib/x86_64-linux-gnu/libcmocka.so.0.5.1)
==30751==    by 0x485EA48: _cmocka_run_group_tests (in /usr/lib/x86_64-linux-gnu/libcmocka.so.0.5.1)
==30751==    by 0x112484: main (unit-tests.c:1445)
==30751==
==30751== Use of uninitialised value of size 8
==30751==    at 0x11DF28: vfu_pci_find_next_ext_capability (pci_caps.c:671)
==30751==    by 0x11DFFD: vfu_pci_find_next_capability (pci_caps.c:694)
==30751==    by 0x110B1E: test_pci_ext_caps (unit-tests.c:958)
==30751==    by 0x485E0D8: ??? (in /usr/lib/x86_64-linux-gnu/libcmocka.so.0.5.1)
[...]

PR #357, opened 2021-02-22
fix false positive under valgrind

jlevon:
valgrind gets confused by a non-aligned pointer load; use an aligned load
so it's quietened.


jlevon:
Fixes #347

jlevon:
This was not a false positive - see #363

PR #364, opened 2021-02-23
ignore VFIO_IRQ_SET_DATA_NONE

tmakatos:

### DIFF
--- a/lib/irq.c
+++ b/lib/irq.c
@@ -355,6 +355,9 @@ handle_device_set_irqs(vfu_ctx_t *vfu_ctx, uint32_t size,
     }
 
     switch (irq_set->flags & VFIO_IRQ_SET_DATA_TYPE_MASK) {
+        case VFIO_IRQ_SET_DATA_NONE:
+            vfu_log(vfu_ctx, LOG_NOTICE, "ignoring IRQ type %d");
+            return 0;
         case VFIO_IRQ_SET_DATA_EVENTFD:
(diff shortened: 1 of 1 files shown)


lib/irq.c
```diff
@@ -355,6 +355,9 @@ handle_device_set_irqs(vfu_ctx_t *vfu_ctx, uint32_t size,
     }
 
     switch (irq_set->flags & VFIO_IRQ_SET_DATA_TYPE_MASK) {
+        case VFIO_IRQ_SET_DATA_NONE:
```
jlevon, line 358: we already have code below this (in dev_set_irqs() and irqs_trigger() -> irqs_set_data_none()). Don't we want to just set data to NULL then pass it through?

I know we don't do MASK/UNMASK (though I'm not sure I really understand that),
but what about TRIGGER?

jlevon (changes requested):

lib/irq.c
tmakatos: > we already have code below this (in dev_set_irqs() and irqs_trigger() -> irqs_set_data_none()). Don't we want to just set data to NULL then pass it through?
> 
> I know we don't do MASK/UNMASK (though I'm not sure I really understand that),
> but what about TRIGGER?

Probably, I don't understand much about this, I just looked at older code and that's how we used to handle it.

jlevon (approved):
thanks!

PR #375, opened 2021-02-26
optionally enable DMA controller in GPIO sample

tmakatos:
This helps debugging.


jlevon (approved):
sure, but any reason not to always do so?

jlevon:
> sure, but any reason not to always do so?

never mind, I see (there's still bugs)

PR #376, opened 2021-02-26
refactor message handling path

jlevon:


lib/libvfio-user.c
```diff
@@ -286,55 +287,94 @@ is_valid_region_access(vfu_ctx_t *vfu_ctx, size_t size, uint16_t cmd,
...
     }
 
-    ret = region_access(vfu_ctx, ra->region, buf, ra->count, ra->offset,
-                        cmd == VFIO_USER_REGION_WRITE);
+    ret = region_access(vfu_ctx, in_ra->region, buf, in_ra->count,
+                        in_ra->offset, msg->hdr.cmd == VFIO_USER_REGION_WRITE);
 
-    if (ret != ra->count) {
+    if (ret != in_ra->count) {
         vfu_log(vfu_ctx, LOG_ERR, "failed to %s %#x-%#lx: %s",
```
jlevon, line 332: There was a bug here: we'd leak *data (found via python tests). Shouldn't occur now we use ->out_data.

tmakatos (approved):
Since this is quite a big patch to review, can you provide some background on which problem it solves and the general idea how we solve it?

jlevon:
> some background

The commit message should - I hope - be enough context for the reason for the change. Should I expand that somehow? Is it not clear enough?

tmakatos:
> > some background
> 
> The commit message should - I hope - be enough context for the reason for the change. Should I expand that somehow? Is it not clear enough?

Ah, I only read the title of the PR, I didn't actually check the commit message in the commit. I think we're good.

Issue #378, opened 2021-03-01
libvfio-user fails to build with -j2


tmakatos:
```/usr/bin/ld: error: cannot find -lvfio-user
[ 73%] Built target migration_ut
CMakeFiles/gpio-pci-idio-16.dir/gpio-pci-idio-16.c.o:gpio-pci-idio-16.c:function migration_device_state_transition: error: undefined reference to 'vfu_log'
CMakeFiles/gpio-pci-idio-16.dir/gpio-pci-idio-16.c.o:gpio-pci-idio-16.c:function main: error: undefined reference to 'vfu_get_migr_register_area_size'
CMakeFiles/gpio-pci-idio-16.dir/gpio-pci-idio-16.c.o:gpio-pci-idio-16.c:function main: error: undefined reference to 'vfu_create_ctx'
CMakeFiles/gpio-pci-idio-16.dir/gpio-pci-idio-16.c.o:gpio-pci-idio-16.c:function main: error: undefined reference to 'vfu_setup_log'
CMakeFiles/gpio-pci-idio-16.dir/gpio-pci-idio-16.c.o:gpio-pci-idio-16.c:function main: error: undefined reference to 'vfu_pci_init'
CMakeFiles/gpio-pci-idio-16.dir/gpio-pci-idio-16.c.o:gpio-pci-idio-16.c:function main: error: undefined reference to 'vfu_pci_set_id'
CMakeFiles/gpio-pci-idio-16.dir/gpio-pci-idio-16.c.o:gpio-pci-idio-16.c:function main: error: undefined reference to 'vfu_setup_region'
CMakeFiles/gpio-pci-idio-16.dir/gpio-pci-idio-16.c.o:gpio-pci-idio-16.c:function main: error: undefined reference to 'vfu_setup_region'
CMakeFiles/gpio-pci-idio-16.dir/gpio-pci-idio-16.c.o:gpio-pci-idio-16.c:function main: error: undefined reference to 'vfu_setup_device_migration_callbacks'
CMakeFiles/gpio-pci-idio-16.dir/gpio-pci-idio-16.c.o:gpio-pci-idio-16.c:function main: error: undefined reference to 'vfu_setup_device_nr_irqs'
CMakeFiles/gpio-pci-idio-16.dir/gpio-pci-idio-16.c.o:gpio-pci-idio-16.c:function main: error: undefined reference to 'vfu_realize_ctx'
[...]

PR #382, opened 2021-03-01
gpio-pci-idio-16: fix library dependency

jlevon:
Normally, target_link_libraries() is sufficient to add a dependency on the
linked library. But, as our CMake target is called "vfio-user-shared", we have
to make this explicit.

This was breaking "make -j" builds.

Reported-by: Karol Latecki <karol.latecki@intel.com>

### DIFF
--- a/samples/CMakeLists.txt
+++ b/samples/CMakeLists.txt
@@ -40,7 +40,7 @@ target_link_libraries(null vfio-user-static pthread)
 
 LINK_DIRECTORIES(${CMAKE_BINARY_DIR}/lib)
 add_executable(gpio-pci-idio-16 gpio-pci-idio-16.c)
-target_link_libraries(gpio-pci-idio-16 vfio-user)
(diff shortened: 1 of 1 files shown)


jlevon:
Fixes #378

tmakatos (approved):

samples/CMakeLists.txt
```diff
@@ -41,6 +41,8 @@ target_link_libraries(null vfio-user-static pthread)
 LINK_DIRECTORIES(${CMAKE_BINARY_DIR}/lib)
 add_executable(gpio-pci-idio-16 gpio-pci-idio-16.c)
 target_link_libraries(gpio-pci-idio-16 vfio-user)
```
swapnili, line 43: I think here it should be linking to vfio-user-shared?

jlevon: yeah, that's much better ,thanks.

swapnili (approved):

Issue #359, opened 2021-02-22
vfio-user lacks support for disabling IRQs


tmakatos:
Need to figure out how to handle this.

tmakatos:
```
[2021-02-22 12:09:22.039285] vfio_user.c:1324:vfio_user_log: *DEBUG*: /var/run/muser/domain/muser0/8: setting IRQ ERR flags=0x24
[2021-02-22 12:09:22.039299] vfio_user.c:1324:vfio_user_log: *DEBUG*: /var/run/muser/domain/muser0/8: event fd[0]=692
[2021-02-22 12:09:22.058335] vfio_user.c:1328:vfio_user_log: *ERROR*: /var/run/muser/domain/muser0/8: invalid IRQ type 1
[2021-02-22 12:09:22.058390] vfio_user.c:1328:vfio_user_log: *ERROR*: /var/run/muser/domain/muser0/8: msg0x16: cmd 7 failed: Invalid argument
```
and:
```
[2021-02-22 12:09:48.971144] vfio_user.c:1324:vfio_user_log: *DEBUG*: /var/run/muser/domain/muser0/8: enable MSI-X

[2021-02-22 12:09:48.971147] vfio_user.c:1324:vfio_user_log: *DEBUG*: /var/run/muser/domain/muser0/8: all MSI-X vectors masked

[2021-02-22 12:09:48.971167] vfio_user.c:1328:vfio_user_log: *ERROR*: /var/run/muser/domain/muser0/8: invalid IRQ type 1
[2021-02-22 12:09:48.971182] vfio_user.c:1328:vfio_user_log: *ERROR*: /var/run/muser/domain/muser0/8: msg0xf2: cmd 7 failed: Invalid argument
[2021-02-22 12:09:48.971256] vfio_user.c:1328:vfio_user_log: *ERROR*: /var/run/muser/domain/muser0/8: invalid IRQ type 1
[2021-02-22 12:09:48.971268] vfio_user.c:1328:vfio_user_log: *ERROR*: /var/run/muser/domain/muser0/8: msg0xf3: cmd 7 failed: Invalid argument
```

tmakatos:
We used to simply ignore this, not sure why.

jlevon:
unfortunately the failure message doesnt' tell us what the "action" was, nor what the IRQ index was. We should fix that.

We shouldn't be getting mask/unmask, as we don't report VFIO_IRQ_INFO_MASKABLE
So that leaves "trigger".

[...]

PR #386, opened 2021-03-08
fix IRQ disable path

jlevon:
Properly fix IRQ disabling:

Allow count == 0 to mean "disable all IRQS of the given type". On our side,
disabling an IRQ means forgetting about the eventfd that was previously passed
over the socket.

Allow individual IRQs to be disabled, by means of a VFIO_IRQ_SET_DATA_EVENTFD
message with no file descriptors passed. In vfio, this is done via setting "-1"
in the fd slots; which isn't possible via auxiliary data. Thus, only one IRQ can
be disabled a a time in vfio-user.

Clean up "->type": this is never set, so wasn't having any effect. Follow up
changes will likely re-introduce this in some form.

### DIFF
--- a/docs/vfio-user.rst
+++ b/docs/vfio-user.rst
@@ -1276,6 +1276,8 @@ VFIO IRQ set format
     was sent in the message meta-data. These descriptors will be signalled when
     the action defined by the action flags occurs. In AF_UNIX sockets, the
     descriptors are sent as SCM_RIGHTS type ancillary data.
+    If no file descriptors are provided, this de-assigns the specified
+    previously configured interrupts.
   * *VFIO_IRQ_SET_ACTION_MASK* indicates a masking event. It can be used with
     VFIO_IRQ_SET_DATA_BOOL or VFIO_IRQ_SET_DATA_NONE to mask an interrupt, or
     with VFIO_IRQ_SET_DATA_EVENTFD to generate an event when the guest masks
@@ -1292,8 +1294,8 @@ VFIO IRQ set format
 * *index* is the index of IRQ type being setup.
 * *start* is the start of the sub-index being set.
 * *count* describes the number of sub-indexes being set. As a special case, a
-  count of 0 with data flags of VFIO_IRQ_SET_DATA_NONE disables all interrupts
-  of the index.
+  count (and start) of 0, with data flags of VFIO_IRQ_SET_DATA_NONE disables
+  all interrupts of the index.
 * *data* is an optional field included when the
   VFIO_IRQ_SET_DATA_BOOL flag is present. It contains an array of booleans
   that specify whether the action is to be performed on the corresponding
--- a/lib/irq.c
+++ b/lib/irq.c
@@ -102,8 +102,8 @@ irqs_disable(vfu_ctx_t *vfu_ctx, uint32_t index, uint32_t start, uint32_t count)
     assert(index < VFU_DEV_NUM_IRQS);
     assert(start + count <= vfu_ctx->irq_count[index]);
 
-    vfu_log(vfu_ctx, LOG_DEBUG, "disabling IRQ type %s: start:%u count:%u",
-            vfio_irq_idx_to_str(index), start, count);
+    vfu_log(vfu_ctx, LOG_DEBUG, "disabling IRQ type %s range [%u-%u)",
+            vfio_irq_idx_to_str(index), start, start + count);
 
     if (count == 0) {
         count = vfu_ctx->irq_count[index];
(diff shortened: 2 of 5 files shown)


jlevon:
Fixes #359

lib/irq.c
```diff
@@ -54,52 +54,87 @@ vfio_irq_idx_to_str(int index)
...
+irqs_disable(vfu_ctx_t *vfu_ctx, uint32_t index, uint32_t start, uint32_t count)
 {
-    int *irq_efd = NULL;
-    uint32_t i;
+    size_t i;
+    int *efds;
 
     assert(vfu_ctx != NULL);
     assert(index < VFU_DEV_NUM_IRQS);
+    assert(start + count <= vfu_ctx->irq_count[index]);
```
tmakatos, line 103: ~~This should be a runtime error.~~ Ignore that, I see that `device_set_irqs_validate` ensures that.

lib/irq.c
```diff
@@ -54,52 +54,87 @@ vfio_irq_idx_to_str(int index)
...
+
+    vfu_log(vfu_ctx, LOG_DEBUG, "disabling IRQ type %s: start:%u count:%u",
```
tmakatos, line 105: Replace `start` and `count` with interval notation? It would make output shorter but equally informative.

lib/irq.c
```diff
@@ -228,162 +235,122 @@ dev_set_irqs_validate(vfu_ctx_t *vfu_ctx, struct vfio_irq_set *irq_set)
...
+    // If fd's are provided, ensure they match ->count
+    if (nr_fds != 0 && nr_fds != irq_set->count) {
+        line = __LINE__;
+        goto invalid;
     }
 
     return 0;
+
+invalid:
+    vfu_log(vfu_ctx, LOG_DEBUG, "invalid SET_IRQS (%d): action=%u data_type=%u "
```
tmakatos, line 294: The line trick is definitely cool but OTOH it's not dead simple for the user to understand why it failed.

lib/irq.c
```diff
@@ -228,162 +235,122 @@ dev_set_irqs_validate(vfu_ctx_t *vfu_ctx, struct vfio_irq_set *irq_set)
...
-        return -EINVAL;
+    if ((data_type == VFIO_IRQ_SET_DATA_NONE && irq_set->count == 0) ||
+        (data_type == VFIO_IRQ_SET_DATA_EVENTFD && nr_fds == 0)) {
+        return irqs_disable(vfu_ctx, irq_set->index,
+                            irq_set->start, irq_set->count);
     }
 
-    return dev_get_irqinfo(vfu_ctx, irq_info_in, irq_info_out);
-}
+    vfu_log(vfu_ctx, LOG_DEBUG, "setting IRQ %s flags=%#x start=%u count=%u",
```
tmakatos, line 338: IMO interval notation would be preferable.

test/unit-tests.c
```diff
@@ -1684,7 +1684,157 @@ test_dirty_pages_without_dma(UNUSED void **state)
...
+
+    vfu_ctx.irq_count[VFU_DEV_MSIX_IRQ] = 2048;
+    vfu_ctx.irq_count[VFU_DEV_ERR_IRQ] = 1;
+    vfu_ctx.irq_count[VFU_DEV_REQ_IRQ] = 1;
+    vfu_ctx.irqs = irqs;
+
+    /* validation tests */
+
+    irq_set.argsz = sizeof (irq_set);
+
```
tmakatos, line 1704: It would be helpful to add comments explaining why it's expected to fail, e.g. for the following call to `handle_device_set_irqs` we can have:
```
/* size (0) != irq_set.argsz */

**Review by tmakatos — CHANGES_REQUESTED** [review]
Looks good, minor (mostly optional) suggestions in the main code.

We should state which specific condition each call to `handle_device_set_irqs` tests.

File: docs/vfio-user.rst [hunk 6]
```diff
@@ -1276,6 +1276,8 @@ VFIO IRQ set format
     was sent in the message meta-data. These descriptors will be signalled when
     the action defined by the action flags occurs. In AF_UNIX sockets, the
     descriptors are sent as SCM_RIGHTS type ancillary data.
+    If no file descriptors are provided, this de-assigns the specified
+    previously configured interrupts.
```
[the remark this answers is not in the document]
jlevon: that's right

lib/irq.c
jlevon: done

lib/irq.c
jlevon: They can start at the error message too, this is just to make it easier in context.

lib/irq.c
jlevon: done

test/unit-tests.c
jlevon: hmm, yes

tmakatos (approved):

Issue #390, opened 2021-03-10
document in the spec how hugepages should be handled


tmakatos:
https://libvfio-user.slack.com/archives/C01AFGCSPTR/p1615373270035500?thread_ts=1615194194.026600&cid=C01AFGCSPTR

jlevon:
We should have something that's sort of like this:

```
        iova.iov_base                              iov_end(iova)
        +---------------------------------------------+
        |               paddr (guest PA ranges)       |
        |                                             |
   +----+---------------------------------------------+---+
   |    +---------------------------------------------+   |
   |                                                      |
   |                    vaddr (mapped ranges)             |
   +------------------------------------------------------+
ALIGN(iova.iov_base)                                 ALIGN(iov_end(iova))


   #define pa2va(pa) (pa - iova.iov_base) + vaddr.iov_base
```

There are some example 1MB mappings we get from qemu that are probably illustrative:

```
28469@1615369738.285335:vfio_listener_region_add_skip SKIPPING region_add 0xfec00000 - 0xfec00fff
28469@1615369738.285339:vfio_listener_region_add_skip SKIPPING region_add 0xfed00000 - 0xfed003ff
28469@1615369738.285341:vfio_listener_region_add_skip SKIPPING region_add 0xfee00000 - 0xfeefffff
```

We should also mention overlapping mappings I think?

tmakatos:
> We should also mention overlapping mappings I think?

Mention what about them?

jlevon:
> > We should also mention overlapping mappings I think?
> 
> Mention what about them?

Just that they can exist with an example?

PR #391, opened 2021-03-10
pass both IOVA and virtual address to DMA map/unmap callbacks

tmakatos:


include/libvfio-user.h
```diff
@@ -60,11 +60,8 @@ extern "C" {
 
 #define VFU_DMA_REGIONS  0x10
 
-// FIXME: too common a name?
-typedef uint64_t dma_addr_t;
-
 typedef struct {
-    dma_addr_t dma_addr;
+    void *dma_addr;
```
jlevon, line 64: do we want to implement the TODO below now maybe?

include/libvfio-user.h
```diff
@@ -316,24 +313,28 @@ vfu_setup_device_reset_cb(vfu_ctx_t *vfu_ctx, vfu_reset_cb_t *reset);
  * Function that is called when the guest maps a DMA region. Optional.
  *
  * @vfu_ctx: the libvfio-user context
- * @iova: iova address
- * @len: length
+ * @iova: IOVA address and length
```
jlevon, line 316: maybe expand what this is?

include/libvfio-user.h
```diff
@@ -316,24 +313,28 @@ vfu_setup_device_reset_cb(vfu_ctx_t *vfu_ctx, vfu_reset_cb_t *reset);
...
+ * @vaddr: virtual address and length if the region is mappable, which can be
+ *  different from IOVA due to alignment. If the region is not mappable then
+ *  @vaddr is set to NULL.
```
jlevon, line 319: nit, "@vaddr is NULL" ?

jlevon: should this mention that only the @iova iovec area should be accessed despite the mapping boundaries?

include/libvfio-user.h
```diff
@@ -316,24 +313,28 @@ vfu_setup_device_reset_cb(vfu_ctx_t *vfu_ctx, vfu_reset_cb_t *reset);
...
  * @prot: memory protection used to map region as defined in <sys/mman.h>
  */
-typedef void (vfu_map_dma_cb_t)(vfu_ctx_t *vfu_ctx,
-                                uint64_t iova, uint64_t len, uint32_t prot);
+typedef void (vfu_map_dma_cb_t)(vfu_ctx_t *vfu_ctx, const struct iovec *iova,
```
jlevon, line 322: if we're going to break compat, we might as well do the rename of these now too?

lib/dma.c
```diff
@@ -195,8 +180,9 @@ dma_controller_remove_regions(dma_controller_t *dma)
     for (i = 0; i < dma->nregions; i++) {
         dma_memory_region_t *region = &dma->regions[i];
 
-        vfu_log(dma->vfu_ctx, LOG_INFO, "unmap vaddr=%p IOVA=%lx",
-               region->virt_addr, region->dma_addr);
+        vfu_log(dma->vfu_ctx, LOG_INFO, "unmap IOVA=%#lx-%#lx) vaddr=%#lx-%#lx)",
```
jlevon, line 183: this isn't the right syntax for the ranges? Should be [...), and then you don't want to take 1 off, e.g. [0-4096) is 0, 1, 2,... 4095. Or if you prefer to keep the -1, it should be [0-4095]

lib/dma.c
```diff
@@ -155,20 +134,26 @@ dma_controller_remove_region(dma_controller_t *dma,
...
+            pvaddr = NULL;
+        } else {
+            pvaddr = &region->vaddr;
+        }
+        err = unmap_dma(data, &region->iova, pvaddr);
         if (err != 0) {
             vfu_log(dma->vfu_ctx, LOG_ERR,
                    "failed to notify of removal of DMA region %#lx-%#lx: %s\n",
-                   region->dma_addr, region->dma_addr + region->size,
+                   region->iova.iov_base, region->iova.iov_base + region->iova.iov_len,
```
jlevon, line 151: optional, but is it worth having

#define iov_end(iov) ((iov)->iov_base + (iov)->iov_len)

jlevon (changes requested):

include/libvfio-user.h
tmakatos: > do we want to implement the TODO below now maybe?

Not sure it's a good idea in the end as we'd have to cast `iovec.iov_base` to `int`. IMO we should drop the FIXME.

include/libvfio-user.h
tmakatos: Something like the followig?

"I/O virtual address and length. This is the guest physical address the guest registers for DMA. This address range refers to the guest's memory and is invalid in the user's address space."

jlevon: "physical address range" but yeah, thanks

include/libvfio-user.h
tmakatos: > nit, "@vaddr is NULL" ?

Sure.

> should this mention that only the @iova iovec area should be accessed despite the mapping boundaries?

Definitely. It just occured to me that the user doesn't know where exactly the valid part of vaddr starts. E.g. when using 4K pages and the IOVA is 0x200-400, the vaddr will be e.g. 0xffff0000-0xffff1000. How will the user know that they should access vaddr at offset 0xffff0000+0x200? I think we need to also provide an offset argument in the callbacks?

include/libvfio-user.h
tmakatos: Sure.

lib/dma.c
tmakatos: Agreed.

lib/dma.c
tmakatos: Right.

include/libvfio-user.h
jlevon: OK, but I was thinking that was a typo, and it meant dma_addr+length, which *is* like an iov?

include/libvfio-user.h
```diff
@@ -316,24 +313,33 @@ vfu_setup_device_reset_cb(vfu_ctx_t *vfu_ctx, vfu_reset_cb_t *reset);
...
  * @vfu_ctx: the libvfio-user context
- * @iova: iova address
- * @len: length
+ * @iova: I/O virtual address and length. This is the guest physical address
+ *  range that the guest registers for DMA. This address range refers to the
+ *  guest's memory and is invalid in the user's address space. This memory is
+ *  memory mapped in the user's address space at @vaddr.
+ * @vaddr: virtual address and length if the region is mappable, which can be
+ *  different from IOVA due to alignment. If the region is not mappable then
+ *  @vaddr is NULL.
```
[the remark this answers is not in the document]
jlevon: Actually, this is better done under #390, and this could link to it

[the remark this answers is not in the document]
jlevon: the "due to alignment" bit isn't true, though: I think it should say something like:

vaddr: NULL if the region is not mapped. Otherwise, this is the virtual address and length by which the region is accessible in this process. Both values may have been aligned to a (large) page boundary. For example, an IOVA region [3MB, 6MB), once aligned, may have @vaddr range of [2MB, 6MB), and thus to access the start of the region, an offset of 1MB must be applied.

include/libvfio-user.h
```diff
@@ -316,24 +313,33 @@ vfu_setup_device_reset_cb(vfu_ctx_t *vfu_ctx, vfu_reset_cb_t *reset);
  * Function that is called when the guest maps a DMA region. Optional.
  *
  * @vfu_ctx: the libvfio-user context
- * @iova: iova address
- * @len: length
+ * @iova: I/O virtual address and length. This is the guest physical address
+ *  range that the guest registers for DMA. This address range refers to the
+ *  guest's memory and is invalid in the user's address space. This memory is
+ *  memory mapped in the user's address space at @vaddr.
```
jlevon, line 319: I think you should remove this last sentence in favour of the suggestion below, as it's potentially misleading.

include/libvfio-user.h
```diff
@@ -316,24 +313,33 @@ vfu_setup_device_reset_cb(vfu_ctx_t *vfu_ctx, vfu_reset_cb_t *reset);
  * Function that is called when the guest maps a DMA region. Optional.
  *
  * @vfu_ctx: the libvfio-user context
- * @iova: iova address
- * @len: length
+ * @iova: I/O virtual address and length. This is the guest physical address
+ *  range that the guest registers for DMA. This address range refers to the
```
jlevon, line 317: can we clarify that IOVA === guest PA ranges, as we don't support vIOMMU?

include/libvfio-user.h
```diff
@@ -316,24 +313,33 @@ vfu_setup_device_reset_cb(vfu_ctx_t *vfu_ctx, vfu_reset_cb_t *reset);
...
 /*
  * Function that is called when the guest unmaps a DMA region. The device
  * must release all references to that region before the callback returns.
  * This is required if you want to be able to access guest memory.
  *
  * @vfu_ctx: the libvfio-user context
- * @iova: iova address
- * @len: length
+ * @iova: IOVA address and length
+ * @vaddr: virtual address and length if the region is mappable, which can be
```
jlevon, line 336: probably best to just refer to the register cb description

jlevon (changes requested):

lib/dma.c
```diff
@@ -195,8 +180,9 @@ dma_controller_remove_regions(dma_controller_t *dma)
     for (i = 0; i < dma->nregions; i++) {
         dma_memory_region_t *region = &dma->regions[i];
 
-        vfu_log(dma->vfu_ctx, LOG_INFO, "unmap vaddr=%p IOVA=%lx",
-               region->virt_addr, region->dma_addr);
+        vfu_log(dma->vfu_ctx, LOG_INFO, "unmap IOVA=%#lx-%#lx vaddr=%#lx-%#lx",
+                region->iova.iov_base, iov_end(&region->iova) - 1,
+                region->vaddr.iov_base, iov_end(&region->vaddr) - 1);
 
         _dma_controller_do_remove_region(dma, region);
```
tmakatos, line 187: `dma_controller_remove_regions` unconditionally calls `_dma_controller_do_remove_region`, so it will still break. We should move the `if (region->vaddr.iov_base != MAP_FAILED) ...` check from `dma_controller_remove_region` inside `_dma_controller_do_remove_region`.

jlevon:
I'm definitely thinking we should pass the offset to use too now.

jlevon:
picked up via #396

PR #394, opened 2021-03-16
fix teardown of DMA controller 

jlevon:
If there are regions remaining when we call dma_controller_destroy(), and one of
those regions is not mmap()ed, we were incorrectly trying to unmap it.

Fix this, and also rename the function so its purpose is clearer.

Reported-by: Changpeng Liu <changpeng.liu@intel.com>


jlevon:
I folded this into the big DMA PR #396, as so much has changed.

PR #395, opened 2021-03-17
assert fd != -1 in unmap path

jlevon:
We shouldn't even be trying to unmap if the region doesn't have an associated
file descriptor.

### DIFF
--- a/lib/dma.c
+++ b/lib/dma.c
@@ -105,11 +105,12 @@ _dma_controller_do_remove_region(dma_controller_t *dma,
                region->fd, region->virt_addr,
                region->virt_addr + region->size - 1);
     }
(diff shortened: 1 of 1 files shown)


tmakatos (commented):
As per https://github.com/nutanix/libvfio-user/pull/391#discussion_r594975228 I'm planning to always call `_dma_controller_do_remove_region` and we internally figure out whether or not there's cleanup to be done.

jlevon:
"internally figure out" - isn't that best in dma_controller_remove_region() ? I think the changes I have for this and:

https://github.com/nutanix/libvfio-user/pull/394/files

make sense no?

jlevon:
I folded this into the big DMA PR #396, as so much has changed.

PR #406, opened 2021-03-30
DRAFT: conn reset changes

jlevon:


lib/libvfio-user.c
```diff
@@ -698,26 +699,28 @@ MOCK_DEFINE(get_next_command)(vfu_ctx_t *vfu_ctx, struct vfio_user_header *hdr,
...
+        case ENOMSG:
+            vfu_reset_ctx(vfu_ctx, "closed");
+            return -ENOTCONN;
+
+        case ECONNRESET:
+            vfu_reset_ctx(vfu_ctx, "reset");
+            return -ENOTCONN;
+
+        case EINTR:
+            return -EINTR;
```
swapnili, line 715: Should we have a log message here? Useful for debugging.

lib/tran_sock.c
```diff
@@ -115,10 +115,15 @@ MOCK_DEFINE(tran_sock_send_iovec)(int sock, uint16_t msg_id, bool is_reply,
...
-    ret = sendmsg(sock, &msg, 0);
+    ret = sendmsg(sock, &msg, MSG_NOSIGNAL);
     if (ret == -1) {
+        /* Treat a failed write due to EPIPE the same as a short write. */
+        if (errno == EPIPE) {
+            return -ECONNRESET;
+        }
         return -errno;
+    } else if ((size_t)ret < hdr.msg_size) {
+        return -ECONNRESET;
```
swapnili, line 126: In case of short socket read/write doing vfu_reset_ctx() will detach. But client(qemu) is not aware of this detach, then wont this will lead to blocking server in attach() again and client will never connect.
Instead should we just assert() in such cases until we fix short socket read/write.

lib/libvfio-user.c
```diff
@@ -698,26 +699,28 @@ MOCK_DEFINE(get_next_command)(vfu_ctx_t *vfu_ctx, struct vfio_user_header *hdr,
     /* FIXME get request shouldn't set errno, it should return it as -errno */
     ret = vfu_ctx->tran->get_request(vfu_ctx, hdr, fds, nr_fds);
     if (unlikely(ret < 0)) {
-        if (ret == -EAGAIN || ret == -EWOULDBLOCK) {
+        switch (-ret) {
+        case EAGAIN:
```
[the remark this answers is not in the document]
jlevon: EWOULDBLOCK == EAGAIN on linux

lib/libvfio-user.c
jlevon: yeah, good idea.

lib/tran_sock.c
jlevon: I should expand on this in the commit message. The assumption here is that a short read is more likely to be due to the client disconnecting than it is due to a client bug not sending enough data. Thus, it's better to reset rather than blow up.

jlevon:
moved into individual series

PR #408, opened 2021-03-31
only unmap regions that were actually mapped

jlevon:


jlevon:
NB: this is fixed in the dma branch too, but I needed this to do the reset work.

tmakatos (commented):
Shouldn't then `dma_controller_remove_region` be unconditionally called by `dma_controller_remove_region`?

jlevon:
dma changes are back, dropping this

PR #410, opened 2021-03-31
use MSG_NOSIGNAL for sendmsg()

jlevon:


tmakatos (commented):
Why do we need this?

jlevon:
I don't think it fixes a particular known issue, but generally, we shouldn't end up accidentally generating a SIGPIPE.

tmakatos:
> I don't think it fixes a particular known issue, but generally, we shouldn't end up accidentally generating a SIGPIPE.

Fair enough, let's add this to the commit message for future reference.

jlevon:
msg updated

tmakatos (approved):

PR #411, opened 2021-03-31
->get_request(): return failure if not connected

jlevon:


tmakatos (commented):
`get_msg` would still return the same error (via `recvmsg`), right?

jlevon:
Yeah, but relying on recvmsg(-1) doing the wrong thing correctly, is a little gross.

tmakatos:
> Yeah, but relying on recvmsg(-1) doing the wrong thing correctly, is a little gross.

Thought so, then can we add this to the commit message for future reference?

jlevon:
done

tmakatos (approved):

PR #413, opened 2021-04-01
dma_controller_remove_regions: reset region array

jlevon:


lib/dma.c
```diff
@@ -199,6 +199,9 @@ dma_controller_remove_regions(dma_controller_t *dma)
             assert(region->fd == -1);
         }
     }
+
+    memset(dma->regions, 0, dma->max_regions * sizeof(dma->regions[0]));
```
tmakatos, line 203: This isn't strictly necessary, right?

tmakatos (approved):

lib/dma.c
jlevon: not strictly, no. I did take a look at a unit test but it looked like it involved a bit too much heavy lifting right now.

tmakatos: OK, then can we raise an issue to make sure we don't forget? Or at least drop a FIXME?

PR #415, opened 2021-04-01
implement short read/write, EOF handling 

jlevon:


lib/tran_sock.c
```diff
@@ -173,6 +173,8 @@ get_msg(void *data, size_t len, int *fds, size_t *nr_fds, int sock_fd,
         return -errno;
     } else if (ret == 0) {
         return -ENOMSG;
+    } else if ((size_t)ret < len) {
+        return -ECONNRESET;
```
tmakatos, line 177: Right, this check was missing. Regarding changing the error code from `EINVAL` to `ECONNRESET`, `EINVAL` wasn't a great choice in the first place, however `ECONNRESET` can be misleading? Also, can we have unit tests? (If not in this PR then have at least an open issue to make sure we don't forget.)

jlevon: I think the others in the series should resolve your concern over ECONNRESET?

Re unit tests: I guess, but I'm not sure how much value they would really have here, given our main aim is to detect real-life disconnects. I'd rather think about how we can test this with something more like client/server "real-life".

jlevon:
@tmakatos this is the same as you reviewed here and in #417, merged. We need the ENOTCONN handling to get server to exit cleanly.

tmakatos (approved):

PR #417, opened 2021-04-01
add vfu_reset_ctx()

jlevon:


include/libvfio-user.h
```diff
@@ -137,18 +137,21 @@ int
...
- *   immediatelly returns and the caller is responsible for periodically
+ *   immediately returns and the caller is responsible for periodically
  *   calling again.
  *
  * @vfu_ctx: The libvfio-user context to poll
  *
- * @returns 0 on success, -1 on error. Sets errno.
+ * @returns 0 on success, -1 on error, with errno set as follows:
+ *
+ * EAGAIN/EWOULDBLOCK: no more commands to process
```
tmakatos, line 153: I didn't look at the code very closing but it seems to me that there can be other types of errno, e.g. `EINVAL`. Are we sure these are the only two errnos that can be set?

lib/libvfio-user.c
```diff
@@ -1519,7 +1548,18 @@ vfu_dma_read(vfu_ctx_t *vfu_ctx, dma_sg_t *sg, void *data)
...
-    memcpy(data, dma_recv->data, sg->length); /* FIXME no need for memcpy */
+
+    if (ret == -ENOMSG) {
+        vfu_reset_ctx(vfu_ctx, "closed");
+        ret = -ENOTCONN;
+    } else if (ret == -ECONNRESET) {
+        vfu_reset_ctx(vfu_ctx, "reset");
+        ret = -ENOTCONN;
+    } else if (ret == 0) {
+        /* FIXME no need for memcpy */
```
tmakatos, line 1559: Are we sure we're not missing some other `ret < 0` case? Shall we at least have an `assert`? I've looked at `tran_sock_send_iovec` and it calls `sendmsg` where more errnos can be returned.

include/libvfio-user.h
jlevon: done

lib/libvfio-user.c
jlevon: Not sure what you mean, we're not missing other cases, just if we get those particular two, we treat it as a reset. Others just fall through to the normal error handling.

jlevon:
merged into #415  - we need both changes to correctly handling ENOTCONN in samples/server.c

PR #418, opened 2021-04-01
vfu_reset_ctx(): tear down DMA and IRQs

jlevon:


tmakatos (approved):
Nice fix, let's add unit tests in a future PR.

PR #423, opened 2021-04-06
clean up newlines in logs

jlevon:


tmakatos (commented):
This is what I meant with the other issue regarding newline in logging: I think it's unconventional to assume  that the user's log function will add a newline. Also, we use function `vfu_log` internally in the library but the use can also use this function to log in callbacks. I think it's more intuitive to explicitly add the newline instead of expecting the library to do it for us.

jlevon:
The problem with adding our own newlines is that we can't know if consumers want to wrap the text in something: it might become part of JSON, SPDK_LOG() etc. Thus no newline is most flexible.

tmakatos:
> The problem with adding our own newlines is that we can't know if consumers want to wrap the text in something: it might become part of JSON, SPDK_LOG() etc. Thus no newline is most flexible.

Good point, then we should document this on the function declaration.

tmakatos (approved):

PR #427, opened 2021-04-07
correct DMA region add/remove returns

jlevon:


lib/dma.c
```diff
@@ -346,16 +351,15 @@ MOCK_DEFINE(dma_controller_add_region)(dma_controller_t *dma,
...
         }
     }
 
+    ret = idx;
     dma->nregions++;
 
-    return idx;
-
-err:
-    return -idx - 1;
```
tmakatos, line 358: The patch looks good, however are we changing the semantics of `dma_controller_add_region` when the region can't be mapped?

jlevon: in what way? if dma_map_region() fails, it returns -errno, and that's what we return

tmakatos: I'm talking about this:
```
 * - On failure, a negative integer (-x - 1) where x is the region number
 *   where this region would have been mapped to if the call could succeed
 *   (e.g. due to conflict with existing region).
 ```

jlevon: oh, yes, I need to remove that comment, thanks.

tmakatos (approved):

PR #431, opened 2021-04-07
tran_sock: use ERROR_INT()

jlevon:


lib/tran_sock.c
```diff
@@ -391,61 +391,61 @@ tran_sock_init(vfu_ctx_t *vfu_ctx)
 {
     struct sockaddr_un addr = { .sun_family = AF_UNIX };
     tran_sock_t *ts = NULL;
-    int ret;
+    int err;
```
tmakatos, line 394: Maybe it's best to keep `ret` since `err` is a function defined in `<err.h>`?

lib/tran_sock.c
```diff
@@ -391,61 +391,61 @@ tran_sock_init(vfu_ctx_t *vfu_ctx)
...
-    ret = snprintf(addr.sun_path, sizeof(addr.sun_path), "%s", vfu_ctx->uuid);
-    if (ret >= (int)sizeof(addr.sun_path)) {
-        ret = -ENAMETOOLONG;
-    }
-    if (ret < 0) {
+    err = snprintf(addr.sun_path, sizeof(addr.sun_path), "%s", vfu_ctx->uuid);
+    if (err >= (int)sizeof(addr.sun_path)) {
+        err = ENAMETOOLONG;
+    } else if (err < 0) {
+        err = errno;
```
tmakatos, line 426: According to the manpage `snprintf` doesn't set `errno`.

samples/client.c
```diff
@@ -1155,8 +1136,7 @@ int main(int argc, char *argv[])
     ret = access_region(sock, VFU_PCI_DEV_CFG_REGION_IDX, false, 0, &config_space,
                         sizeof(config_space));
     if (ret < 0) {
-        errx(EXIT_FAILURE, "failed to read PCI configuration space: %s",
-             strerror(-ret));
+        errx(EXIT_FAILURE, "failed to read PCI configuration space");
```
tmakatos, line 1139: `err` instead of `errx`?

tmakatos (approved):
Minor comments.

lib/tran_sock.c
jlevon: sure

lib/tran_sock.c
jlevon: good spot

samples/client.c
jlevon: done

tmakatos (approved):

PR #432, opened 2021-04-07
migration: use ERROR_INT()

jlevon:


lib/migration.c
```diff
@@ -289,7 +289,7 @@ handle_data_offset(vfu_ctx_t *vfu_ctx, struct migration *migr,
         if (is_write) {
             /* TODO writing to read-only registers should be simply ignored */
             vfu_log(vfu_ctx, LOG_ERR, "bad write to migration data_offset");
-            return -EINVAL;
+            return ERROR_INT(EINVAL);
         }
         ret = migr->callbacks.prepare_data(vfu_ctx, offset, NULL);
```
tmakatos, line 294: This is a good opportunity to update `prepare_data` documentation specifing that -1 must be returned on failure and errno must be set.

lib/migration.c
```diff
@@ -442,14 +442,14 @@ migration_region_access(vfu_ctx_t *vfu_ctx, char *buf, size_t count,
             vfu_log(vfu_ctx, LOG_WARNING,
                     "bad access to dead space %#lx-%#lx in migration region",
                     pos, pos + count - 1);
-            return -EINVAL;
+            return ERROR_INT(EINVAL);
         }
 
         pos -= migr->data_offset;
         if (is_write) {
             ret = migr->callbacks.write_data(vfu_ctx, buf, count, pos);
```
tmakatos, line 450: Check that errno is non-zero?

lib/migration.c
```diff
@@ -461,16 +461,13 @@ migration_region_access(vfu_ctx_t *vfu_ctx, char *buf, size_t count,
              * Does this mean that partial reads are not allowed?
              */
             ret = migr->callbacks.read_data(vfu_ctx, buf, count, pos);
-            if (ret == -1) {
-                ret = -errno;
+            if (ret < 0) {
```
tmakatos, line 464: ditto

tmakatos (approved):
Looks good, minor comments and suggestions.

lib/migration.c
jlevon: yep thanks

lib/migration.c
jlevon: as per the other set, I don't think it really makes sense to try and check this.

tmakatos (approved):

tmakatos (approved):

PR #434, opened 2021-04-07
Fix compiler error comparison is always true

mpiszczek:
Changed variable type for getopt() to fix compiler warning when
compiling on arm

This patch fixes this error: 
/libvfio-user/samples/gpio-pci-idio-16.c: In function ‘main’:
/libvfio-user/samples/gpio-pci-idio-16.c:160:44: error: comparison is always true due to limited range of data type [-Werror=type-limits]
  160 |     while ((opt = getopt(argc, argv, "v")) != -1) {
      |                                            ^~
cc1: all warnings being treated as errors

Seen when building on arm with gcc (Ubuntu 9.3.0-17ubuntu1~20.04) 9.3.0

### DIFF
--- a/samples/gpio-pci-idio-16.c
+++ b/samples/gpio-pci-idio-16.c
@@ -141,7 +141,7 @@ main(int argc, char *argv[])
 {
     int ret;
     bool verbose = false;
-    char opt;
+    int opt;
     struct sigaction act = { .sa_handler = _sa_handler };
     vfu_ctx_t *vfu_ctx;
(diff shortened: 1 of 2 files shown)


jlevon:
Patch itself looks good, thanks. Do you mind signing off your commit (git commit -s, as per https://en.wikipedia.org/wiki/Developer_Certificate_of_Origin) ?

Somehow the CI won't run either, hopefully that's just a blip.

jlevon (approved):

mpiszczek:
Sure. I will fix that.

PR #445, opened 2021-04-16
vfu_realize_ctx(): fix default PCI config space region

jlevon:


lib/libvfio-user.c
```diff
@@ -1227,6 +1225,10 @@ vfu_create_ctx(vfu_trans_t trans, const char *path, int flags, void *pvt,
         goto err_out;
     }
 
+    for (i = 0; i< vfu_ctx->nr_regions; i++) {
```
tmakatos, line 1228: nit: space missing before `<`

tmakatos (approved):

lib/libvfio-user.c
jlevon: done

PR #449, opened 2021-04-19
reduce VFU_MAX_CAPS

jlevon:


swapnili (approved):
Should we add an,
  TODO: Dynamically allocate capabilities so that we do not have this limit.
I hope linux do not enforces such limit

jlevon:
I don't think we need the TODO yet: if we have a user who needs more, we can deal with it then. I'd not want to use allocations unless we really need to

PR #450, opened 2021-04-22
cap_write_pm(): remove assert()

jlevon:


lib/pci_caps.c
```diff
@@ -141,7 +141,7 @@ cap_write_pm(vfu_ctx_t *vfu_ctx, struct pci_cap *cap, char * buf,
         if (count != sizeof(struct pc)) {
             return ERROR_INT(EINVAL);
         }
-        assert(false); /* FIXME implement */
+        vfu_log(vfu_ctx, LOG_ERR, "FIXME: write to pmcp::pc unimplemented");
```
swapnili, line 144: Why not return ENOTSUP?

swapnili (approved):
Thanks

jlevon:
Actually, it needs a further fix I think, sorry.

PR #454, opened 2021-04-23
python tests: add VFIO_USER_DEVICE_GET_INFO

jlevon:


test/py/test_device_get_info.py
```diff
@@ -0,0 +1,88 @@
...
+
+    hdr = vfio_user_header(VFIO_USER_DEVICE_GET_INFO, size=len(payload))
+    sock.send(hdr + payload)
+    vfu_run_ctx(ctx)
+    get_reply(sock, expect=errno.EINVAL)
+
+    # bad argsz
+
+    # struct vfio_device_info
+    payload = struct.pack("IIII", 8, 0, 0, 0)
```
tmakatos, line 63: Shall we add a class for vfio_device_info instead of manually packing?

tmakatos (approved):

test/py/test_device_get_info.py
jlevon: Maybe we can consider doing that sort of cleanup. There are a few tests where we purposely don't create the right incoming struct sizes.

I agree it's not very readable in this form, but I'm not sure yet what the best approach is as to what it might look like. OK to revisit after more experience with the patch series?

tmakatos: sure

PR #464, opened 2021-05-05
fix dma unregister callback during region removal

jlevon:


include/libvfio-user.h
```diff
@@ -655,7 +655,9 @@ vfu_unmap_sg(vfu_ctx_t *vfu_ctx, const dma_sg_t *sg,
              struct iovec *iov, int cnt);
 
 /**
- * Read from the dma region exposed by the client.
+ * Read from the dma region exposed by the client. This can be used as an
+ * alternative to vfu_map_sg(), if the region is not directly mappable, or DMA
+ * callbacks have not been provided.
```
swapnili, line 660: Should we be more clear and call it as "DMA notification callbacks"

include/libvfio-user.h
```diff
@@ -667,7 +669,9 @@ int
 vfu_dma_read(vfu_ctx_t *vfu_ctx, dma_sg_t *sg, void *data);
 
 /**
- * Write to the dma region exposed by the client.
+ * Write to the dma region exposed by the client.  This can be used as an
+ * alternative to vfu_map_sg(), if the region is not directly mappable, or DMA
+ * callbacks have not been provided.
```
swapnili, line 674: ditto "DMA notification callbacks"

lib/dma.c
```diff
@@ -178,21 +178,32 @@ MOCK_DEFINE(dma_controller_remove_region)(dma_controller_t *dma,
 }
 
 void
-dma_controller_remove_regions(dma_controller_t *dma)
+dma_controller_remove_regions(dma_controller_t *dma,
```
swapnili, line 181: I think we should rename this func to dma_controller_remove_all_regions()
This name really confuses with dma_controller_remove_region()

lib/dma.c
```diff
@@ -207,11 +218,6 @@ dma_controller_remove_regions(dma_controller_t *dma)
 void
 dma_controller_destroy(dma_controller_t *dma)
 {
-    if (dma == NULL) {
```
swapnili, line 210: Why following snippet is removed?

jlevon: no need to check NULL when we only call free()

lib/dma.c
jlevon: sure, done.

include/libvfio-user.h
jlevon: done

include/libvfio-user.h
jlevon: done

lib/dma.c
```diff
@@ -207,11 +218,6 @@ dma_controller_remove_regions(dma_controller_t *dma)
...
-        return;
-    }
-
-    dma_controller_remove_regions(dma);
```
swapnili, line 214: Sorry I think previous comment was not clear. I mean to say that why call to dma_controller_remove_regions() is been removed.

swapnili (approved):

PR #471, opened 2021-05-07
python tests: add VFIO_USER_DEVICE_GET_REGION_INFO

jlevon:


test/py/test_device_get_region_info.py
```diff
@@ -0,0 +1,223 @@
...
+#  DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+#  (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+#  SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+#  CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+#  LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY
+#  OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH
+#  DAMAGE.
+#
+
+from libvfio_user import *
```
tmakatos, line 30: I've been badly burned from `import *` in the past, if you think it's fine to do here because this is a test then OK.

test/py/test_device_get_region_info.py
```diff
@@ -0,0 +1,223 @@
...
+ctx = None
+sock = None
+
+def test_device_get_region_info_setup():
+    global ctx, sock
+
+    ctx = vfu_create_ctx(flags=LIBVFIO_USER_FLAG_ATTACH_NB)
+    assert ctx != None
+
+    ret = vfu_setup_region(ctx, index=VFU_PCI_DEV_BAR1_REGION_IDX, size=4096,
```
tmakatos, line 43: Just a thought: wouldn't it be nicer to make function like e.g. `vfu_setup_region` as method of the `ctx` class?

test/py/test_device_get_region_info.py
```diff
@@ -0,0 +1,223 @@
...
+                           flags=(VFU_REGION_FLAG_RW | VFU_REGION_FLAG_MEM))
+    assert ret == 0
+
+    f = open("/tmp/vfio-user.map", "wb")
```
tmakatos, line 47: nit: use `tempfile.NamedTemporaryFile` instead of hard-coded paths?

test/py/test_device_get_region_info.py
```diff
@@ -0,0 +1,223 @@
...
+                          VFIO_REGION_INFO_FLAG_WRITE |
+                          VFIO_REGION_INFO_FLAG_MMAP |
+                          VFIO_REGION_INFO_FLAG_CAPS)
+    assert info.index == VFU_PCI_DEV_BAR2_REGION_IDX
+    assert info.cap_off == 0
+    assert info.size == 0x10000
+    assert info.offset == vfu_region_to_offset(VFU_PCI_DEV_BAR2_REGION_IDX)
+
+    # skip reading the SCM_RIGHTS
+    disconnect_client(ctx, sock)
```
tmakatos, line 147: Why do we need this?

test/py/test_device_get_region_info.py
```diff
@@ -0,0 +1,223 @@
...
+
+    hdr = vfio_user_header(VFIO_USER_DEVICE_GET_REGION_INFO, size=len(payload))
+    sock.send(hdr + payload)
+    vfu_run_ctx(ctx)
+    get_reply(sock, expect=errno.EINVAL)
+
+def test_device_get_region_info_bad_argsz():
+
+    # struct vfio_region_info
+    payload = struct.pack("IIIIQQ", 8, 0, VFU_PCI_DEV_BAR1_REGION_IDX, 0, 0, 0)
```
tmakatos, line 86: Would it be possible to generate the payload in a more self-explanatory way, e.g. via some constructor or function? Otherwise we'd have to refer to the struct definition and make sure the order and sizing of the arguments to `struct.pack` is correct.

test/py/test_device_get_region_info.py
```diff
@@ -0,0 +1,223 @@
...
+
+    assert info.argsz == 80
+    assert info.cap_off == 32
+
+    assert cap.id == VFIO_REGION_INFO_CAP_SPARSE_MMAP
+    assert cap.version == 1
+    assert cap.next == 0
+    assert cap.nr_areas == 2
+
+    assert area1.offset == 0x2000
```
tmakatos, line 175: I suppose this assumes that `test_device_get_region_info_setup` will have run first. Maybe introduce a setup function or add decorators so that each test is independent?

test/py/libvfio_user.py
```diff
@@ -70,6 +71,17 @@
 VFIO_DEVICE_FLAGS_RESET = (1 << 0)
 VFIO_DEVICE_FLAGS_PCI = (1 << 1)
 
+VFIO_REGION_INFO_FLAG_READ = (1 << 0)
```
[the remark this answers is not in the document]
jlevon: See line :31 above

```
# Note that we don't use enum here, as class.value is a little verbose
```

test/py/test_device_get_region_info.py
jlevon: Yeah, I'm not going to worry about something adjacent to the other part that's only for test, and never delivered anywhere.

test/py/test_device_get_region_info.py
jlevon: tempfile.TemporaryFile will do in fact, thanks.

test/py/test_device_get_region_info.py
jlevon: I'd prefer to keep the functions looking as much like as the real C ones as possible I think.

test/py/test_device_get_region_info.py
jlevon: Swapnil had the same complaint, and the answer is yes - but I have that cleanup later in the series.

test/py/test_device_get_region_info.py
jlevon: It's just a cheesy way to not have to have code reading the file descriptors from the socket.

test/py/test_device_get_region_info.py
jlevon: All of these individual files require the relevant _setup() to have run first.
By default, pytest runs the tests in serial order as written, which is exactly what we need.
Adding decorators for each individual test would involve a ton more scaffolding, and I don't really see the need - outside of valgrind, the `make pytest` is not slow at all.

tmakatos (approved):

PR #476, opened 2021-05-08
check max message size for initial negotiation too

jlevon:


jlevon:
cc @tmakatos  - I'm thinking it still makes sense to have a limit on this part, but not as part of the spec. We don't need to support arbitrarily long JSON.

Issue #477, opened 2021-05-10
fix bad cast


tmakatos:
From the discussion on Slack:

> Changpeng Liu  2 days ago
@Thanos I root caused the issue, in the function dma_controller_dirty_page_get(), you shouldn't use int  for len parameter, the client is using __u64  for the size.
Changpeng Liu  2 days ago
after changing the int to u64, I didn't see the error logs again.

tmakatos:
[qemu.log](https://github.com/nutanix/libvfio-user/files/6450793/qemu.log)
[spdk.log](https://github.com/nutanix/libvfio-user/files/6450794/spdk.log)

PR #457, opened 2021-04-28
migration: various dirty page tracking fixes

tmakatos:

PR #479, opened 2021-05-10
dma: Use correct len type

swapnili:
vfio_iommu_type1_dirty_bitmap_get.size is of type __u64
dma_controller_dirty_page_get() recieves it as int.

Fixes: #477


include/libvfio-user.h
```diff
@@ -66,7 +66,7 @@ typedef void *vfu_dma_addr_t;
 typedef struct {
     vfu_dma_addr_t dma_addr;
     int region;
-    int length;
+    uint64_t length;
```
jlevon, line 69: but dma_init_sg(), and various friends, still take a 32-bit value, so I'm not sure how this works.

Also, we'd need to do something about vfu_dma_read / vfu_dma_write() - right now they are truncating a potentially 64-bit sg->length

tmakatos (changes requested):
IMO there should be a unit test where the bug triggers without the fix and then a commit that fixes the bug. The SPDK log I've attached contains the values that in the dirty pages message, so it should be trivially to repro.

include/libvfio-user.h
swapnili: Shouldn't it be safe to put 32bit(unsigned) value into 64bit(unsigned)?
Or should we also change dma_init_sg() and friends to receive length in uint64_t

For vfu_dma_read / vfu_dma_write() right now the protocol expects 'count' to be uint32_t. How about having value check for <UINT32_MAX?

swapnili:
> IMO there should be a unit test where the bug triggers without the fix and then a commit that fixes the bug. The SPDK log I've attached contains the values that in the dirty pages message, so it should be trivially to repro.

Sure I will try to repro it with Unit test

include/libvfio-user.h
jlevon: Was the original bug from using int rather uint32_t ?
You made a change such that dma_controller_dirty_page_get() takes a uint64_t, but without changing dma_init_sg() etc. you're dropping the top 32 bits on the floor

tmakatos:
FYI Changpeng was using this branch: https://github.com/nutanix/libvfio-user/pull/457

swapnili:
Please have another look. Also added a UT. 
I am gonna fix vfu_dma_read / vfu_dma_write() in separate patch.

jlevon (approved):
OK, please remember to file a bug on vfu_dma_read/write() as it affects the spec.

swapnili:
https://github.com/nutanix/libvfio-user/issues/495

PR #484, opened 2021-05-10
cleanup: Use uint*_t typedefs

swapnili:
To be consistent with other declarations use uint*_t typedefs.

### DIFF
--- a/include/vfio-user.h
+++ b/include/vfio-user.h
@@ -163,33 +163,33 @@ struct vfio_user_irq_info {
 #define VFIO_REGION_SUBTYPE_MIGRATION           (1)
 
 struct vfio_device_migration_info {
-	__u32 device_state;         /* VFIO device state */
-	__u32 reserved;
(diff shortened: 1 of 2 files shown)


include/vfio-user.h
```diff
@@ -163,33 +163,33 @@ struct vfio_user_irq_info {
 #define VFIO_REGION_SUBTYPE_MIGRATION           (1)
 
 struct vfio_device_migration_info {
-	__u32 device_state;         /* VFIO device state */
-	__u32 reserved;
-	__u64 pending_bytes;
-	__u64 data_offset;
-	__u64 data_size;
+	uint32_t device_state;         /* VFIO device state */
```
jlevon, line 166: I left these with __ on purpose, as they're from the kernel headers.

swapnili: Then maybe we should use them from linux/vfio.h?

Issue #478, opened 2021-05-10
stop requiring fixed region offsets


jlevon:
According to the spec today, socket clients are supposed to mmap() device regions at the given "offset". For libvfio-user, the offset is fixed (see vfu_region_to_offset()) based upon the region index.

This mimics traditional vfio - where there is one 'fd' that exposes each region at a different fixed offset. However, we have no need to do the same at all: the underlying fd(s) come from the API user (e.g. spdk) - there's no need for them to be the same file, and no reason to have fixed offsets.

We should fix this, probably by:

1) clarifying the spec
2) removing the fixed region/offset mapping
3) making clients specify an offset in vfu_setup_region() (for SPDK's nvmf, this would just be '0')

jlevon:
fixed now

PR #485, opened 2021-05-10
fix region offset handling

jlevon:
The specification states that the region offset given in the region info should
be used as the "offset" when mmap()ing the region from the client side. However,
the library instead implemented a fixed offset scheme  similar to that of vfio -
and no clients actually set up the file like that.

Instead, let servers define their own offsets, and pass them through to clients
as is. It's up to the server to decide how its backing file or files is
organized.

### DIFF
--- a/samples/server.c
+++ b/samples/server.c
@@ -430,7 +430,7 @@ int main(int argc, char *argv[])
         }
     };
     vfu_ctx_t *vfu_ctx;
-    FILE *bar1_fp, *migr_fp;
+    FILE *fp;
     const vfu_migration_callbacks_t migr_callbacks = {
         .version = VFU_MIGR_CALLBACKS_VERS,
         .transition = &migration_device_state_transition,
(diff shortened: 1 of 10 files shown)


jlevon:
Fixes #478

samples/server.c
```diff
@@ -491,15 +491,29 @@ int main(int argc, char *argv[])
...
+    migr_regs_size = vfu_get_migr_register_area_size();
+    migr_data_size = page_align(bar1_size + sizeof(time_t));
+    migr_size = migr_regs_size + migr_data_size;
+
+    /*
+     * We choose to use a single file which contains both bar1 and the migration
+     * registers. They could also be completely different files.
+     */
+    if (ftruncate(fileno(fp), server_data.bar1_size + migr_size) == -1) {
         err(EXIT_FAILURE, "failed to truncate BAR1 file");
```
tmakatos, line 513: nit: BAR1

samples/server.c
```diff
@@ -491,15 +491,29 @@ int main(int argc, char *argv[])
...
     }
     server_data.bar1 = mmap(NULL, server_data.bar1_size, PROT_READ | PROT_WRITE,
-                            MAP_SHARED, fileno(bar1_fp), 0);
+                            MAP_SHARED, fileno(fp), 0);
     if (server_data.bar1 == MAP_FAILED) {
         err(EXIT_FAILURE, "failed to mmap BAR1");
```
tmakatos, line 518: ditto

tmakatos (approved):

samples/server.c
jlevon: Yes, thanks, I re-ordered the comments a bit too.

samples/server.c
jlevon: this one is fine I think, we are mapping BAR1 from the backing file only.

PR #493, opened 2021-05-12
Use linux vfio header

swapnili:


tmakatos (commented):
Shall we from which exact version we copied the file? Preferably somewhere from https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/tree?

swapnili:
I've already mentioned the exact version in the commit message.
It's v5.8 https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/tag/?h=v5.8

tmakatos:
> I've already mentioned the exact version in the commit message.
> It's v5.8 https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/tag/?h=v5.8

Ah yes, you have.

tmakatos (approved):

PR #497, opened 2021-05-14
Fix dma read write count

### LINKED ISSUE #495: Fix vfu_dma_read/write() region size to uint64_t
DMA region size is maxed to uint64_t.
So vfu_dma_read/write() message count should also be uint64_t.

swapnili:
Fixes #495


jlevon (approved):
Looks good. I'll pick the qemu-kvm change as part of the rest of incompat changes I'm making.

PR #498, opened 2021-05-14
add a gcov target

jlevon:


Makefile
```diff
@@ -107,7 +112,11 @@ pytest-valgrind: all
 
 endif
 
+ifdef WITH_GCOV
+test: all pytest
```
swapnili, line 116: nit: tab

jlevon: this is a rule, I don't think we want it indented

lib/CMakeLists.txt
```diff
@@ -65,24 +65,13 @@ set_target_properties(vfio-user-static PROPERTIES
...
-
-add_library_ut(pci_caps pci_caps.c)
-add_library_ut(dma dma.c)
-add_library_ut(irq irq.c)
-add_library_ut(libvfio-user libvfio-user.c)
-add_library_ut(migration migration.c)
-add_library_ut(pci pci.c)
-add_library_ut(tran_sock tran_sock.c)
+add_library(pci_caps OBJECT pci_caps.c)
+add_library(dma OBJECT OBJECT dma.c)
```
swapnili, line 69: Double OBJECT type?

lib/CMakeLists.txt
```diff
@@ -65,24 +65,13 @@ set_target_properties(vfio-user-static PROPERTIES
...
-endfunction(add_library_ut)
-
-add_library_ut(pci_caps pci_caps.c)
-add_library_ut(dma dma.c)
-add_library_ut(irq irq.c)
-add_library_ut(libvfio-user libvfio-user.c)
-add_library_ut(migration migration.c)
-add_library_ut(pci pci.c)
-add_library_ut(tran_sock tran_sock.c)
+add_library(pci_caps OBJECT pci_caps.c)
```
swapnili, line 68: Can all of these add_library() moved closer to set(LIBOBJS, ) at line:40 ?
This way if someone adds new file then its clear to also update set(LIBOBJS)

Makefile
swapnili: Aah yes

swapnili (approved):

PR #499, opened 2021-05-14
validate_irq_subindex: skip check of vfu_ctx pointer

jlevon:


swapnili (changes requested):
How about also adding assert(vfu_ctx != NULL) in vfu_irq_trigger()?

swapnili (approved):

PR #517, opened 2021-05-21
Handle support of PCI FLR capability

swapnili:
If device supports FLR cap then call vfu_reset_cb_t when FLR is
initiated by client.

### DIFF
--- a/lib/pci_caps.c
+++ b/lib/pci_caps.c
@@ -276,14 +276,17 @@ handle_px_pxdc_write(vfu_ctx_t *vfu_ctx, struct pxcap *px,
         vfu_log(vfu_ctx, LOG_DEBUG, "MRRS set to %d", p->mrrs);
     }
 
-    if (p->iflr && vfu_ctx->reset) {
+    if (p->iflr) {
         if (px->pxdcap.flrc == 0) {
-            vfu_log(vfu_ctx, LOG_ERR, "FLR capability is not supported\n");
+            vfu_log(vfu_ctx, LOG_ERR, "FLR capability is not supported");
             return ERROR_INT(EINVAL);
         }
-        vfu_log(vfu_ctx, LOG_DEBUG,
-            "initiate function level reset");
-        return vfu_ctx->reset(vfu_ctx, VFU_RESET_PCI_FLR);
+        if (vfu_ctx->reset != NULL) {
+            vfu_log(vfu_ctx, LOG_DEBUG, "initiate function level reset");
+            return vfu_ctx->reset(vfu_ctx, VFU_RESET_PCI_FLR);
+        } else {
+            vfu_log(vfu_ctx, LOG_ERR, "FLR callback is not implemented");
+        }
     }
 
     return 0;
--- a/lib/pci_caps.c
+++ b/lib/pci_caps.c
@@ -276,9 +276,14 @@ handle_px_pxdc_write(vfu_ctx_t *vfu_ctx, struct pxcap *px,
         vfu_log(vfu_ctx, LOG_DEBUG, "MRRS set to %d", p->mrrs);
     }
 
-    if (p->iflr) {
+    if (p->iflr && vfu_ctx->reset) {
+        if (px->pxdcap.flrc == 0) {
+            vfu_log(vfu_ctx, LOG_ERR, "FLR capability is not supported\n");
+            return ERROR_INT(EINVAL);
+        }
         vfu_log(vfu_ctx, LOG_DEBUG,
             "initiate function level reset");
+        return vfu_ctx->reset(vfu_ctx, VFU_RESET_PCI_FLR);
     }
 
     return 0;
--- a/test/py/test_pci_caps.py
+++ b/test/py/test_pci_caps.py
@@ -294,9 +294,12 @@ def test_pci_cap_write_pmcs():
 
     disconnect_client(ctx, sock)
 
+reset_flag = -1
 @c.CFUNCTYPE(c.c_int, c.c_void_p, c.c_int)
 def vfu_reset_cb(ctx, reset_type):
     assert reset_type == VFU_RESET_PCI_FLR or reset_type == VFU_RESET_LOST_CONN
+    global reset_flag
+    reset_flag = reset_type
     return 0
 
 def test_pci_cap_write_px():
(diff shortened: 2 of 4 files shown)


test/py/test_pci_caps.py
```diff
@@ -293,9 +294,29 @@ def test_pci_cap_write_pmcs():
...
 def test_pci_cap_write_px():
-    # FIXME
-    pass
+    sock = connect_client(ctx)
+    ret = vfu_setup_device_reset_cb(ctx, vfu_reset_cb)
+    assert ret == 0
+
+    #flrc
+    cap = struct.pack("ccHHcc52c", to_byte(PCI_CAP_ID_EXP), b'\0', 0, 0, b'\0',
+                      b'\x02', *[b'\0' for _ in range(52)])
```
jlevon, line 309: doesn't just
```
b'\0' * 52
```

work ?

test/py/test_pci_caps.py
```diff
@@ -293,9 +294,29 @@ def test_pci_cap_write_pmcs():
...
+    pos = vfu_pci_add_capability(ctx, pos=cap_offsets[5], flags=0, data=cap)
+    assert pos == cap_offsets[5]
+
+    #iflr
+    offset=cap_offsets[5] + 8
+    data=b'\x00\x80'
+    write_region(ctx, sock, VFU_PCI_DEV_CFG_REGION_IDX, offset=offset,
+                 count=len(data), data=data)
+
```
jlevon, line 318: can we assert that the reset CB was actually called?

jlevon (changes requested):
Thanks for adding the tests, I just have two minor comments

jlevon:
I noticed that in my centos73 guest, I have:

```
	Capabilities: [48] Express (v2) Endpoint, MSI 00
		DevCap:	MaxPayload 128 bytes, PhantFunc 0, Latency L0s <64ns, L1 <1us
			ExtTag- AttnBtn+ AttnInd- PwrInd- RBE- FLReset- SlotPowerLimit 128.000W
```

Isn't that a bit weird? spdk is supposed to have set flrc, so why isn't it supported?

lib/pci_caps.c
```diff
@@ -276,9 +276,14 @@ handle_px_pxdc_write(vfu_ctx_t *vfu_ctx, struct pxcap *px,
         vfu_log(vfu_ctx, LOG_DEBUG, "MRRS set to %d", p->mrrs);
     }
 
-    if (p->iflr) {
+    if (p->iflr && vfu_ctx->reset) {
```
tmakatos, line 279: nit: `vfu_ctx->reset != NULL`

tmakatos, line 279: How do we handle the case where `iflr` is written to but `vfu_ctx->reset == NULL`? Shall we at least log something?

tmakatos (commented):
minor nits

lib/pci_caps.c
jlevon: yeah, this would be an API user error, so it should log a warning at least.
Perhaps we should add to the reset docs that if they set flrc, they must handle that type of reset CB?

test/py/test_pci_caps.py
swapnili: It doesn't work,  

>>> pack("52c", b'\0' * 52)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
struct.error: pack expected 52 items for packing (got 1)

test/py/test_pci_caps.py
swapnili: Sure

test/py/test_pci_caps.py
```diff
@@ -100,7 +100,8 @@ def test_pci_cap_setup_cb():
     # NB: note 4-byte alignment of vsc2
     PCI_STD_HEADER_SIZEOF + PCI_PM_SIZEOF + 8,
     0x80,
-    0x90
+    0x90,
+    0xa0
```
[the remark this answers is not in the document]
swapnili: That is the offset where PCI_CAP_ID_EXP will be added.

test/py/test_pci_caps.py
```diff
@@ -293,9 +294,29 @@ def test_pci_cap_write_pmcs():
...
+
+    #flrc
+    cap = struct.pack("ccHHcc52c", to_byte(PCI_CAP_ID_EXP), b'\0', 0, 0, b'\0',
+                      b'\x02', *[b'\0' for _ in range(52)])
+    pos = vfu_pci_add_capability(ctx, pos=cap_offsets[5], flags=0, data=cap)
+    assert pos == cap_offsets[5]
+
+    #iflr
+    offset=cap_offsets[5] + 8
+    data=b'\x00\x80'
```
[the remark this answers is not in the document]
swapnili: Done

lib/pci_caps.c
swapnili: Done

swapnili: IMO that should be device implementation bug, as it advertises flrc but did not implemented it.   Added error message for such cases.

swapnili:
> I noticed that in my centos73 guest, I have:
> 
> ```
> 	Capabilities: [48] Express (v2) Endpoint, MSI 00
> 		DevCap:	MaxPayload 128 bytes, PhantFunc 0, Latency L0s <64ns, L1 <1us
> 			ExtTag- AttnBtn+ AttnInd- PwrInd- RBE- FLReset- SlotPowerLimit 128.000W
> ```
> 
> Isn't that a bit weird? spdk is supposed to have set flrc, so why isn't it supported?

Yes I also saw that. I do see same lspci output even if we modify test-lspci.sh to support FLR. Opened an issue to track it
https://github.com/nutanix/libvfio-user/issues/524

test/py/test_pci_caps.py
jlevon: right, makes sense.

test/py/test_pci_caps.py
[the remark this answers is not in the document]
jlevon: not changed?

jlevon (changes requested):

test/py/test_pci_caps.py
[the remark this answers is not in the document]
swapnili: oops sorry somehow its missed

jlevon (approved):

PR #518, opened 2021-05-24
log dirty pages for new DMA regions as well

tmakatos:
New DMA regions don't get their pages tracked if dirty page logging has
already been started, this patch fixes this bug.

### DIFF
--- a/lib/dma.c
+++ b/lib/dma.c
@@ -256,6 +256,33 @@ dma_map_region(dma_controller_t *dma, dma_memory_region_t *region)
     return 0;
 }
 
+static ssize_t
+get_bitmap_size(size_t region_size, size_t pgsize)
+{
+    if (pgsize == 0) {
+        return ERROR_INT(EINVAL);
+    }
+    if (region_size < pgsize) {
+        return ERROR_INT(EINVAL);
+    }
+    size_t nr_pages = (region_size / pgsize) + (region_size % pgsize != 0);
+    return ROUND_UP(nr_pages, sizeof(uint64_t) * CHAR_BIT) / CHAR_BIT;
+}
+
+static int
+dirty_page_logging_start_on_region(dma_memory_region_t *region, size_t pgsize)
+{
+    ssize_t size = get_bitmap_size(region->info.iova.iov_len, pgsize);
+    if (size < 0) {
+        return size;
+    }
+    region->dirty_bitmap = calloc(size, sizeof(char));
+    if (region->dirty_bitmap == NULL) {
+        return ERROR_INT(errno);
+    }
+    return 0;
+}
+
 int
 MOCK_DEFINE(dma_controller_add_region)(dma_controller_t *dma,
                                        vfu_dma_addr_t dma_addr, size_t size,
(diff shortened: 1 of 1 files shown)


lib/dma.c
```diff
@@ -256,6 +256,33 @@ dma_map_region(dma_controller_t *dma, dma_memory_region_t *region)
     return 0;
 }
 
+static ssize_t
+get_bitmap_size(size_t region_size, size_t pgsize)
+{
+    if (pgsize == 0) {
```
jlevon, line 262: nit, can these both be asserts, then you can return size_t ?

lib/dma.c
```diff
@@ -340,7 +367,23 @@ MOCK_DEFINE(dma_controller_add_region)(dma_controller_t *dma,
...
-        int ret = dma_map_region(dma, region);
+        int ret;
+
+        /*
+         * TODO introduce a function that tells whether dirty page logging is
+         * enabled
+         */
+        if (dma->dirty_pgsize != 0) {
+            if (dirty_page_logging_start_on_region(region, dma->dirty_pgsize) < 0) {
+                /*
```
jlevon, line 378: I'm not sure it's worth fixing this just for a calloc() failure?

lib/dma.c
```diff
@@ -351,7 +394,9 @@ MOCK_DEFINE(dma_controller_add_region)(dma_controller_t *dma,
                 vfu_log(dma->vfu_ctx, LOG_WARNING,
                         "failed to close fd %d: %m", region->fd);
             }
-
+            if (region->dirty_bitmap != NULL) {
```
jlevon, line 397: nit, don't need to check NULL before a free()

lib/dma.c
```diff
@@ -445,24 +477,16 @@ int dma_controller_dirty_page_logging_start(dma_controller_t *dma, size_t pgsize
...
-
-        bitmap_size = get_bitmap_size(region->info.iova.iov_len, pgsize);
-
-        if (bitmap_size < 0) {
-            return bitmap_size;
-        }
-        region->dirty_bitmap = calloc(bitmap_size, sizeof(char));
-        if (region->dirty_bitmap == NULL) {
-            int ret = errno;
+        if (dirty_page_logging_start_on_region(region, pgsize) < 0) {
```
jlevon, line 480: don't we want this to be idempotent?

jlevon (changes requested):

lib/dma.c
tmakatos: We can, but `get_bitmap_size` has two callers so we'd have to duplicate checks there and I'd like to avoid that. However, in case of error we can return 0 therefore we _can_ change the return type to `size_t`. How does this sound?

lib/dma.c
tmakatos: I've noticed that dirty page tracking errors aren't fatal (e.g. worst that can happen is that migration fails but the VM keeps running on the source) but failing to add a DMA region would be. I wanted this documented to make sure we don't forget, shall I drop the `TODO` and keep the comment?

lib/dma.c
tmakatos: ok

lib/dma.c
tmakatos: Do you mean effective ignoring calls to `dma_controller_dirty_page_logging_start` if dirty page tracking has already started? If so, then this is checked right before this `for` loop.

lib/dma.c
jlevon: I think it's fine as is, then.

jlevon (approved):

PR #521, opened 2021-05-24
support VFIO_DMA_UNMAP_FLAG_GET_DIRTY_BITMAP

tmakatos:


tmakatos:
Need to restore `argsz`. Also, need to check whether it breaks compatibility with QEMU.

lib/libvfio-user.c
```diff
@@ -546,14 +547,46 @@ handle_dma_unmap(vfu_ctx_t *vfu_ctx, vfu_msg_t *msg,
...
 
-    if (region->flags == 0) {
+    if (region->flags == VFIO_DMA_UNMAP_FLAG_GET_DIRTY_BITMAP) {
+        if (msg->in_size != sizeof(struct vfio_user_dma_region) + sizeof(struct vfio_user_bitmap)) {
+            vfu_log(vfu_ctx, LOG_ERR, "bad message size %#lx", msg->in_size);
+            return ERROR_INT(EINVAL);
+        }
+        /*
+         * TODO this could be a separate function, but the implementation is
+         * temprorary anyway since we're moving dirty page tracking out of
```
jlevon, line 563: nit, "temporary"

lib/libvfio-user.c
```diff
@@ -546,14 +547,46 @@ handle_dma_unmap(vfu_ctx_t *vfu_ctx, vfu_msg_t *msg,
...
+        if (msg->out_iovecs == NULL) {
+            return -1;
+        }
+        msg->out_iovecs->iov_base = malloc(region->bitmap->size);
+        if (msg->out_iovecs == NULL) {
+            free(msg->out_iovecs);
+            msg->out_iovecs = NULL;
+            return ERROR_INT(ENOMEM);
+        }
+        memcpy(msg->out_iovecs->iov_base, bitmap, region->bitmap->size);
```
jlevon, line 586: I don't get why we're doing a copy here rather than just populating your allocation?

jlevon (changes requested):
Don't you want to remove the separate dirty pages message here too?

lib/libvfio-user.c
tmakatos: right

lib/libvfio-user.c
tmakatos: I'm not sure I understand your question, do you mean why not doing `msg->out_iovecs->iov_base = bitmap`? If so, the reason is that `dma_controller_remove_region` is supposed to free `bitmap`, but I just discovered that it doesn't, so that's a memleak. I suppose we can do:
```
msg->out_iovecs->iov_base = region->bitmap;
region->bitmap = NULL;
...
Right?

**Comment by tmakatos** [comment]
> Don't you want to remove the separate dirty pages message here too?

Which separate dirty pages message?

File: lib/libvfio-user.c (hunk 2 above)
Comment (jlevon) [reply lib/libvfio-user.c:586]: We don't free iovecs in freemsg() either, but we can add that facility. It's certainly better than two allocations?

**Comment by jlevon** [comment]
> Which separate dirty pages message?

Sorry, ignore that.

File: lib/libvfio-user.c [hunk 3]
```diff
@@ -546,14 +547,46 @@ handle_dma_unmap(vfu_ctx_t *vfu_ctx, vfu_msg_t *msg,
...
+        if (ret < 0) {
+            vfu_log(vfu_ctx, LOG_ERR, "failed to get dirty page bitmap: %m");
+            return -1;
+        }
+        msg->out_iovecs = malloc(sizeof(struct iovec));
+        if (msg->out_iovecs == NULL) {
+            return -1;
+        }
+        msg->out_iovecs->iov_base = malloc(region->bitmap->size);
+        if (msg->out_iovecs == NULL) {
```
jlevon, line 581: btw, it's probably going to change, but you have the wrong check here

lib/libvfio-user.c
jlevon: Actually you don't need to use out_iovecs at all since it's just one entry. Why not out_data?

tmakatos: good point

tmakatos:
@jlevon I've addressed your comments, please take another look.

lib/libvfio-user.c
```diff
@@ -524,25 +524,57 @@ handle_dma_unmap(vfu_ctx_t *vfu_ctx, vfu_msg_t *msg,
...
-            region->offset, region->prot, region->flags);
+    snprintf(rstr, sizeof(rstr), "[%#lx, %#lx) offset=%#lx prot=%#x flags=%#x",
+             region->addr, region->addr + region->size,
+             region->offset, region->prot, region->flags);
 
     vfu_log(vfu_ctx, LOG_DEBUG, "removing DMA region %s", rstr);
 
-    if (region->flags != 0) {
+    if (region->flags == VFIO_DMA_UNMAP_FLAG_GET_DIRTY_BITMAP) {
+        if (msg->in_size != sizeof(struct vfio_user_dma_region) + sizeof(struct vfio_user_bitmap)) {
```
jlevon, line 547: nit, use sizeof(*region) like :534
Also, this should be msg->in_size < not != too

lib/libvfio-user.c
```diff
@@ -524,25 +524,57 @@ handle_dma_unmap(vfu_ctx_t *vfu_ctx, vfu_msg_t *msg,
...
+                                            (vfu_dma_addr_t)region->addr,
+                                            region->size,
+                                            region->bitmap->pgsize,
+                                            region->bitmap->size,
+                                            &bitmap);
+        if (ret < 0) {
+            vfu_log(vfu_ctx, LOG_ERR, "failed to get dirty page bitmap: %m");
+            return -1;
+        }
+        msg->out_data = malloc(region->bitmap->size);
```
jlevon, line 566: I thought you were going to do:

```
msg->out_data = region->bitmap;
msg->out_size = region->bitmap->size;
region->bitmap = NULL;
```

??

lib/libvfio-user.c
```diff
@@ -524,25 +524,57 @@ handle_dma_unmap(vfu_ctx_t *vfu_ctx, vfu_msg_t *msg,
...
+        if (msg->out_data == NULL) {
+            return ERROR_INT(ENOMEM);
+        }
+        memcpy(msg->out_data, bitmap, region->bitmap->size);
+        msg->out_size = region->bitmap->size;
+    } else if (region->flags == 0) {
+        if (msg->in_size != sizeof(struct vfio_user_dma_region)) {
```
jlevon, line 573: this branch should be removed I think?

jlevon (changes requested):

lib/libvfio-user.c
tmakatos: I thought I was, but I was wrong in my example: `region->bitmap` is the struct in the request where the size/pgsize is set, not the actual dirty page bitmap. The dirty page bitmap lives in `dma_memory_region_t` and I'm not sure it's worth the effort to expose this as we're removing dirty page logging from the DMA controller in the mid-term. So I opted to allocate memory as that seems the easiest thing to do for now.

lib/libvfio-user.c
tmakatos: ok

lib/libvfio-user.c
tmakatos: yes

tmakatos:
@jlevon addressed your comments, please take another look.

jlevon (approved):
thanks

PR #523, opened 2021-05-24
restore argsz for DMA map/unmap

tmakatos:

### DIFF
--- a/docs/vfio-user.rst
+++ b/docs/vfio-user.rst
@@ -536,15 +536,17 @@ The request payload for this message is an entry of the following format:
 +-------------+--------+-------------+
 | Name        | Offset | Size        |
 +=============+========+=============+
-| Address     | 0      | 8           |
+| argsz       | 0      | 4           |
 +-------------+--------+-------------+
-| Size        | 8      | 8           |
+| Address     | 4      | 8           |
 +-------------+--------+-------------+
-| Offset      | 16     | 8           |
+| Size        | 12     | 8           |
 +-------------+--------+-------------+
-| Protections | 24     | 4           |
+| Offset      | 20     | 8           |
 +-------------+--------+-------------+
-| Flags       | 28     | 4           |
+| Protections | 28     | 4           |
++-------------+--------+-------------+
+| Flags       | 32     | 4           |
 +-------------+--------+-------------+
 |             | +-----+------------+ |
 |             | | Bit | Definition | |
@@ -553,6 +555,7 @@ The request payload for this message is an entry of the following format:
 |             | +-----+------------+ |
 +-------------+--------+-------------+
 
+* *argsz* is the size of the above format.
 * *Address* is the base DMA address of the region.
 * *Size* is the size of the region.
 * *Offset* is the file offset of the region with respect to the associated file
@@ -564,7 +567,7 @@ The request payload for this message is an entry of the following format:
   * *Mappable* indicates that the region can be mapped via the mmap() system
     call using the file descriptor provided in the message meta-data.
 
-This structure is 32 bytes in size, so the message size is 16 + 32 == 48 bytes.
+This structure is 32 bytes in size, so the message size is 16 + 36 bytes.
 
 If the DMA region being added can be directly mapped by the server, a file
 descriptor must be sent as part of the message meta-data. On ``AF_UNIX``
@@ -597,15 +600,17 @@ The request payload for this message is an entry of the following format:
 +--------------+--------+---------------------------------------+
 | Name         | Offset | Size                                  |
 +==============+========+=======================================+
-| Address      | 0      | 8                                     |
+| argsz        | 0      | 4                                     |
 +--------------+--------+---------------------------------------+
-| Size         | 8      | 8                                     |
+| Address      | 4      | 8                                     |
 +--------------+--------+---------------------------------------+
-| Offset       | 16     | 8                                     |
+| Size         | 12     | 8                                     |
 +--------------+--------+---------------------------------------+
-| Protections  | 24     | 4                                     |
+| Offset       | 20     | 8                                     |
 +--------------+--------+---------------------------------------+
-| Flags        | 28     | 4                                     |
+| Protections  | 28     | 4                                     |
++--------------+--------+---------------------------------------+
+| Flags        | 32     | 4                                     |
 +--------------+--------+---------------------------------------+
 |              | +-----+--------------------------------------+ |
 |              | | Bit | Definition                           | |
@@ -614,6 +619,7 @@ The request payload for this message is an entry of the following format:
 |              | +-----+--------------------------------------+ |
 +--------------+--------+---------------------------------------+
 
+* *argsz* is the maximum size of the reply payload.
 * *Address* is the base DMA address of the DMA region.
 * *Size* is the size of the DMA region.
 * *Offset* must be zero (ignored)
@@ -624,10 +630,12 @@ The request payload for this message is an entry of the following format:
     must be populated before unmapping the DMA region. The client must provide
     a ``struct vfio_user_bitmap`` immediately following this table entry.
 
-If ``VFIO_DMA_UNMAP_FLAG_GET_DIRTY_BITMAP`` is set in Flags, the size of the
-total request message is: 16 + 32 + 16.
-Otherwise, if ``VFIO_DMA_UNMAP_FLAG_GET_DIRTY_BITMAP`` is not set in Flags, the
-size of the total request message is: 16 + 32.
+The size of request message depends on whether or not the
+``VFIO_DMA_UNMAP_FLAG_GET_DIRTY_BITMAP`` is set in Flags:
+
+* If not set, the size of the total request message is: 16 + 36.
+
+* If set, the size of the total request message is: 16 + 36 + 16.
 
 .. _VFIO bitmap format:
 
@@ -653,18 +661,14 @@ mapped then the server must release all references to that DMA region before
 replying, which potentially includes in-flight DMA transactions. Removing a
 portion of a DMA region is possible.
 
-The payload depends on whether or not ``VFIO_DMA_UNMAP_FLAG_GET_DIRTY_BITMAP``
-is set in Flags in the request:
+The server responds with the original DMA entry in the request. If
+``VFIO_DMA_UNMAP_FLAG_GET_DIRTY_BITMAP`` is set in Flags in the request, then
+the server includes the ``struct vfio_user_bitmap`` sent in the request,
+followed by the corresponding dirty page bitmap, where each bit represents
+one page of size ``struct vfio_user_bitmap.pgsize``.
 
-* ``VFIO_DMA_UNMAP_FLAG_GET_DIRTY_BITMAP`` is set in Flags in the request:
-  The server responds with a payload that contains the original
-  ``struct vfio_user_bitmap``  sent in the request, followed by the
-  corresponding dirty page bitmap, where each bit represents one page of size
-  ``struct vfio_user_bitmap.pgsize``.  The size of the total reply message is:
-  16 + 16 + size of dirty page bitmap.
-
-* ``VFIO_DMA_UNMAP_FLAG_GET_DIRTY_BITMAP`` is not set in Flags in the request:
-   There is no payload. The size of the reply message is: 16.
+The total size of the total reply message is:
+16 + 36 + (16 + ``strcut vfio_user_bitmap.size`` if ``VFIO_DMA_UNMAP_FLAG_GET_DIRTY_BITMAP`` is set).
 
 ``VFIO_USER_DEVICE_GET_INFO``
 -----------------------------
--- a/docs/vfio-user.rst
+++ b/docs/vfio-user.rst
@@ -664,7 +664,7 @@ replying, which potentially includes in-flight DMA transactions. Removing a
 portion of a DMA region is possible.
 
 The server responds with the original DMA entry in the request. If the
-*get dirty page bitmap* bit is set in Flags in the request, then
+*get dirty page bitmap* bit is set in flags in the request, then
 the server also includes the `VFIO bitmap`_ structure sent in the request,
 followed by the corresponding dirty page bitmap, where each bit represents
 one page of size *pgsize* in `VFIO bitmap`_ .
--- a/docs/vfio-user.rst
+++ b/docs/vfio-user.rst
@@ -531,8 +531,7 @@ or if the client uses a vIOMMU.
 Request
 ^^^^^^^
 
-The request payload for this message is an array of table entries of the
-following format:
+The request payload for this message is an entry of the following format:
 
 +-------------+--------+-------------+
 | Name        | Offset | Size        |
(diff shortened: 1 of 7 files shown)


tmakatos:
This change breaks compatibility with QEMU; before merging we need to update the relevant QEMU code.

lib/libvfio-user.c
```diff
@@ -487,7 +487,7 @@ handle_dma_map(vfu_ctx_t *vfu_ctx, vfu_msg_t *msg,
     assert(msg != NULL);
     assert(region != NULL);
 
-    if (msg->in_size != sizeof(*region)) {
+    if (msg->in_size != sizeof(*region) || msg->in_size != region->argsz) {
```
jlevon, line 490: See my recent changes e.g. line :358, this should be:

```
if (msg->in_size < sizeof(*region) || in_info->argsz < sizeof(*region)) {
```

so client can specify a larger expected argsz

lib/libvfio-user.c
```diff
@@ -540,7 +540,7 @@ handle_dma_unmap(vfu_ctx_t *vfu_ctx, vfu_msg_t *msg,
     assert(msg != NULL);
     assert(region != NULL);
 
-    if (msg->in_size < sizeof(*region)) {
+    if (msg->in_size < sizeof(*region) || msg->in_size < region->argsz) {
```
jlevon, line 543: ditto, this is incorrectly requiring an "in" buffer of ->argsz, but that value defines the max *reply* payload

lib/libvfio-user.c
```diff
@@ -554,7 +554,8 @@ handle_dma_unmap(vfu_ctx_t *vfu_ctx, vfu_msg_t *msg,
     vfu_log(vfu_ctx, LOG_DEBUG, "removing DMA region %s", rstr);
 
     if (region->flags == VFIO_DMA_UNMAP_FLAG_GET_DIRTY_BITMAP) {
-        if (msg->in_size != sizeof(struct vfio_user_dma_region) + sizeof(struct vfio_user_bitmap)) {
+        if (msg->in_size != sizeof(struct vfio_user_dma_region) + sizeof(struct vfio_user_bitmap)
```
jlevon, line 557: ditto

lib/libvfio-user.c
```diff
@@ -587,7 +588,8 @@ handle_dma_unmap(vfu_ctx_t *vfu_ctx, vfu_msg_t *msg,
         msg->out_iovecs->iov_len = region->bitmap->size;
         msg->nr_out_iovecs = 1;
     } else if (region->flags == 0) {
-        if (msg->in_size != sizeof(struct vfio_user_dma_region)) {
+        if (msg->in_size != sizeof(struct vfio_user_dma_region)
```
jlevon, line 591: ditto

jlevon (changes requested):

lib/libvfio-user.c
```diff
@@ -587,7 +588,8 @@ handle_dma_unmap(vfu_ctx_t *vfu_ctx, vfu_msg_t *msg,
         msg->out_iovecs->iov_len = region->bitmap->size;
```
jlevon, line 588: don't we need the reply to have the same vfio_user_dma_region header? So it can specify the actual needed argsz?

jlevon (changes requested):

lib/libvfio-user.c
tmakatos: ok

lib/libvfio-user.c
tmakatos: ok

lib/libvfio-user.c
tmakatos: Actually, I understand why `argsz` can be larger, but why `msg->in_size`? Won't we be ptentially masking errors if the client sends stuff that we ignore?

lib/libvfio-user.c
tmakatos: In this particular case, `region->argsz` must be large enough to accomodate the respone, which is `sizeof(struct vfio_user_bitmap) + region->bitmap->size`, right?

lib/libvfio-user.c
tmakatos: I don't think so, the client knows the size since it's the one that sets `region->bitmap->size` in the first place, so it knows that `argsz` but be at least `sizeof(strcut vfio_user_bitmap) + region->bitmap->size`.

lib/libvfio-user.c
jlevon: It must be at least sizeof(*region): in that case, we need to reply with an out_region, with argsz set to the full required size, as per the spec. That's why you need to have the region struct in the reply as I think I mentioned elsewhere?

So it's clearer as:

```
if (msg->in_size < sizeof(*region) || region->argsz < sizeof(*region))
```

like, for example, handle_device_get_irq_info()

IOW, if we can't even read the in_region, or reply with an out_region, we can't progress.

Then in the bitmap case, we check that ->in_size also has a vfio_user_bitmap, then in the reply when bitmap is requested

```
out_region->argsz = sizeof(*region) + sizeof(struct vfio_user_bitmap) + region->bitmap->size
```

and, *if* it fits into region->argsz, add on the vfio_user_bitmap and the bitmap itself

Does that make sense?

jlevon: handle_device_get_region_info is a better example

lib/libvfio-user.c
jlevon: I don't expect it to be common to use these argsz semantics, but I also don't see a good reason to differ from the expectations of the spec and the current code elsewhere. We should respond

tmakatos:
This PR should also contain the relevant doc changes where `argsz` is part of the request which must conform to the following:
> Message sizes
Some requests have an argsz field. In a request, it defines the maximum expected reply payload size, which should be at least the size of the fixed reply payload headers defined here. For messages that don't have a reply payload, it defines the size of the incoming request (not including the standard header); otherwise it's not related to the request message size.
>
> In a reply, the server sets argsz field to the size needed for a full payload size. This may be less than the requested maximum size. This may be larger than the requested maximum size: in that case, the full payload is not included in the reply, but the argsz field in the reply indicates the needed size, allowing a client to allocate a larger buffer for holding the reply before trying again.

tmakatos:
We should restore the original `vfio_iommu_type1_dma_map` / `vfio_iommu_type1_dma_unmap` according to https://github.com/oracle/qemu/commit/9baf27a8fb99392a4ff85071670cd40cec1ef4af.

tmakatos:
@jlevon I've rewritten DMA map/unmap to match VFIO's, please take another look.

include/vfio-user.h
```diff
@@ -114,18 +114,27 @@ struct vfio_user_bitmap {
     char data[];
 } __attribute__((packed));
 
-struct vfio_user_dma_region {
-    uint64_t                addr;
-    uint64_t                size;
-    uint64_t                offset;
-    uint32_t                prot;
+/* based on strcut vfio_iommu_type1_dma_map */
```
jlevon, line 117: "struct"

include/vfio-user.h
```diff
@@ -114,18 +114,27 @@ struct vfio_user_bitmap {
...
+struct vfio_user_dma_map {
+    uint32_t argsz;
+#define VFIO_USER_F_DMA_REGION_MAPPABLE (1 << 2)
```
jlevon, line 120: Can we just have VFIO_USER_F_DMA_REGION_READ/WRITE too please? It's a bit weird to mix-and-match IMO

lib/libvfio-user.c
```diff
@@ -468,29 +468,39 @@ consume_fd(int *fds, size_t nr_fds, size_t index)
 
 int
 handle_dma_map(vfu_ctx_t *vfu_ctx, vfu_msg_t *msg,
-               struct vfio_user_dma_region *region)
+               struct vfio_user_dma_map *region)
```
jlevon, line 471: can we rename "region" now too? to "dma_map" or something?

lib/libvfio-user.c
```diff
@@ -468,29 +468,39 @@ consume_fd(int *fds, size_t nr_fds, size_t index)
...
     int ret;
+    uint32_t prot = 0;
 
     assert(vfu_ctx != NULL);
     assert(msg != NULL);
     assert(region != NULL);
 
-    if (msg->in_size != sizeof(*region)) {
+    if (msg->in_size < sizeof(*region) || region->argsz < sizeof(*region)) {
         vfu_log(vfu_ctx, LOG_ERR, "bad size of DMA map region %zu",
```
jlevon, line 483: probably need to report argsz in the error message now too

lib/libvfio-user.c
```diff
@@ -499,11 +509,16 @@ handle_dma_map(vfu_ctx_t *vfu_ctx, vfu_msg_t *msg,
                     rstr);
             return -1;
         }
+        region->flags &= ~VFIO_USER_F_DMA_REGION_MAPPABLE;
+    }
```
jlevon, line 513: nit, do else if { ... ?

lib/libvfio-user.c
```diff
@@ -521,31 +536,34 @@ handle_dma_map(vfu_ctx_t *vfu_ctx, vfu_msg_t *msg,
 
 int
 handle_dma_unmap(vfu_ctx_t *vfu_ctx, vfu_msg_t *msg,
-                 struct vfio_user_dma_region *region)
+                 struct vfio_user_dma_unmap *region)
```
jlevon, line 539: nit, rename variable too?

lib/libvfio-user.c
```diff
@@ -521,31 +536,34 @@ handle_dma_map(vfu_ctx_t *vfu_ctx, vfu_msg_t *msg,
...
+             region->addr, region->addr + region->size, region->flags);
 
     vfu_log(vfu_ctx, LOG_DEBUG, "removing DMA region %s", rstr);
 
+    msg->out_size = sizeof(*region);
+
     if (region->flags == VFIO_DMA_UNMAP_FLAG_GET_DIRTY_BITMAP) {
-        if (msg->in_size < sizeof(*region) + sizeof(struct vfio_user_bitmap)) {
-            vfu_log(vfu_ctx, LOG_ERR, "bad message size %#lx", msg->in_size);
+        if (msg->in_size < sizeof(*region) + sizeof(*region->bitmap)
```
jlevon, line 563: first check if fine, but you're supposed to be able to ask for a smaller argsz in the request,
and then get the actual expected argsz in the reply (which would just be *region in this case). Don't expect clients to need this ever, but I think we should keep argsz semantics for consistency.

lib/private.h
```diff
@@ -181,11 +181,11 @@ exec_command(vfu_ctx_t *vfu_ctx, vfu_msg_t *msg);
 
 int
 handle_dma_map(vfu_ctx_t *vfu_ctx, vfu_msg_t *msg,
-               struct vfio_user_dma_region *dma_regions);
+               struct vfio_user_dma_map *region);
```
jlevon, line 184: nit, here, below, rename region

samples/client.c
```diff
@@ -1154,11 +1154,11 @@ int main(int argc, char *argv[])
     dma_region_fds = alloca(sizeof(*dma_region_fds) * nr_dma_regions);
 
     for (i = 0; i < nr_dma_regions; i++) {
+        dma_regions[i].argsz =  sizeof(struct vfio_user_dma_map);
```
jlevon, line 1157: extra space?

samples/client.c
```diff
@@ -1213,18 +1213,19 @@ int main(int argc, char *argv[])
      * unmap the first group of the DMA regions
      */
     {
```
jlevon, line 1215: don't need this scope now

samples/client.c
```diff
@@ -1213,18 +1213,19 @@ int main(int argc, char *argv[])
...
-        struct vfio_user_dma_region r[server_max_fds];
-        memcpy(r, dma_regions, sizeof(r));
-        for (i = 0; i < (int)ARRAY_SIZE(r); i++) {
-            r[i].flags = 0;
-            ret = tran_sock_msg(sock, 7, VFIO_USER_DMA_UNMAP, &r[i],
-                                sizeof(struct vfio_user_dma_region),
-                                NULL, NULL, 0);
+        for (i = 0; i < server_max_fds; i++) {
```
jlevon, line 1216: should be nr_dma_regions

test/unit-tests.c
```diff
@@ -121,7 +121,8 @@ setup(void **state UNUSED)
 static void
 test_dma_map_mappable_without_fd(void **state UNUSED)
 {
-    struct vfio_user_dma_region dma_region = {
+    struct vfio_user_dma_map dma_region = {
```
jlevon, line 124: (generally) rename as dma_map/unmap as needed?

jlevon (changes requested):
Looks great, I have a few nits and one minor bug in client.c

docs/vfio-user.rst
```diff
@@ -531,40 +531,49 @@ or if the client uses a vIOMMU.
...
+| argsz       | 0      | 4           |
 +-------------+--------+-------------+
-| Size        | 8      | 8           |
-+-------------+--------+-------------+
-| Offset      | 16     | 8           |
-+-------------+--------+-------------+
-| Protections | 24     | 4           |
-+-------------+--------+-------------+
-| Flags       | 28     | 4           |
+| Flags       | 4      | 4           |
```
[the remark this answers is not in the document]
tmakatos: ok

include/vfio-user.h
tmakatos: ok

lib/libvfio-user.c
tmakatos: sure

include/vfio-user.h
tmakatos: sure

lib/libvfio-user.c
tmakatos: sure

samples/client.c
tmakatos: right

lib/private.h
tmakatos: ok

samples/client.c
tmakatos: This is from the time we could unmap multiple regions in a single request, `server_max_fds == nr_dma_regions / 2`.

test/unit-tests.c
tmakatos: sure

lib/libvfio-user.c
tmakatos: ok

lib/libvfio-user.c
tmakatos: ah yes

samples/client.c
tmakatos: indeed

lib/libvfio-user.c
tmakatos: ok

tmakatos:
@jlevon I've addressed your comments, please take another look.

docs/vfio-user.rst
```diff
@@ -653,18 +663,14 @@ mapped then the server must release all references to that DMA region before
...
-is set in Flags in the request:
-
-* ``VFIO_DMA_UNMAP_FLAG_GET_DIRTY_BITMAP`` is set in Flags in the request:
-  The server responds with a payload that contains the original
-  ``struct vfio_user_bitmap``  sent in the request, followed by the
-  corresponding dirty page bitmap, where each bit represents one page of size
-  ``struct vfio_user_bitmap.pgsize``.  The size of the total reply message is:
-  16 + 16 + size of dirty page bitmap.
+The server responds with the original DMA entry in the request. If the
+*get dirty page bitmap* bit is set in Flags in the request, then
```
jlevon, line 667: nit, s/Flags/*flags*/

lib/libvfio-user.c
```diff
@@ -521,61 +533,73 @@ handle_dma_map(vfu_ctx_t *vfu_ctx, vfu_msg_t *msg,
...
+        vfu_log(vfu_ctx, LOG_ERR, "bad flags=%#x", dma_unmap->flags);
         return ERROR_INT(ENOTSUP);
     }
+
+    msg->out_data = malloc(msg->out_size);
+    if (msg->out_data == NULL) {
+        return ERROR_INT(ENOMEM);
+    }
+    memcpy(msg->out_data, dma_unmap, sizeof(*dma_unmap));
+    ((struct vfio_user_dma_unmap*)msg->out_data)->argsz = argsz;
```
jlevon, line 593: nit, space before *

jlevon (changes requested):
tiny nits!

jlevon (approved):
Thanks! Please make sure the new commit message covers all these changes, as it's definitely not just argsz any more :)

tmakatos:
@jlevon I just realized that my implementation is wrong: if `VFIO_DMA_UNMAP_FLAG_GET_DIRTY_BITMAP` is set but `argsz` is too small then we don't return the bitmap but proceed with unmapping the DMA region. The client can't retry the request since the DMA region will have been removed, so there's no way to retrieve the dirty page bitmap. I'll check to see what VFIO does in this case.

jlevon:
Good point, I hadn't thought this through properly. We can't return a payload *and* an error, so I think we should go back to your earlier version that requires a valid argsz after all. Sorry for asking you to do that :)

tmakatos:
> Good point, I hadn't thought this through properly. We can't return a payload _and_ an error, so I think we should go back to your earlier version that requires a valid argsz after all. Sorry for asking you to do that :)

no problem

lib/libvfio-user.c
```diff
@@ -521,31 +533,41 @@ handle_dma_map(vfu_ctx_t *vfu_ctx, vfu_msg_t *msg,
...
+    if (dma_unmap->flags == VFIO_DMA_UNMAP_FLAG_GET_DIRTY_BITMAP) {
+        if (msg->in_size < sizeof(*dma_unmap) + sizeof(*dma_unmap->bitmap)
+            || dma_unmap->argsz < sizeof(*dma_unmap) + sizeof(*dma_unmap->bitmap) + dma_unmap->bitmap->size) {
+            vfu_log(vfu_ctx, LOG_ERR, "bad message size=%#lx argsz=%#x",
+                    msg->in_size, dma_unmap->argsz);
+
+            /*
+             * Ideally we should set argsz in the reply and fail the request
+             * with a struct vfio_user_dma_unmap payload, however this isn't
+             * currently supported. Instread, we simply fail the request,
```
jlevon, line 568: "Instead"

jlevon (approved):
Thanks; one typo you can fix as you push, don't need another cycle.

PR #530, opened 2021-05-26
spec: allow exactly one DMA region to be added/removed per command

tmakatos:


jlevon (approved):
There's more changes coming, but let's get this in now

Issue #529, opened 2021-05-26
broken make install


karlatec:
When using master branch with commit `0a04aeae6ae904762c3abafb8de9708a9728d503` it is not possible to do `make install`.

```
[libvfio-user]$ make -j16
-- The C compiler identification is GNU 10.2.1
-- Check for working C compiler: /usr/bin/cc
-- Check for working C compiler: /usr/bin/cc - works
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Detecting C compile features
-- Detecting C compile features - done
-- Found PkgConfig: /usr/bin/pkg-config (found version "1.6.3")
-- Checking for module 'json-c'
--   Found json-c, version 0.13.1
-- Configuring done
-- Generating done
-- Build files have been written to: /home/klateck/work/spdk/libvfio-user/build/dbg
Scanning dependencies of target pci_caps
Scanning dependencies of target irq
Scanning dependencies of target libvfio-user
Scanning dependencies of target migration
Scanning dependencies of target tran_sock
Scanning dependencies of target pci
Scanning dependencies of target dma
Scanning dependencies of target client
Scanning dependencies of target unit-tests
[  9%] Building C object lib/CMakeFiles/migration.dir/migration.o
[  9%] Building C object lib/CMakeFiles/pci.dir/pci.o
[  9%] Building C object lib/CMakeFiles/pci_caps.dir/pci_caps.o
[ 12%] Building C object lib/CMakeFiles/libvfio-user.dir/libvfio-user.o
[ 16%] Building C object lib/CMakeFiles/irq.dir/irq.o
[ 19%] Building C object lib/CMakeFiles/dma.dir/dma.o
[ 22%] Building C object lib/CMakeFiles/tran_sock.dir/tran_sock.o
[ 25%] Building C object samples/CMakeFiles/client.dir/client.o
[ 29%] Building C object samples/CMakeFiles/client.dir/__/lib/tran_sock.o
[ 32%] Building C object samples/CMakeFiles/client.dir/__/lib/migration.o
[ 35%] Building C object test/CMakeFiles/unit-tests.dir/unit-tests.o
[ 38%] Building C object test/CMakeFiles/unit-tests.dir/mocks.o
[ 48%] Building C object test/CMakeFiles/unit-tests.dir/__/lib/libvfio-user.o
[ 48%] Building C object test/CMakeFiles/unit-tests.dir/__/lib/dma.o
[...]

PR #531, opened 2021-05-26
fix `make install`

jlevon:


jlevon:
Fixes #529

tmakatos (approved):

tmakatos:
Maybe we should add a CI test for `make install`?

tmakatos (approved):

PR #532, opened 2021-05-26
Cleanup and fix structs padding

swapnili:
In case of bitfields compiler may use data type to allocate struct size
and add additional paddings. Instead use data type which is closest to
the total size of the bitfields.
This patch also uses uint* version consistently throughout.

### DIFF
--- a/include/pci_caps/common.h
+++ b/include/pci_caps/common.h
@@ -59,17 +59,17 @@ struct vsc {
  * PCI Express extended capability header.
  */
 struct pcie_ext_cap_hdr {
-    unsigned int id:16;
-    unsigned int version:4;
-    unsigned int next:12;
+    uint32_t id:16;
+    uint32_t version:4;
+    uint32_t next:12;
 } __attribute__((packed));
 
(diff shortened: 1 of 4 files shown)


tmakatos:
Are you saying it is possible that, e.g.:
```
struct pcie_ext_cap_hdr {
    unsigned int id:16;
    unsigned int version:4;
    unsigned int next:12;
} __attribute__((packed));
```
Maybe not end up being 4 bytes while
```
struct pcie_ext_cap_hdr {
    uint32_t id:16;
    uint32_t version:4;
    uint32_t next:12;
} __attribute__((packed));
```
is guranteed to be 4 bytes?

jlevon:
I already had to fix something similar in 13e3f8d38bc23c2a95654c46addea84ca6de365a

It's not unsigned int versus uint32_t, but about using the correct width at all (i.e. the important changes are unsigned int -> uint16_t).

Swapnil, does this fix flrc?

swapnili:
No the flr is still not fixed by this. This fix was just while code browsing.

jlevon (approved):

PR #534, opened 2021-05-27
Fix struct pxcap

### LINKED ISSUE #524: lspci shows FLR not supported
Even if FLRC is set lspci shows it is not supported,

Capabilities: [48] Express (v2) Endpoint, MSI 00
	DevCap:	MaxPayload 128 bytes, PhantFunc 0, Latency L0s <64ns, L1 <1us
	ExtTag- AttnBtn+ AttnInd- PwrInd- RBE- FLReset- SlotPowerLimit 128.000W

swapnili:
* Added missing reserved bits and renamed per to rer nameing as the
  nvme specs
* Add pxcap capability in lspci test

Fixes #524 

### DIFF
--- a/include/pci_caps/px.h
+++ b/include/pci_caps/px.h
@@ -58,12 +58,13 @@ struct pxdcap {
     uint32_t etfs:1;
     uint32_t l0sl:3;
     uint32_t l1l:3;
-    uint32_t per:1;
-    uint32_t res1:2;
+    uint32_t res1:3;
+    uint32_t rer:1;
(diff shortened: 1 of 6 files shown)


jlevon (approved):
oops!
Does a "reset" sysfs entry show in up a linux guest now? Does it work?

swapnili:
Yes it works now. I can also trigger reset from the sysfs.

> [root@localhost ~]# echo 1 > /sys/bus/pci/drivers/nvme/0000\:00\:07.0/reset

PR #547, opened 2021-06-01
samples/client: free bitmap data

jlevon:


jlevon:
I don't really know why this doesn't fail CI, but I don't want to spend time on that really (I'd rather move the whole dance to python ;)

tmakatos (approved):

PR #553, opened 2021-06-07
drop mappable flag from DMA map

tmakatos:
The flags field belongs to VFIO and it's not a good idea to reuse as new
VFIO flags can break things. Instead, we derive whether or not a region
is mappable if a file descriptor is passed.


lib/libvfio-user.c
```diff
@@ -508,16 +508,16 @@ handle_dma_map(vfu_ctx_t *vfu_ctx, vfu_msg_t *msg,
         dma_map->flags &= ~VFIO_USER_F_DMA_REGION_WRITE;
     }
 
-    if (dma_map->flags == VFIO_USER_F_DMA_REGION_MAPPABLE) {
+    if (dma_map->flags != 0) {
```
swapnili, line 511: I think we must be strict and fail if client passes unknown flags?

swapnili (changes requested):
VFIO_USER_F_DMA_REGION_MAPPABLE is also mentioned in include/libvfio-user.h and also used in test/py/
Can you also fix them?

tmakatos:
> VFIO_USER_F_DMA_REGION_MAPPABLE is also mentioned in include/libvfio-user.h and also used in test/py/
> Can you also fix them?

Sure.

lib/libvfio-user.c
tmakatos: Ah I forgot the `return` statement.

tmakatos:
@swapnili addressed your comments, please take another look.

swapnili (approved):

tmakatos:
Since this is a spec change I'll wait for @jlevon to double check.

jlevon (approved):

Issue #542, opened 2021-05-28
make pytest succeeds even if python-3-pytest is not installed


tmakatos:
We should at least print that no tests have been run because python-3-pytest is not installed.

jlevon:
This is so we don't require Python when actually being used.

As we test in our CI, I don't think this is a big deal or worth fixing.

tmakatos:
I implemented a unit test and as I didn't have pytest installed it didn't do anything, wouldn't a warning be a reasonable compromise?

jlevon:
If you like, though I feel it'd be pretty lost in the noise of make test, and you'd find out when pushing to the PR branch anyway.

PR #556, opened 2021-06-08
Make pytest succeed even if python-3-pytest is not installed

awarus:
Fix for issue #542. Fixing issue for Makefile by means of 'which pytest-3'
When python-3-pytest isn't installed the following message will be displayed:

make pytest
[  3%] Built target pci_caps
[  6%] Built target irq
[  9%] Built target libvfio-user
[ 12%] Built target migration
[ 16%] Built target tran_sock
[ 19%] Built target pci
[ 22%] Built target dma
[ 25%] Built target vfio-user-static
[ 29%] Built target vfio-user-shared
[ 35%] Built target lspci
[ 41%] Built target gpio-pci-idio-16
[ 48%] Built target null
[ 54%] Built target server
[ 67%] Built target client
[100%] Built target unit-tests
=== Running python tests ===
No pytest-3 found in system. Try 'apt-get install python3-pytest' or try 'which pytest-3'
0 tests passed


Makefile
```diff
@@ -105,6 +105,10 @@ else
 
 pytest: all
 	@echo "=== Running python tests ==="
+ifeq ($(shell which pytest-3),)
```
jlevon, line 108: Thanks, but we don't want to depend on having which installed.

Instead, you can check this by looking at PYTESTCMD: if it's /bin/true, then we didn't find pytest-3 and can report the message

Makefile
```diff
@@ -105,6 +105,10 @@ else
...
+	@echo "No pytest-3 found in system. Try 'apt-get install python3-pytest' or try 'which pytest-3'"
+	@echo "0 tests passed"
```
jlevon, line 110: can we make this say "0 Python tests run" instead?

jlevon (changes requested):

Makefile
awarus: Yes, sure

jlevon (approved):

PR #559, opened 2021-06-09
spec: require DMA unmap to match exactly previous mapping

tmakatos:

### DIFF
--- a/docs/vfio-user.rst
+++ b/docs/vfio-user.rst
@@ -629,7 +629,8 @@ The request payload for this message is a structure of the following format:
 * *address* is the base DMA address of the DMA region.
 * *size* is the size of the DMA region.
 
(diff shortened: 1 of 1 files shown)


jlevon (approved):

docs/vfio-user.rst
```diff
@@ -629,7 +629,8 @@ The request payload for this message is a structure of the following format:
 * *address* is the base DMA address of the DMA region.
 * *size* is the size of the DMA region.
 
-The size of request message depends on whether or not the
+The address and size of the DMA region being unmapped must match exactly a
```
swapnili, line 632: line 661 says "Removing a portion of a DMA region is possible."
Does that means client can unmap portion of previously mapped region? how is that possible if exact addr and size is expected

Issue #341, opened 2021-02-12
make CFLAGS="-DNDEBUG" fails


tmakatos:
```
make CFLAGS="-DNDEBUG"
...
[ 53%] Building C object samples/CMakeFiles/server.dir/server.c.o
/root/src/libvfio-user/samples/server.c: In function ‘migration_write_data’:
/root/src/libvfio-user/samples/server.c:361:9: error: variable ‘ret’ set but not used [-Werror=unused-but-set-variable]
     int ret;
```

jlevon:
Not a bug, we intentionally make assert() work in release mode too

tmakatos:
> Not a bug, we intentionally make assert() work in release mode too

I know, my understanding is that `CFLAGS` can be overridden and `-DNDEBUG` can be specified, breaking the build.

jlevon:
Sure, but we don't need to support arbitrary cflags passed in. We decided to always have assert() on, so this is not a bug.

tmakatos:
The problem is that the compilation fails, so we should either redefine `NDEBUG` or at least document that `-DNDEBUG` can't be specified in `CFLAGS`.

jlevon:
Is there something that's expecting to do this?

In general we surely can't be expected to handle arbitrary compile options.

tmakatos:
I don't know if there's currently a valid use case, I accidentally run into when SPDK builts libvfio-user as a submodule.

I think it's quite common for some compile options to be set, and `DNDEBUG` is one of them. If we rely on `DNEBUG`, then at least there should be a sanity check where we fail to compile if it's not set.

jlevon:
fixed by #563

PR #563, opened 2021-06-09
Fix make CFLAGS="-DNDEBUG" failure

awarus:
Redefining NDEBUG in files where assert.h is included to prevent build fail due to -DNDEBUG flag
Fix for #341 issue
Signed-off-by: Ivan Efremov <i350300800e@gmail.com>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -147,6 +147,9 @@ realclean:
 	rm -rf $(GCOVS)
 
 $(BUILD_DIR)/Makefile:
+ifneq (,$(filter -DNDEBUG,$(CFLAGS)))
+	$(error -DNDEBUG flag is not supported)
+endif
 	mkdir -p $(BUILD_DIR)
 	cd $(BUILD_DIR); $(CMAKE) \
 		-D "CMAKE_C_COMPILER:STRING=$(CC)" \


jlevon:
Thanks for this, but I don't think this makes sense as a change we want, thoughts @tmakatos?

awarus:
> Thanks for this, but I don't think this makes sense as a change we want, thoughts @tmakatos?	

I saw the opened issue and comment of @tmakatos. If it's not needed, then we should close the issue or specify in docs this limitation

tmakatos:
> Thanks for this, but I don't think this makes sense as a change we want, thoughts @tmakatos?

While we can't expect the user to use any kind of complication/linking flags they want, I `DNDEBUG` is a very common one so I think it's worth fixing.

jlevon:
The problem is that NDEBUG has a very specific meaning: disable assertions. But we don't - and shouldn't - support doing that. We always build with assertions on. So it doesn't make sense to support it.

And this version - which basically just ignores the requested CFLAGS - doesn't seem right either.

awarus:
> The problem is that NDEBUG has a very specific meaning: disable assertions. But we don't - and shouldn't - support doing that. We always build with assertions on. So it doesn't make sense to support it.
> 
> And this version - which basically just ignores the requested CFLAGS - doesn't seem right either.

Do you mean something like just a sanity check of CFLAGS in Makefile before building the project and some message that DNDEGUB force disabled?

jlevon:
Sure, we could do something like that if you like.

Makefile
```diff
@@ -147,6 +147,9 @@ realclean:
 	rm -rf $(GCOVS)
 
 $(BUILD_DIR)/Makefile:
+ifneq (,$(filter -DNDEBUG,$(CFLAGS)))
+	$(error -DNDEBUG flag is not support)
```
jlevon, line 151: nit, this should say "supported"

jlevon (changes requested):

Makefile
awarus: Yeah, sure

jlevon (approved):

PR #564, opened 2021-06-10
dma: Fix returned sg array

swapnili:
_dma_addr_sg_split() is supposed to return back sg's if the requested
dma area spans across dma regions.
Also adding unit tests to cover these cases.


tmakatos (approved):
Good spot!

PR #569, opened 2021-06-18
split writes to PCI header

tmakatos:
Multiple registers can be written in one go so we need to split the
write.


lib/pci.c
```diff
@@ -221,47 +221,55 @@ pci_hdr_write(vfu_ctx_t *vfu_ctx, const char *buf, size_t count, loff_t offset)
...
-    default:
-        vfu_log(vfu_ctx, LOG_INFO, "PCI config write %#lx-%#lx not handled",
-                offset, offset + count);
-        ret = ERROR_INT(EINVAL);
+    while (count > 0) {
+        _count = count;
+        switch (_offset) {
+        case PCI_COMMAND:
+            _count = MIN(sizeof(vfu_pci_hdr_cmd_t), count);
+            ret = handle_command_write(vfu_ctx, cfg_space, buf, _count);
```
jlevon, line 237: I don't think we want to over-write a previous failure do we ? same for :264

lib/pci.c
```diff
@@ -221,47 +221,55 @@ pci_hdr_write(vfu_ctx_t *vfu_ctx, const char *buf, size_t count, loff_t offset)
...
-        pci_hdr_write_bar(vfu_ctx, BAR_INDEX(offset), buf);
-        break;
-    case PCI_ROM_ADDRESS:
-        ret = handle_erom_write(vfu_ctx, cfg_space, buf, count);
-        break;
-    default:
-        vfu_log(vfu_ctx, LOG_INFO, "PCI config write %#lx-%#lx not handled",
-                offset, offset + count);
-        ret = ERROR_INT(EINVAL);
+    while (count > 0) {
```
jlevon, line 232: If we're trying to handle more than one register at once, I think we need to do so for all registers. Right now if I write 20 bytes starting at PCI_STATUS, the whole write is ignored, even though PCI_STATUS isn't 20 bytes long.

We also already have code that splits up writes - pci_config_space_next_segment(). Can you extend *that* instead - it can call into a utility version that looks at offset and returns the size of the register when inside the standard header. Then, you don't need to track offset/count yourself, the code in this func can stay as it is.

jlevon (changes requested):

lib/pci.c
tmakatos: Yeah doing that in pci_config_space_next_segment() is a much better idea.

lib/pci.c
```diff
@@ -330,15 +321,50 @@ pci_nonstd_access(vfu_ctx_t *vfu_ctx, char *buf, size_t count,
...
  * @cb is set to the callback to use for accessing the segment.
  */
-static size_t
+static ssize_t
 pci_config_space_next_segment(vfu_ctx_t *ctx, size_t count, loff_t offset,
-                              vfu_region_access_cb_t **cb)
+                              bool is_write, vfu_region_access_cb_t **cb)
 {
     struct pci_cap *cap;
+    static const size_t off2sz[] = {
```
jlevon, line 329: could you move this out to a utility function please?
Also, instead of hard-coding, you can do something like:

     vfu_pci_hdr_t hdr;                                                           
                                                                                 
     static const size_t off2sz[] = {                                             
       [offsetof(vfu_pci_hdr_t, bist)] = sizeof(hdr.bist)  
...                      
     };                                                                           

although you'd still need to account for the BAR array a bit more manually

lib/pci.c
```diff
@@ -379,9 +405,15 @@ pci_config_space_access(vfu_ctx_t *vfu_ctx, char *buf, size_t count,
 
     while (count > 0) {
         vfu_region_access_cb_t *cb;
-        size_t size;
-
-        size = pci_config_space_next_segment(vfu_ctx, count, offset, &cb);
+        ssize_t size;
+
+        size = pci_config_space_next_segment(vfu_ctx, count, offset, is_write,
```
jlevon, line 410: could we keep it as size_t and just not set "cb" for the error condition maybe? what do you think?

jlevon (changes requested):

lib/pci.c
tmakatos: > could you move this out to a utility function please?

Sure.

> Also, instead of hard-coding, you can do something like:

OK.

lib/pci.c
tmakatos: sure

jlevon (approved):
thanks

PR #578, opened 2021-07-09
add VFU_REGION_FLAG_ALWAYS_CB to receive callback always

jraman567:
Adds VFU_REGION_FLAG_ALWAYS_CB flag which always invokes a
callback for region accesses

### DIFF
--- a/include/libvfio-user.h
+++ b/include/libvfio-user.h
@@ -221,6 +221,7 @@ typedef ssize_t (vfu_region_access_cb_t)(vfu_ctx_t *vfu_ctx, char *buf,
 #define VFU_REGION_FLAG_WRITE   (1 << 1)
 #define VFU_REGION_FLAG_RW      (VFU_REGION_FLAG_READ | VFU_REGION_FLAG_WRITE)
 #define VFU_REGION_FLAG_MEM     (1 << 2)    // if unset, bar is IO
+#define VFU_REGION_FLAG_ALWAYS_CB   (1 << 3)
 
 /**
  * Set up a device region.
--- a/lib/libvfio-user.c
+++ b/lib/libvfio-user.c
@@ -208,7 +208,8 @@ region_access(vfu_ctx_t *vfu_ctx, size_t region_index, char *buf,
         dump_buffer("buffer write", buf, count);
     }
 
-    if (region_index == VFU_PCI_DEV_CFG_REGION_IDX) {
+    if ((region_index == VFU_PCI_DEV_CFG_REGION_IDX) &&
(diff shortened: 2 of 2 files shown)


include/libvfio-user.h
```diff
@@ -221,6 +221,7 @@ typedef ssize_t (vfu_region_access_cb_t)(vfu_ctx_t *vfu_ctx, char *buf,
 #define VFU_REGION_FLAG_WRITE   (1 << 1)
 #define VFU_REGION_FLAG_RW      (VFU_REGION_FLAG_READ | VFU_REGION_FLAG_WRITE)
 #define VFU_REGION_FLAG_MEM     (1 << 2)    // if unset, bar is IO
+#define VFU_REGION_FLAG_ALWAYS_CB   (1 << 3)
```
jlevon, line 224: please document this on line :266

lib/libvfio-user.c
```diff
@@ -1445,7 +1446,7 @@ vfu_setup_region(vfu_ctx_t *vfu_ctx, int region_idx, size_t size,
      * PCI config space is never mappable or of type mem.
      */
     if (region_idx == VFU_PCI_DEV_CFG_REGION_IDX &&
-        flags != VFU_REGION_FLAG_RW) {
+        !(flags & VFU_REGION_FLAG_RW)) {
```
jlevon, line 1449: this isn't right, it should be checking that VFU_REGION_FLAG_MEM isn't set too

include/libvfio-user.h
jlevon: also, I think you need to check in the setup region call: if this flag is set, then a callback function must be provided.

jlevon (changes requested):
Do you mind adding tests for this in test/py/test_setup_region.py ?
You can see how to add a region CB in test_pci_ext_caps.py:

```
 67     ret = vfu_setup_region(ctx, index=VFU_PCI_DEV_CFG_REGION_IDX,                
 68                            size=PCI_CFG_SPACE_EXP_SIZE, cb=pci_region_cb,        
 69                            flags=VFU_REGION_FLAG_RW)                             
```

```
 94 @vfu_region_access_cb_t                                                          
 95 def pci_region_cb(ctx, buf, count, offset, is_write):                            
 96     if not is_write:                                                             
 97         return read_pci_cfg_space(ctx, buf, count, offset, extended=True)        
 98                                                                                  
 99     return write_pci_cfg_space(ctx, buf, count, offset, extended=True)           
```

So we'd just want a test that confirms that a PCI config space access we usually handle in the library, instead goes to the CB.

Thanks!

jlevon:
merged separately

PR #579, opened 2021-07-09
check for valid vfu_setup_region() flags

jlevon:


jlevon:
I'll wait on Jag's PR

lib/libvfio-user.c
```diff
@@ -1430,6 +1430,17 @@ vfu_setup_region(vfu_ctx_t *vfu_ctx, int region_idx, size_t size,
 
     assert(vfu_ctx != NULL);
 
+    if ((flags & ~(VFU_REGION_FLAG_MASK)) ||
+        (!(flags & VFU_REGION_FLAG_RW))) {
```
tmakatos, line 1434: Probably not important, but why not allow a region is not readable/writeable, is not an I/O region, and there are no callabacks?

tmakatos (approved):

lib/libvfio-user.c
jlevon: What would the point in such a region be? nothing could ever interact with it could it?

tmakatos: agreed

PR #580, opened 2021-07-13
don't call DMA region add callabck when duplicate region is added

tmakatos:

### DIFF
--- a/lib/dma.c
+++ b/lib/dma.c
@@ -299,7 +299,8 @@ dirty_page_logging_start_on_region(dma_memory_region_t *region, size_t pgsize)
 int
 MOCK_DEFINE(dma_controller_add_region)(dma_controller_t *dma,
                                        vfu_dma_addr_t dma_addr, size_t size,
-                                       int fd, off_t offset, uint32_t prot)
+                                       int fd, off_t offset, uint32_t prot,
+                                       bool *exist)
 {
     dma_memory_region_t *region;
     int page_size = 0;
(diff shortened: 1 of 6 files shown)


jlevon:
I think I'm missing a bit of context here. Is this working around a qemu bug or something? 
Do we need something in the spec to say we are required to ignore duplicate exact mappings?

jlevon:
Small typo in the commit title too

tmakatos:
> Small typo in the commit title too

ok

tmakatos:
> I think I'm missing a bit of context here. Is this working around a qemu bug or something?
> Do we need something in the spec to say we are required to ignore duplicate exact mappings?

QEMU sends two DMA map requests with the same DMA region. In our DMA code we don't re-add this region so I assumed that this behavior is legitimate. But I suppose your question makes sense, we need establish whether this behavior should be allowed by the spec.

tmakatos:
> > I think I'm missing a bit of context here. Is this working around a qemu bug or something?
> > Do we need something in the spec to say we are required to ignore duplicate exact mappings?
> 
> QEMU sends two DMA map requests with the same DMA region. In our DMA code we don't re-add this region so I assumed that this behavior is legitimate. But I suppose your question makes sense, we need to establish whether this behavior should be allowed by the spec in the first place.

@jlevon we already cover this in the spec:
> A command to map over an existing region must be failed by the server with EEXIST set in error field in the reply.

So we need to tell JJ to fix it in QEMU. I'll implement a unit test for this.

tmakatos:
Haven't hit this in a very long time, closing.

PR #586, opened 2021-08-16
fix dma_{map,unmap}_sg() array handling

jlevon:


lib/dma.h
```diff
@@ -245,15 +245,15 @@ dma_map_sg(dma_controller_t *dma, dma_sg_t *sg, struct iovec *iov,
             return ERROR_INT(EFAULT);
         }
 
-        if (sg->writeable) {
+        if (sg[i].writeable) {
             if (dma->dirty_pgsize > 0) {
                 _dma_mark_dirty(dma, region, sg);
```
swapnili, line 250: &sg[i] here?

lib/dma.h
```diff
@@ -282,7 +282,7 @@ dma_unmap_sg(dma_controller_t *dma, const dma_sg_t *sg,
             /* bad region */
             continue;
         }
-        if (sg->writeable) {
+        if (sg[i].writeable) {
             LIST_REMOVE(sg, entry);
```
swapnili, line 286: Also &sg[i] here

sboeuf:
I'm trying to make the vIOMMU case work, which means multiple DMA mapping are performed (map/unmap), therefore I had a try with this patch. Unfortunately this is not enough to make the handling of multiple region working. I run into a very specific error where the NVME queue 1 can't be allocated properly because its size is 16KiB while the mapping the client provided to the server are 4KiB, meaning the code in `dma.c` find that we need 4 regions to cover the whole thing.
Problem is, there's a strong assumption in `spdk/lib/nvmf/vfio_user.c` `map_one()` function that makes the handling of multiple regions impossible.
```c
ret = vfu_addr_to_sg(ctx, (void *)(uintptr_t)addr, len, sg, 1, prot);
```
should not limit the scatter/gather maximum to 1, and later on:
```c
ret = vfu_map_sg(ctx, sg, iov, 1, 0);
```
should rely on the number of regions returned by the previous function to map correctly, instead of assuming 1 again.

/cc @jraman567

swapnili (changes requested):

jlevon:
@sboeuf please file a bug at https://github.com/spdk/spdk/issues/ on that. The code looks like it's presuming large page mappings, but you're running with 4K page size.

swapnili (approved):

sboeuf:
@jlevon 
>please file a bug at https://github.com/spdk/spdk/issues/ on that. The code looks like it's presuming large page mappings, but you're running with 4K page size.

Sounds good I'll do that.

PR #591, opened 2021-08-18
dma: cleanup dma_{map,unmap}_sg

swapnili:
Instead of using index to traverse sg and iovec, better to use it as
pointers. Its more readable and less prone from coding mistakes.
Also adding unit tests for the same.

### DIFF
--- a/test/unit-tests.c
+++ b/test/unit-tests.c
@@ -361,7 +361,7 @@ static void
 test_dma_map_sg(void **state UNUSED)
 {
     dma_sg_t sg[] = { {.region = 0}, {.region = 1}, {.region = 2} };
-    struct iovec iovec[] = { {0} , {0}, {0} };
+    struct iovec iovec[] = { {0}, {0}, {0} };
 
     /* bad region */
     assert_int_equal(-1, dma_map_sg(vfu_ctx.dma, &sg[0], &iovec[0], 1));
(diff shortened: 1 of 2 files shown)


jlevon (commented):
Code is much better like this, thanks! But could we do the test side in test/py/ instead? I much prefer a component test in python to making the C unit tests longer.

Would just need to extend the python SG handling code to handle > 1 SG

swapnili:
Actually there are multiple {map,unmap}_sg test cases in C unit tests which are not in test/py. So I think it's better to have a separate PR handling them along with these.

jlevon:
Sure, if you like.

test/unit-tests.c
```diff
@@ -360,28 +360,67 @@ test_dma_controller_remove_region_unmapped(void **state UNUSED)
 static void
 test_dma_map_sg(void **state UNUSED)
 {
-    dma_sg_t sg = { .region = 1 };
-    struct iovec iovec = { 0 };
+    dma_sg_t sg[] = { {.region = 0} , {.region = 1}, {.region = 2}};
```
jlevon, line 363: nit, here, below, consistent spaces around { and } ?

test/unit-tests.c
```diff
@@ -360,28 +360,67 @@ test_dma_controller_remove_region_unmapped(void **state UNUSED)
 static void
 test_dma_map_sg(void **state UNUSED)
 {
-    dma_sg_t sg = { .region = 1 };
-    struct iovec iovec = { 0 };
+    dma_sg_t sg[] = { {.region = 0}, {.region = 1}, {.region = 2} };
+    struct iovec iovec[] = { {0} , {0}, {0} };
```
jlevon, line 364: weird extra space here still and below?

swapnili: Oops sorry somehow it got lost in stash clear

jlevon (approved):
thanks

PR #597, opened 2021-08-24
Fix err path of handle_dma_unmap()

swapnili:
Set msg->out_size before successful return. Otherwise in case of error
reply path we may endup setting iovecs[1].iov_len with invalid
iovecs[1].iov_base in tran_sock_reply()


jlevon (commented):
Thanks for the fix! Change looks good, but we can test this path too.
Can you pull in:

https://github.com/nutanix/libvfio-user/pull/599/files

the tests should pass with your fix in place (I hope)

swapnili:
Sorry I was not clear here. I already have the tests cases for this which I wanted to open for review with the UNMAP_ALL code changes.

jlevon (approved):
thanks!

swapnili:
I have pulled in the test case, made some fixes in it. Later when this is merged I will rebase my UNMAP_ALL changes on it.
Thanks

PR #601, opened 2021-08-27
libvfio-user phase one

JAKelly10:
Currently here to look at the diff more easily

### DIFF
--- a/include/libvfio-user.h
+++ b/include/libvfio-user.h
@@ -89,52 +89,15 @@ typedef enum {
     VFU_DEV_TYPE_PCI
 } vfu_dev_type_t;
 
-typedef struct request {
-    uint32_t argsz;
-    uint32_t flags;
-    uint32_t index;
-    uint32_t count;
-} __attribute__((packed)) request_t;
+typedef struct region_io_fds region_io_fds_request_t;
 
-typedef struct sub_region_ioeventfd {
-    uint64_t offset;
-    uint64_t size;
-    uint32_t fd_index;
-    uint32_t type;
-    uint32_t flags;
-    uint32_t padding;
-    uint64_t datamatch;
-} __attribute__((packed)) sub_region_ioeventfd_t;
+typedef struct sub_region_ioregionfd sub_region_ioregionfd_t;
 
-typedef struct sub_region_ioregionfd {
-    uint64_t offset;
-    uint64_t size;
-    uint32_t fd_index;
-    uint32_t type;
-    uint32_t flags;
-    uint32_t padding;
-    uint64_t user_data;
-} __attribute__((packed)) sub_region_ioregionfd_t;
+typedef struct sub_region_ioeventfd sub_region_ioeventfd_t;
 
-typedef union sub_region {
-    sub_region_ioeventfd_t ioeventfd;
-    sub_region_ioregionfd_t toregionfd;
-} __attribute__((packed)) sub_region_t;
+typedef struct reply_sub_region reply_sub_region_t;
 
-typedef struct ioeventfd {
-    uint64_t offset;
-    uint64_t size;
-    int32_t fd;
-    uint32_t flags;
-} __attribute__((packed)) ioeventfd_t;
-
-typedef struct reply {
-    uint32_t argsz;
-    uint32_t flags;
-    uint32_t index;
-    uint32_t count;
-    sub_region_t sub_regions[];
-} __attribute__((packed)) reply_t;
+typedef struct ioeventfd ioeventfd_t;
 
 /**
  * Creates libvfio-user context. By default one ERR and one REQ IRQs are
--- a/include/libvfio-user.h
+++ b/include/libvfio-user.h
@@ -89,6 +89,52 @@ typedef enum {
     VFU_DEV_TYPE_PCI
 } vfu_dev_type_t;
 
+typedef struct request {
+    uint32_t argsz;
+    uint32_t flags;
+    uint32_t index;
+    uint32_t count;
+} request_t;
+
+typedef struct sub_region_ioeventfd {
+    uint64_t offset;
+    uint64_t size;
+    uint32_t fd_index;
+    uint32_t type;
+    uint32_t flags;
+    uint32_t padding;
+    uint64_t datamatch;
+} sub_region_ioeventfd_t;
+
+typedef struct sub_region_ioregionfd {
+    uint64_t offset;
+    uint64_t size;
+    uint32_t fd_index;
+    uint32_t type;
+    uint32_t flags;
+    uint32_t padding;
+    uint64_t user_data;
+} sub_region_ioregionfd_t;
+
+typedef union sub_region {
+    sub_region_ioeventfd_t ioeventfd;
+    sub_region_ioregionfd_t toregionfd;
+} sub_region_t;
+
+typedef struct ioeventfd {
+    uint64_t offset;
+    uint64_t size;
+    int32_t fd;
+} ioeventfd_t;
+
+typedef struct reply {
+    uint32_t argsz;
+    uint32_t flags;
+    uint32_t index;
+    uint32_t count;
+    sub_region_t sub_regions[];
+} reply_t;
+
 /**
  * Creates libvfio-user context. By default one ERR and one REQ IRQs are
  * initialized, this can be overridden with vfu_setup_device_nr_irqs.
@@ -883,6 +929,19 @@ vfu_pci_find_next_capability(vfu_ctx_t *vfu_ctx, bool extended,
 
 bool
 vfu_sg_is_mappable(vfu_ctx_t *vfu_ctx, dma_sg_t *sg);
+
+/*
+ *
+ * @vfu_ctx: the libvfio-user context
+ * @offset: The offset into the memory region
+ * @flags: Any flags to set up the ioeventfd
+ * @index: The index of the memory region to set up the ioeventfd
+ */
+int
+create_ioeventfd(vfu_ctx_t *vfu_ctx, size_t offset, uint32_t flags,
+                 uint32_t index);
+
+int delete_ioeventfd(vfu_ctx_t *vfu_ctx, uint32_t index, uint32_t fd_index);
 
 #ifdef __cplusplus
 }
--- a/include/libvfio-user.h
+++ b/include/libvfio-user.h
@@ -894,31 +894,14 @@ vfu_sg_is_mappable(vfu_ctx_t *vfu_ctx, dma_sg_t *sg);
  * @region_idx: The index of the memory region to set up the ioeventfd
  * @fd: the value of the file descriptor
  * @offset: The offset into the memory region
- * @size: size of the memory region
+ * @size: size of the ioeventfd
  * @flags: Any flags to set up the ioeventfd
- * @datamtach: sets the datamatch value
+ * @datamatch: sets the datamatch value
  */
 int
 vfu_create_ioeventfd(vfu_ctx_t *vfu_ctx, uint32_t region_idx, int fd,
                      size_t offset, uint32_t size, uint32_t flags,
                      uint64_t datamatch);
-
-/*
- * Deletes an ioeventfd at the given setup memory region with @offset, @size,
- * @fd_index, @flags and @datamatch.
- *
- * Returns 0 on success and -1 on failure.
- *
- * @vfu_ctx: the libvfio-user context
- * @region_idx: The index of the memory region to set up the ioeventfd
- * @offset: offset within the memory region
- * @size: the size of the region
- * @fd_index: the index of the fd from the subregion
- * @flags: the flags of the region
- */
-int
-vfu_delete_ioeventfd(vfu_ctx_t *vfu_ctx, uint32_t region_idx, size_t offset,
-                     uint32_t size, uint32_t fd_index, uint32_t flags);
 
 #ifdef __cplusplus
 }
--- a/include/libvfio-user.h
+++ b/include/libvfio-user.h
@@ -99,6 +99,8 @@ typedef struct reply_sub_region reply_sub_region_t;
 
 typedef struct ioeventfd ioeventfd_t;
 
+typedef struct ioeventfd_list ioeventfd_list_t;
+
 /**
  * Creates libvfio-user context. By default one ERR and one REQ IRQs are
  * initialized, this can be overridden with vfu_setup_device_nr_irqs.
--- a/include/libvfio-user.h
+++ b/include/libvfio-user.h
@@ -89,18 +89,6 @@ typedef enum {
     VFU_DEV_TYPE_PCI
 } vfu_dev_type_t;
 
-typedef struct region_io_fds region_io_fds_request_t;
-
-typedef struct sub_region_ioregionfd sub_region_ioregionfd_t;
-
-typedef struct sub_region_ioeventfd sub_region_ioeventfd_t;
-
-typedef struct reply_sub_region reply_sub_region_t;
-
-typedef struct ioeventfd ioeventfd_t;
-
-typedef struct ioeventfd_list ioeventfd_list_t;
-
 /**
  * Creates libvfio-user context. By default one ERR and one REQ IRQs are
  * initialized, this can be overridden with vfu_setup_device_nr_irqs.
@@ -897,19 +885,40 @@ bool
 vfu_sg_is_mappable(vfu_ctx_t *vfu_ctx, dma_sg_t *sg);
 
 /*
+ * Creates a new ioeventfd at the given setup memory region with @offset, @size,
+ * @fd, @flags and @datamatch.
+ *
+ * Returns 0 on success and -1 on failure with errno set.
  *
  * @vfu_ctx: the libvfio-user context
+ * @region_idx: The index of the memory region to set up the ioeventfd
+ * @fd: the value of the file descriptor
  * @offset: The offset into the memory region
+ * @size: size of the memory region
  * @flags: Any flags to set up the ioeventfd
- * @index: The index of the memory region to set up the ioeventfd
+ * @datamtach: sets the datamatch value
  */
 int
-create_ioeventfd(vfu_ctx_t *vfu_ctx, size_t offset, uint32_t flags,
-                 uint32_t index);
+vfu_create_ioeventfd(vfu_ctx_t *vfu_ctx, uint32_t region_idx, int fd,
+                     size_t offset, uint32_t size, uint32_t flags,
+                     uint64_t datamatch);
 
-int vfu_create_ioeventfd(vfu_ctx_t *vfu_ctx, size_t offset, uint32_t size, int fd, uint32_t flags, uint32_t index);
-
-int vfu_delete_ioeventfd(vfu_ctx_t *vfu_ctx, uint32_t index, uint32_t fd_index);
+/*
+ * Deletes an ioeventfd at the given setup memory region with @offset, @size,
+ * @fd_index, @flags and @datamatch.
+ *
+ * Returns 0 on success and -1 on failure.
+ *
+ * @vfu_ctx: the libvfio-user context
+ * @region_idx: The index of the memory region to set up the ioeventfd
+ * @offset: offset within the memory region
+ * @size: the size of the region
+ * @fd_index: the index of the fd from the subregion
+ * @flags: the flags of the region
+ */
+int
+vfu_delete_ioeventfd(vfu_ctx_t *vfu_ctx, uint32_t region_idx, size_t offset,
+                     uint32_t size, uint32_t fd_index, uint32_t flags);
 
 #ifdef __cplusplus
 }
--- a/include/libvfio-user.h
+++ b/include/libvfio-user.h
@@ -94,7 +94,7 @@ typedef struct request {
     uint32_t flags;
     uint32_t index;
     uint32_t count;
-} request_t;
+} __attribute__((packed)) request_t;
 
 typedef struct sub_region_ioeventfd {
     uint64_t offset;
@@ -104,7 +104,7 @@ typedef struct sub_region_ioeventfd {
     uint32_t flags;
     uint32_t padding;
     uint64_t datamatch;
-} sub_region_ioeventfd_t;
+} __attribute__((packed)) sub_region_ioeventfd_t;
 
 typedef struct sub_region_ioregionfd {
     uint64_t offset;
@@ -114,18 +114,19 @@ typedef struct sub_region_ioregionfd {
     uint32_t flags;
     uint32_t padding;
     uint64_t user_data;
-} sub_region_ioregionfd_t;
+} __attribute__((packed)) sub_region_ioregionfd_t;
 
 typedef union sub_region {
     sub_region_ioeventfd_t ioeventfd;
     sub_region_ioregionfd_t toregionfd;
-} sub_region_t;
+} __attribute__((packed)) sub_region_t;
 
 typedef struct ioeventfd {
     uint64_t offset;
     uint64_t size;
     int32_t fd;
-} ioeventfd_t;
+    uint32_t flags;
+} __attribute__((packed)) ioeventfd_t;
 
 typedef struct reply {
     uint32_t argsz;
@@ -133,7 +134,7 @@ typedef struct reply {
     uint32_t index;
     uint32_t count;
     sub_region_t sub_regions[];
-} reply_t;
+} __attribute__((packed)) reply_t;
 
 /**
  * Creates libvfio-user context. By default one ERR and one REQ IRQs are
@@ -941,7 +942,9 @@ int
 create_ioeventfd(vfu_ctx_t *vfu_ctx, size_t offset, uint32_t flags,
                  uint32_t index);
 
-int delete_ioeventfd(vfu_ctx_t *vfu_ctx, uint32_t index, uint32_t fd_index);
+int vfu_create_ioeventfd(vfu_ctx_t *vfu_ctx, size_t offset, uint32_t size, int fd, uint32_t flags, uint32_t index);
+
+int vfu_delete_ioeventfd(vfu_ctx_t *vfu_ctx, uint32_t index, uint32_t fd_index);
 
 #ifdef __cplusplus
 }
--- a/include/vfio-user.h
+++ b/include/vfio-user.h
@@ -156,6 +156,45 @@ struct vfio_user_irq_info {
     uint32_t    subindex;
 } __attribute__((packed));
 
+struct region_io_fds {
+    uint32_t argsz;
+    uint32_t flags;
+    uint32_t index;
+    uint32_t count;
+} __attribute__((packed));
+
+struct sub_region_ioeventfd {
+    uint64_t offset;
+    uint64_t size;
+    uint32_t fd_index;
+    uint32_t type;
+    uint32_t flags;
+    uint32_t padding;
+    uint64_t datamatch;
+} __attribute__((packed));
+
+struct sub_region_ioregionfd {
+    uint64_t offset;
+    uint64_t size;
+    uint32_t fd_index;
+    uint32_t type;
+    uint32_t flags;
+    uint32_t padding;
+    uint64_t user_data;
+} __attribute__((packed));
+
+struct reply_sub_region {
+    uint32_t argsz;
+    uint32_t flags;
+    uint32_t index;
+    uint32_t count;
+    union sub_region {
+        struct sub_region_ioeventfd ioeventfd;
+        struct sub_region_ioregionfd toregionfd;
+    } sub_regions[];
+} __attribute__((packed));
+
+
 /* Analogous to vfio_iommu_type1_dirty_bitmap. */
 struct vfio_user_dirty_pages {
     uint32_t    argsz;
--- a/include/vfio-user.h
+++ b/include/vfio-user.h
@@ -156,29 +156,17 @@ struct vfio_user_irq_info {
     uint32_t    subindex;
 } __attribute__((packed));
 
-typedef struct vfio_user_region_io_fds_request vfio_user_region_io_fds_request_t;
-
-typedef struct vfio_user_sub_region_ioregionfd vfio_user_sub_region_ioregionfd_t;
-
-typedef struct vfio_user_sub_region_ioeventfd vfio_user_sub_region_ioeventfd_t;
-
-typedef struct vfio_user_region_io_fds_reply vfio_user_region_io_fds_reply_t;
-
-typedef struct ioeventfd ioeventfd_t;
-
-typedef struct ioeventfd_list ioeventfd_list_t;
-
-struct vfio_user_region_io_fds_request {
+typedef struct vfio_user_region_io_fds_request {
     uint32_t argsz;
     uint32_t flags;
     uint32_t index;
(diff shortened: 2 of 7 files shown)


test/py/test_device_get_region_io_fds.py
```diff
@@ -0,0 +1,181 @@
...
+
+    reply, ret = region_io_fds_reply.pop_from_buffer(ret)
+    assert reply.argsz == 16
+
+def test_device_get_region_io_fds_fds_read_write_dupe_fd():
+
+    t = eventfd(0,0)
+    assert vfu_create_ioeventfd(ctx, 6 * 8, 8, t, 0, VFU_PCI_DEV_BAR2_REGION_IDX) != -1
+    assert vfu_create_ioeventfd(ctx, 7 * 8, 8, t, 0, VFU_PCI_DEV_BAR2_REGION_IDX) != -1
+
```
jlevon, line 132: spurious blank line

test/py/test_device_get_region_io_fds.py
```diff
@@ -0,0 +1,181 @@
...
+        ioevent, ret = sub_region_ioeventfd.pop_from_buffer(ret)
+        ioevents.append(ioevent)
+        os.write(newfds[ioevent.fd_index], c.c_ulonglong(1))
+
+    print(newfds, [i.fd_index for i in ioevents])
+
+    for i in range(0, reply.count):
+        out = os.read(newfds[ioevents[i].fd_index], ioevent.size)
+        [out] = struct.unpack("@Q",out)
+        assert out  == 1
```
jlevon, line 114: spurious space

test/py/test_device_get_region_io_fds.py
```diff
@@ -0,0 +1,181 @@
...
+        assert ioevents[i].size == 8
+        assert ioevents[i].offset == 56 - (8 * i)
+
+
+    assert ioevents[0].fd_index == ioevents[1].fd_index
+    assert ioevents[0].offset != ioevents[1].offset
+
+    os.write(newfds[ioevents[0].fd_index], c.c_ulonglong(1))
```
jlevon, line 163: I think you want to write to the ioevent[1] fd_index too, then confirm that it reads on the server side as 2 ?

jlevon (changes requested):

lib/libvfio-user.c
```diff
@@ -1469,6 +1657,7 @@ vfu_setup_region(vfu_ctx_t *vfu_ctx, int region_idx, size_t size,
     reg->cb = cb;
     reg->fd = fd;
     reg->offset = offset;
+    LIST_INIT(&reg->sub_reg_l);
```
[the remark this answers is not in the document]
JAKelly10: No it won't I just checked and since LIST_INIT just ensures that the value is null and the code for  REGION_IO_FDS makes a check to count how many fds it is possible to return for purposes of allocation. As the end of the list and in this case its uninitalised states are both NULL it just sets number of regions to 0 and returns no sub regions in the reply. Added a test for it and can confirm it returns a reply with no sub regions.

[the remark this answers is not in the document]
jlevon: You're right, it just inits to NULL anyway, thanks.

But, logically, it's the correct thing to do still. So whilst there's no actual bug here, it should still logically be moved over to where I suggested.

include/libvfio-user.h
```diff
@@ -884,6 +896,21 @@ vfu_pci_find_next_capability(vfu_ctx_t *vfu_ctx, bool extended,
...
+ * @flags: Any flags to set up the ioeventfd
+ * @index: The index of the memory region to set up the ioeventfd
+ */
+int
+create_ioeventfd(vfu_ctx_t *vfu_ctx, size_t offset, uint32_t flags,
+                 uint32_t index);
+
+int vfu_create_ioeventfd(vfu_ctx_t *vfu_ctx, size_t offset, uint32_t size, int fd, uint32_t flags, uint32_t index);
+
+int vfu_delete_ioeventfd(vfu_ctx_t *vfu_ctx, uint32_t index, uint32_t fd_index);
```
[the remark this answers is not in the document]
JAKelly10: I think we found that the kernel seemed to allow absolutely everything.

include/libvfio-user.h
```diff
@@ -897,19 +885,40 @@ bool
...
+ * Returns 0 on success and -1 on failure with errno set.
  *
  * @vfu_ctx: the libvfio-user context
+ * @region_idx: The index of the memory region to set up the ioeventfd
+ * @fd: the value of the file descriptor
  * @offset: The offset into the memory region
+ * @size: size of the memory region
  * @flags: Any flags to set up the ioeventfd
- * @index: The index of the memory region to set up the ioeventfd
+ * @datamtach: sets the datamatch value
```
jlevon, line 899: nit, "datamatch"

include/libvfio-user.h
```diff
@@ -897,19 +885,40 @@ bool
...
+ * Creates a new ioeventfd at the given setup memory region with @offset, @size,
+ * @fd, @flags and @datamatch.
+ *
+ * Returns 0 on success and -1 on failure with errno set.
  *
  * @vfu_ctx: the libvfio-user context
+ * @region_idx: The index of the memory region to set up the ioeventfd
+ * @fd: the value of the file descriptor
  * @offset: The offset into the memory region
+ * @size: size of the memory region
```
jlevon, line 897: this should be "size of the sub-region" or some such

include/libvfio-user.h
```diff
@@ -897,19 +885,40 @@ bool
...
+/*
+ * Deletes an ioeventfd at the given setup memory region with @offset, @size,
+ * @fd_index, @flags and @datamatch.
+ *
+ * Returns 0 on success and -1 on failure.
+ *
+ * @vfu_ctx: the libvfio-user context
+ * @region_idx: The index of the memory region to set up the ioeventfd
+ * @offset: offset within the memory region
+ * @size: the size of the region
```
jlevon, line 915: ditto

include/libvfio-user.h
```diff
@@ -897,19 +885,40 @@ bool
...
+ * @fd_index: the index of the fd from the subregion
+ * @flags: the flags of the region
```
jlevon, line 917: what about datamatch?

include/libvfio-user.h
```diff
@@ -897,19 +885,40 @@ bool
...
+ * Deletes an ioeventfd at the given setup memory region with @offset, @size,
+ * @fd_index, @flags and @datamatch.
+ *
+ * Returns 0 on success and -1 on failure.
+ *
+ * @vfu_ctx: the libvfio-user context
+ * @region_idx: The index of the memory region to set up the ioeventfd
+ * @offset: offset within the memory region
+ * @size: the size of the region
+ * @fd_index: the index of the fd from the subregion
```
jlevon, line 916: this should just be the fd, not the index - how could a client know the index?

lib/libvfio-user.c
```diff
@@ -462,186 +462,198 @@ handle_device_get_region_info(vfu_ctx_t *vfu_ctx, vfu_msg_t *msg)
...
+    vfu_reg_info_t *vfu_reg = &vfu_ctx->reg_info[region_idx];
 
     assert(vfu_ctx != NULL);
     assert(fd >= 0);
 
-    vfu_reg_info_t *vfu_reg = &vfu_ctx->reg_info[index];
+    vfu_reg = &vfu_ctx->reg_info[region_idx];
+
+    if (offset + size > vfu_reg->size) {
+        ERROR_INT(EINVAL);
```
jlevon, line 477: here, elsewhere (?), you can just do:

```
return ERROR_INT(EINVAL);
```

lib/libvfio-user.c
```diff
@@ -462,186 +462,198 @@ handle_device_get_region_info(vfu_ctx_t *vfu_ctx, vfu_msg_t *msg)
...
 EXPORT int
-vfu_create_ioeventfd(vfu_ctx_t *vfu_ctx, size_t offset, uint32_t size, int fd, uint32_t flags, uint32_t index)
+vfu_create_ioeventfd(vfu_ctx_t *vfu_ctx, uint32_t region_idx, int fd,
+                     size_t offset, uint32_t size, uint32_t flags,
+                     uint64_t datamatch)
 {
+    vfu_reg_info_t *vfu_reg = &vfu_ctx->reg_info[region_idx];
 
     assert(vfu_ctx != NULL);
     assert(fd >= 0);
```
jlevon, line 473: still need to check region_idx is valid

lib/libvfio-user.c
```diff
@@ -462,186 +462,198 @@ handle_device_get_region_info(vfu_ctx_t *vfu_ctx, vfu_msg_t *msg)
...
 
-    vfu_reg_info_t *vfu_reg = &vfu_ctx->reg_info[index];
+    vfu_reg = &vfu_ctx->reg_info[region_idx];
+
+    if (offset + size > vfu_reg->size) {
+        ERROR_INT(EINVAL);
+        return -1;
+    }
+
     ioeventfd_list_t *elem = malloc(sizeof(ioeventfd_list_t));
```
jlevon, line 481: nit, this makes it sounds like it's a list, but each element in as an ioeventfd, so shouldn't this be called "ioeventfd_t" ? Or given the name of the list, "subregion_t" ?

lib/libvfio-user.c
```diff
@@ -462,186 +462,198 @@ handle_device_get_region_info(vfu_ctx_t *vfu_ctx, vfu_msg_t *msg)
...
+vfu_delete_ioeventfd(vfu_ctx_t *vfu_ctx, uint32_t region_idx, size_t offset,
+                     uint32_t size, uint32_t fd_index, uint32_t flags)
 {
+    vfu_reg_info_t *vfu_reg = NULL;
+    ioeventfd_list_t *rem = NULL;
+    size_t current_fd_index = 0;
+    int fd = -1;
 
     assert(vfu_ctx != NULL);
-    int tmp = eventfd(0, 0);
```
jlevon, line 506: should check for region_idx too

lib/libvfio-user.c
```diff
@@ -462,186 +462,198 @@ handle_device_get_region_info(vfu_ctx_t *vfu_ctx, vfu_msg_t *msg)
...
-    assert(vfu_ctx != NULL);
-    vfu_reg_info_t *vfu_reg = &vfu_ctx->reg_info[index];
-    ioeventfd_list_t *rem = LIST_FIRST(&vfu_reg->sub_reg_l);
-
-    uint32_t i = 0;
-    for (i = 0; i < fd_index; i ++) {
-        if (rem == NULL) {
-            break;
+    LIST_FOREACH(rem, &vfu_reg->subregions, entry) {
+        if (current_fd_index == fd_index) {
```
jlevon, line 510: as above, just use "fd" like all the other params.

lib/libvfio-user.c
```diff
@@ -462,186 +462,198 @@ handle_device_get_region_info(vfu_ctx_t *vfu_ctx, vfu_msg_t *msg)
...
-        return -1;
+        if (rem->offset == offset && rem->size == size &&
+            rem->fd == fd && rem->flags == flags ) {
+                LIST_REMOVE(rem, entry);
+                free(rem);
+                return 0;
+        }
+        LIST_NEXT(rem, entry);
+        current_fd_index++;
     }
```
jlevon, line 522: need  to set ENOENT or somesuch here

lib/libvfio-user.c
```diff
@@ -462,186 +462,198 @@ handle_device_get_region_info(vfu_ctx_t *vfu_ctx, vfu_msg_t *msg)
...
-
-    return 0;
+    return -1;
 }
 
 static void
-cleanup_ioeventfds(vfu_ctx_t *vfu_ctx)
+free_regions(vfu_ctx_t *vfu_ctx)
 {
+    int index = 0;
```
jlevon, line 529: nit, size_t please

lib/libvfio-user.c
```diff
@@ -462,186 +462,198 @@ handle_device_get_region_info(vfu_ctx_t *vfu_ctx, vfu_msg_t *msg)
...
+                           sizeof(vfio_user_sub_region_ioeventfd_t));
+    if (msg->out_data == NULL) {
+        return -1;
+    }
+    reply = msg->out_data;
+    reply->index = req->index;
+    reply->count = max_sent_sub_regions;
+    reply->flags = 0;
+    reply->argsz = sizeof(vfio_user_region_io_fds_reply_t) +
+                          max_sent_sub_regions *
```
jlevon, line 630: still doesn't look right, this should be nr_sub_reg, not what we actually send

test/py/libvfio_user.py
```diff
@@ -359,14 +363,14 @@ class sub_region_ioregionfd(Structure):
...
     ]
 
-class sub_region_io_fd(c.Union):
+class vfio_user_sub_region_io_fd(c.Union):
     _pack_ = 1
     _fields_ = [
-        ("sub_region_ioeventfd", sub_region_ioeventfd),
-        ("sub_region_ioregionfd",sub_region_ioregionfd)
+        ("sub_region_ioeventfd", vfio_user_sub_region_ioeventfd),
+        ("sub_region_ioregionfd",vfio_user_sub_region_ioregionfd)
```
jlevon, line 370: nit, space after comma

test/py/libvfio_user.py
```diff
@@ -558,22 +564,27 @@ def msg(ctx, sock, cmd, payload, expect=0, fds=None):
     vfu_run_ctx(ctx)
     return get_reply(sock, expect=expect)
 
-def get_reply_fd(sock, expect=0):
+def get_reply_fds(sock, expect=0):
+    """Recvies a message from a socket and pulls the returned file descriptors
```
jlevon, line 568: "Receives"

test/py/libvfio_user.py
```diff
@@ -558,22 +564,27 @@ def msg(ctx, sock, cmd, payload, expect=0, fds=None):
...
+                                                              data[0:16])
     assert errno == expect
 
-    cmsgLevel, cmsgType, packedFD = ancillary[0] if len(ancillary) != 0 else (0, 0, [])
-    unpackedFDs = []
-    for i in range(0, len(packedFD), 4):
-        [unpackedFD] = struct.unpack_from("i", packedFD, offset = i)
-        unpackedFDs.append(unpackedFD)
-    assert len(packedFD)/4 == len(unpackedFDs)
+    cmsgLevel, cmsgType, packed_FD = ancillary[0] if len(ancillary) != 0 else \
```
jlevon, line 577: still "packed_fd" etc. here please

test/py/test_device_get_region_io_fds.py
```diff
@@ -33,149 +31,232 @@ def test_device_get_region_io_fds_setup():
...
     ret = vfu_realize_ctx(ctx)
     assert ret == 0
 
     sock = connect_client(ctx)
-
     for i in range(0,6):
         tmp = eventfd(0,0)
-        assert vfu_create_ioeventfd(ctx, i * 8, 8, tmp, 0, VFU_PCI_DEV_BAR2_REGION_IDX) != -1
+        fds.append(tmp)
+        assert vfu_create_ioeventfd(ctx, VFU_PCI_DEV_BAR2_REGION_IDX, tmp, \
```
jlevon, line 70: you don't need the backslashes where you have an open bracket

test/py/test_device_get_region_io_fds.py
```diff
@@ -33,149 +31,232 @@ def test_device_get_region_io_fds_setup():
...
+                                    i * SIZE, SIZE, 0, 0) != -1
 
 def test_device_get_region_io_fds_bad_flags():
 
-    payload = region_io_fds_request(argsz = 16+40*5, flags = 1, index = VFU_PCI_DEV_BAR2_REGION_IDX, count = 0)
+    payload = vfio_user_region_io_fds_request(\
+    argsz = ctypes.sizeof(vfio_user_region_io_fds_reply) + \
+        ctypes.sizeof(vfio_user_sub_region_ioeventfd) * 5, flags = 1, \
+                                index = VFU_PCI_DEV_BAR2_REGION_IDX, count = 0)
```
jlevon, line 78: nit, odd indentation here, below

test/py/test_device_get_region_io_fds.py
```diff
@@ -33,149 +31,232 @@ def test_device_get_region_io_fds_setup():
...
-        assert vfu_create_ioeventfd(ctx, i * 8, 8, tmp, 0, VFU_PCI_DEV_BAR2_REGION_IDX) != -1
+        fds.append(tmp)
+        assert vfu_create_ioeventfd(ctx, VFU_PCI_DEV_BAR2_REGION_IDX, tmp, \
+                                    i * SIZE, SIZE, 0, 0) != -1
 
 def test_device_get_region_io_fds_bad_flags():
 
-    payload = region_io_fds_request(argsz = 16+40*5, flags = 1, index = VFU_PCI_DEV_BAR2_REGION_IDX, count = 0)
+    payload = vfio_user_region_io_fds_request(\
+    argsz = ctypes.sizeof(vfio_user_region_io_fds_reply) + \
```
jlevon, line 76: there's a shorter version:

```
len(vfio_user_region_io_fds_reply())
```

that should work for you

test/py/test_device_get_region_io_fds.py
```diff
@@ -33,149 +31,232 @@ def test_device_get_region_io_fds_setup():
...
+        ctypes.sizeof(vfio_user_sub_region_ioeventfd) * 5, flags = 0, \
+                                index = VFU_PCI_DEV_BAR2_REGION_IDX, count = 1)
+
+    msg(ctx, sock, VFIO_USER_DEVICE_GET_REGION_IO_FDS, payload, \
+        expect=errno.EINVAL)
+
+def test_device_get_region_io_fds_buffer_too_small():
+
+    payload = vfio_user_region_io_fds_request(\
+        argsz = ctypes.sizeof(vfio_user_region_io_fds_reply) -1 , flags = 0, \
```
jlevon, line 96: nit, s/ -1 ,/- 1,/

test/py/test_device_get_region_io_fds.py
```diff
@@ -33,149 +31,232 @@ def test_device_get_region_io_fds_setup():
...
 
-def test_device_get_region_io_fds_buffer_to_small():
+def test_device_get_region_io_fds_no_regions_setup():
 
-    payload = region_io_fds_request(argsz = 15, flags = 0, index = VFU_PCI_DEV_BAR2_REGION_IDX, count = 1)
+    payload = vfio_user_region_io_fds_request(argsz = 512, flags = 0, \
+                                index = VFU_PCI_DEV_BAR3_REGION_IDX, count = 0)
 
-    msg(ctx, sock, VFIO_USER_DEVICE_GET_REGION_IO_FDS, payload, expect=errno.EINVAL)
+    ret = msg(ctx, sock, VFIO_USER_DEVICE_GET_REGION_IO_FDS, payload, expect=0)
```
jlevon, line 131: should we really get  errno of 0, if the region isn't even set up? Probably not right?

test/py/test_device_get_region_io_fds.py
```diff
@@ -0,0 +1,181 @@
...
+
+    payload = region_io_fds_request(argsz = 16+40*4, flags = 0, index = VFU_PCI_DEV_BAR2_REGION_IDX, count = 0)
+
+    newfds, ret = msg_fd(ctx, sock, VFIO_USER_DEVICE_GET_REGION_IO_FDS, payload, expect=0)
+
+    reply, ret = region_io_fds_reply.pop_from_buffer(ret)
+    ioevent, ret = sub_region_ioeventfd.pop_from_buffer(ret)
+
+    print(newfds)
+    for i in range(0,4):
```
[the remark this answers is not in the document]
jlevon: both these comments are still current

test/py/test_device_get_region_io_fds.py
```diff
@@ -0,0 +1,181 @@
...
+        os.write(newfds[i], c.c_ulonglong(10))
+        out = os.read(newfds[i], 8)
```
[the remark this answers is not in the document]
jlevon: still here

test/py/test_device_get_region_io_fds.py
```diff
@@ -0,0 +1,181 @@
...
+        [out] = struct.unpack("@Q",out)
+        assert out  == 10
+
```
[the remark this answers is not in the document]
jlevon: ?

test/py/test_device_get_region_io_fds.py
```diff
@@ -33,149 +31,232 @@ def test_device_get_region_io_fds_setup():
...
 def test_device_get_region_io_fds_fds_read_write_nothing():
 
-    payload = region_io_fds_request(argsz = 16, flags = 0, index = VFU_PCI_DEV_BAR2_REGION_IDX, count = 0)
+    payload = vfio_user_region_io_fds_request(\
+            argsz = ctypes.sizeof(vfio_user_region_io_fds_reply), flags = 0, \
+                                index = VFU_PCI_DEV_BAR2_REGION_IDX, count = 0)
 
-    newfds, ret = msg_fd(ctx, sock, VFIO_USER_DEVICE_GET_REGION_IO_FDS, payload, expect=0)
+    _, ret = msg_fds(ctx, sock, VFIO_USER_DEVICE_GET_REGION_IO_FDS, \
+                         payload, expect=0)
```
jlevon, line 201: indentation?

test/py/test_device_get_region_io_fds.py
```diff
@@ -4,14 +4,12 @@
...
-libc = c.cdll.LoadLibrary("libc.so.6")
-def eventfd(init_val, flags):
-    return libc.eventfd(init_val, flags)
-
+import ctypes
 
 ctx = None
 sock = None
+fds = []
+SIZE = 8
```
jlevon, line 41: IOEVENT_SIZE or something maybe?

test/py/test_device_get_region_io_fds.py
```diff
@@ -0,0 +1,181 @@
...
+        assert ioevents[i].offset == 40 - (8 * i)
+
+def test_device_get_region_io_fds_fds_read_write_nothing():
+
+    payload = region_io_fds_request(argsz = 16, flags = 0, index = VFU_PCI_DEV_BAR2_REGION_IDX, count = 0)
+
+    newfds, ret = msg_fd(ctx, sock, VFIO_USER_DEVICE_GET_REGION_IO_FDS, payload, expect=0)
+
+    reply, ret = region_io_fds_reply.pop_from_buffer(ret)
+    assert reply.argsz == 16
```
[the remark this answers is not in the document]
jlevon: not fixed

test/py/test_device_get_region_io_fds.py
```diff
@@ -0,0 +1,181 @@
...
+
+def test_device_get_region_io_fds_fds_read_write_dupe_fd():
```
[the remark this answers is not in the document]
jlevon: not fixed

include/vfio-user.h
```diff
@@ -156,6 +156,45 @@ struct vfio_user_irq_info {
...
+} __attribute__((packed));
+
+struct reply_sub_region {
+    uint32_t argsz;
+    uint32_t flags;
+    uint32_t index;
+    uint32_t count;
+    union sub_region {
+        struct sub_region_ioeventfd ioeventfd;
+        struct sub_region_ioregionfd toregionfd;
```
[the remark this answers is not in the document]
jlevon: not fixed

lib/libvfio-user.c
```diff
@@ -1657,7 +1668,7 @@ vfu_setup_region(vfu_ctx_t *vfu_ctx, int region_idx, size_t size,
     reg->cb = cb;
     reg->fd = fd;
     reg->offset = offset;
-    LIST_INIT(&reg->sub_reg_l);
+    LIST_INIT(&reg->subregions);
```
jlevon, line 1671: I still think this belongs in vfu_create_ctx

jlevon (changes requested):
I've left open a couple of comments that I don't think are fixed yet - please take a look at those as well as the new ones. Once you've made a change you can mark it resolved, or if there's a follow up add a comment?

include/libvfio-user.h
JAKelly10: I don't think we can just use the fd because we never directly return the fd. When we use get_region_io_fds it returns a separate list of fds from the socket which don't have to match the exact number that was used to create the ioeventfds on the server. As such I don't think we can just send the fds as we can never directly get the exact same fds from the server for comparison. As such we would have to pass the index which we get as a return for the sub regions from get_region_io_fds.

lib/libvfio-user.c
JAKelly10: Double checked on our spec doc against the full one in kvm. Our spec doc is missing information about the behavior of the system as it contains explicit reference to the behavior you are referencing here along with needing a change to be behavior when dealing with requests where the buffer is too small. Will start on these required changes and add the paragraphs to our spec doc on the repo.

include/libvfio-user.h
jlevon: get_region_io_fds is for the client.

What I'm saying is, when the API consumer (the device impl like SPDK) wants to delete an ioeventfd, then it knows the fd it's for, along with its identifying information. It doesn't know anything about `fd_index` - that's a protocol-level thing to do with the client.

Now, there's an obvious issue here in that we have no way to tell the *client* that an ioeventfd got deleted. So it's really of limited use. In fact, I'd be inclined not to have it at all the more I think about it, at least until we find a need. So, in terms of "less code", shall we remove all the ioeventfd_delete code?

Apart from the clean up in vfu_destroy_ctx of course.

lib/libvfio-user.c
jlevon: Are you sure there's something missing? When you say "the full one in kvm" - what are you talking about exactly? KVM knows nothing about vfio-user.

This is where in the spec this behaviour is documented:

```
argsz is the size required for the full reply payload (region info structure plus the size of any region capabilities)
```

plus a longer explanation in https://github.com/nutanix/libvfio-user/blob/master/docs/vfio-user.rst#message-sizes

include/libvfio-user.h
JAKelly10: Yes I agree with the idea to remove it due to the lack of notification to the client as that is certain to cause a problem with out extreme care being taken.

lib/libvfio-user.c
JAKelly10: moved

lib/libvfio-user.c
JAKelly10: Sorry for the previous comment, I had found archives of the mailing lists with the draft proposal that had 2 additional paragraphs relating to the explicit behavior of argsz which appear to have been removed at some point and is instead explained in message size.


From the draft version of the spec
```
> The client sets the ``argsz`` field to indicate the maximum size of the response
> that the server can send, which must be at least the size of the response header
> plus space for the subregion array. If the full response size exceeds ``argsz``,
> then the server must respond only with the response header and the Region IO FD
> info structure, setting in ``argsz`` the buffer size required to store the full
> response. In this case, no file descriptors are passed back.  The client then
> retries the operation with a larger receive buffer.
```

test/py/test_device_get_region_io_fds.py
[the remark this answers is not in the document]
JAKelly10: updated with 2 loops and comments

test/py/test_device_get_region_io_fds.py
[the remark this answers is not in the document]
JAKelly10: Added a loop to close all of the fds returned from newfds

lib/libvfio-user.c
```diff
@@ -459,6 +461,182 @@ handle_device_get_region_info(vfu_ctx_t *vfu_ctx, vfu_msg_t *msg)
...
+        return ERROR_INT(EINVAL);
+    }
+
+    if (req->index >= vfu_ctx->nr_regions) {
+        vfu_log(vfu_ctx, LOG_DEBUG, "bad region index %d in get region io fds "
+                "info", req->index);
+        return ERROR_INT(EINVAL);
+    }
+
+    if (req->index >= VFU_PCI_DEV_NUM_REGIONS) {
```
jlevon, line 575: you can remove this

lib/libvfio-user.c
```diff
@@ -459,6 +461,182 @@ handle_device_get_region_info(vfu_ctx_t *vfu_ctx, vfu_msg_t *msg)
...
+    if (req->argsz < sizeof(vfio_user_region_io_fds_reply_t) ||
+        req->argsz > SERVER_MAX_DATA_XFER_SIZE) {
+        return ERROR_INT(EINVAL);
+    }
+
+    max_sent_sub_regions = MIN((req->argsz -
+                                sizeof(vfio_user_region_io_fds_reply_t)) /
+                                sizeof(vfio_user_sub_region_ioeventfd_t),
+                                nr_sub_reg);
+    max_sent_sub_regions = (max_sent_sub_regions >= nr_sub_reg) ? nr_sub_reg : 0;
```
jlevon, line 598: this is a little weird now. Really you're saying "if we can't fit nr_sub_reg, then don't add anything". But what you do now looks a bit confusing, and in fact you end up doing a calloc of size 0 in this case, on line :617

Instead, take a look at how we do this in `handle_device_get_region_info`, and do something like this:

```
region_array_size = nr_sub_reg * sizeof(vfio_user_sub_region_ioeventfd_t);
...
msg->out_size = MIN(sizeof(*reply) + region_array_size, in_info->argsz);
msg->out_data = calloc(1, msg->out_size);

...

reply = msg->out_data;
reply->argsz = sizeof(*reply) + region_array_size;
...
/* Only provide sub-region info if it fits. */
if (in_info->argsz >= out_info->argsz) {
   /* alloc/set ->out_fds here too, then output each region... */
   sub_reg = LIST_FIRST(...
}
```

make sense?

lib/private.h
```diff
@@ -133,6 +134,8 @@ typedef struct  {
     int fd;
     /* offset of region within fd. */
     uint64_t offset;
+    /* The subregions for ioregions and ioevents */
```
jlevon, line 137: nit, "ioregionfds" and "ioeventfds"

test/py/libvfio_user.py
```diff
@@ -502,6 +561,40 @@ def msg(ctx, sock, cmd, payload, expect=0, fds=None):
...
+    """Receives a message from a socket and pulls the returned file descriptors
+       out of the message."""
+    fds = array.array("i")
+    data, ancillary, flags, addr = sock.recvmsg(4096,
+                                            socket.CMSG_LEN(64 * fds.itemsize))
+    (msg_id, cmd, msg_size, msg_flags, errno) = struct.unpack("HHIII",
+                                                              data[0:16])
+    assert errno == expect
+
+    cmsgLevel, cmsgType, packed_fd = ancillary[0] if len(ancillary) != 0 else \
```
jlevon, line 574: still camel case!

jlevon (changes requested):
Just a couple of nits, plus one cleanup I'd suggest with the argsz handling

lib/libvfio-user.c
JAKelly10: yep fixing.

jlevon (approved):

PR #604, opened 2021-09-24
VFIO_USER_DEVICE_GET_REGION_IO_FDS fix

JAKelly10:
Fixes 2 issues where VFIO_USER_DEVICE_GET_REGION_IO_FDS fails on regions with an fd of -1 and following a patch for that allowing commands on unset up regions.

Both issues are fixed by this patch


jlevon:
split into separate PRS #606 and #607

PR #606, opened 2021-09-27
fix issue with get_region_io_fds when fd = -1

jlevon:


lib/libvfio-user.c
```diff
@@ -558,7 +558,7 @@ handle_device_get_region_io_fds(vfu_ctx_t *vfu_ctx, vfu_msg_t *msg)
     assert(msg != NULL);
     assert(msg->out_fds == NULL);
 
-    if (msg->in_size != sizeof(vfio_user_region_io_fds_request_t)) {
+    if (msg->in_size < sizeof(vfio_user_region_io_fds_request_t)) {
```
tmakatos, line 561: Is this related to this issue?

lib/libvfio-user.c
```diff
@@ -576,10 +576,6 @@ handle_device_get_region_io_fds(vfu_ctx_t *vfu_ctx, vfu_msg_t *msg)
 
     vfu_reg = &vfu_ctx->reg_info[req->index];
 
-    if (vfu_reg->fd == -1) {
```
tmakatos, line 579: I'm lacking some context, IIUC the problem is that we don't allow a region to have ioregionfd sub-regions if the region is not memory mappable? ioregionfd and memory-mappable region are orthogonal, right?

jlevon: right, exactly.

lib/libvfio-user.c
tmakatos: no it's fine, just wanted to make sure I'm not missing something

lib/libvfio-user.c
tmakatos: Could we document that on the commit message?

tmakatos (approved):

lib/libvfio-user.c
jlevon: I tried to clarify the commit message a bit more?

tmakatos: You did, I only looked at the PR title.. my bad.

PR #607, opened 2021-09-27
disable VFIO_USER_DEVICE_GET_REGION_IO_FDS on invalid region

jlevon:


lib/libvfio-user.c
```diff
@@ -576,6 +576,11 @@ handle_device_get_region_io_fds(vfu_ctx_t *vfu_ctx, vfu_msg_t *msg)
 
     vfu_reg = &vfu_ctx->reg_info[req->index];
 
+    // At least one flag must be set for a valid region.
+    if (!(vfu_reg->flags & VFU_REGION_FLAG_MASK)) {
```
tmakatos, line 580: Won't we have failed to set up this region in the first place?

jlevon: Sure, but we shouldn't allow an incorrect consumer to ask for iofds for such an unsetup region.

tmakatos (approved):

PR #608, opened 2021-10-01
make migration state callback optionally asynchronous

tmakatos:
Some devices need the migration state callback to be asynchronous. The
simplest way to implement this is to required from the callback to
return EBUSY, not process any other new messages (return EBUSY), and
provide a way to the user to complete migration (vfu_migr_done).

### DIFF
--- a/include/libvfio-user.h
+++ b/include/libvfio-user.h
@@ -586,10 +586,10 @@ typedef struct {
  * behavior.
  *
  * @vfu_ctx: the libvfio-user context
- * @err: 0 for success, -1 on error having set errno.
+ * @reply_errno: 0 for success or errno on error.
  */
 void
-vfu_migr_done(vfu_ctx_t *vfu_ctx, int err);
+vfu_migr_done(vfu_ctx_t *vfu_ctx, int reply_errno);
 
 
 #ifndef VFIO_DEVICE_STATE_STOP
--- a/include/libvfio-user.h
+++ b/include/libvfio-user.h
@@ -501,7 +501,13 @@ typedef struct {
     /*
      * Migration state transition callback.
      *
-     * The callback should return -1 on error, setting errno.
+     * The callback should return 0 on success and -errno on error.
+     *
+     * When operating in non-blocking mode (LIBVFIO_USER_FLAG_ATTACH_NB was
+     * passed to vfu_create_ctx) and -EBUSY is returned, transitioning to the
+     * new state becomes asynchronous: libvfio-user does not send a response to
+     * the client and does not process any new messages.  Transitioning to the
+     * new device state is completed by calling vfu_migr_done.
      *
      * TODO rename to vfu_migration_state_transition_callback
      * FIXME maybe we should create a single callback and pass the state?
@@ -568,6 +574,17 @@ typedef struct {
     int (*data_written)(vfu_ctx_t *vfu_ctx, uint64_t count);
 
 } vfu_migration_callbacks_t;
+
+/*
+ * Completes a pending migration state transition. Calling this function when
+ * there is no pending migration state transition results in undefined
+ * behavior.
+ *
+ * @vfu_ctx: the libvfio-user context
+ * @err: 0 for success, -errno for error..
+ */
+void
+vfu_migr_done(vfu_ctx_t *vfu_ctx, int err);
 
 
 #ifndef VFIO_DEVICE_STATE_STOP
--- a/include/libvfio-user.h
+++ b/include/libvfio-user.h
@@ -512,7 +512,7 @@ typedef struct {
      * threading model does not allow blocking.
      *
      * The user must not call functions vfu_dma_read or vfu_dma_write, doing so
-     * rresults in undefined behavior.
+     * results in undefined behavior.
      *
      * TODO rename to vfu_migration_state_transition_callback
      * FIXME maybe we should create a single callback and pass the state?
--- a/include/libvfio-user.h
+++ b/include/libvfio-user.h
@@ -501,13 +501,18 @@ typedef struct {
     /*
      * Migration state transition callback.
      *
-     * The callback should return 0 on success and -errno on error.
+     * The callback should return -1 on error, setting errno.
      *
      * When operating in non-blocking mode (LIBVFIO_USER_FLAG_ATTACH_NB was
-     * passed to vfu_create_ctx) and -EBUSY is returned, transitioning to the
-     * new state becomes asynchronous: libvfio-user does not send a response to
-     * the client and does not process any new messages.  Transitioning to the
-     * new device state is completed by calling vfu_migr_done.
+     * passed to vfu_create_ctx) and -1 is returned with errno set to EBUSY,
+     * transitioning to the new state becomes asynchronous: libvfio-user does
+     * not send a response to the client and does not process any new messages.
+     * Transitioning to the new device state is completed by calling
+     * vfu_migr_done. This behavior can be beneficial for devices whose
+     * threading model does not allow blocking.
+     *
+     * The user must not call functions vfu_dma_read or vfu_dma_write, doing so
+     * rresults in undefined behavior.
      *
      * TODO rename to vfu_migration_state_transition_callback
      * FIXME maybe we should create a single callback and pass the state?
@@ -581,7 +586,7 @@ typedef struct {
  * behavior.
  *
  * @vfu_ctx: the libvfio-user context
- * @err: 0 for success, -errno for error..
+ * @err: 0 for success, -1 on error having set errno.
  */
 void
 vfu_migr_done(vfu_ctx_t *vfu_ctx, int err);
--- a/lib/libvfio-user.c
+++ b/lib/libvfio-user.c
@@ -1291,17 +1291,18 @@ process_request(vfu_ctx_t *vfu_ctx)
     }
 
 out:
+    if (vfu_ctx->migr_trans_pending) {
+        assert(ret == 0);
+        vfu_ctx->migr_trans_msg = msg;
+        /* NB the message is freed in vfu_migr_done */
+        return 0;
+    }
     if (msg->hdr.flags.no_reply) {
         /*
          * A failed client request is not a failure of process_request() itself.
          */
         ret = 0;
     } else {
-        if (vfu_ctx->migr_trans_pending) {
-            vfu_ctx->migr_trans_msg = msg;
-            /* NB the message is freed in vfu_migr_done */
-            return 0;
-        }
 
         ret = do_reply(vfu_ctx, msg, ret);
     }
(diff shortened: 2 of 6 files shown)


include/libvfio-user.h
```diff
@@ -501,7 +501,13 @@ typedef struct {
     /*
      * Migration state transition callback.
      *
-     * The callback should return -1 on error, setting errno.
+     * The callback should return 0 on success and -errno on error.
```
jlevon, line 504: we moved to expecting -1 + errno, a while back

include/libvfio-user.h
```diff
@@ -569,6 +575,17 @@ typedef struct {
 
 } vfu_migration_callbacks_t;
 
+/*
+ * Completes a pending migration state transition. Calling this function when
+ * there is no pending migration state transition results in undefined
+ * behavior.
+ *
+ * @vfu_ctx: the libvfio-user context
+ * @err: 0 for success, -errno for error..
```
jlevon, line 584: ditto

include/libvfio-user.h
```diff
@@ -501,7 +501,13 @@ typedef struct {
...
      * Migration state transition callback.
      *
-     * The callback should return -1 on error, setting errno.
+     * The callback should return 0 on success and -errno on error.
+     *
+     * When operating in non-blocking mode (LIBVFIO_USER_FLAG_ATTACH_NB was
+     * passed to vfu_create_ctx) and -EBUSY is returned, transitioning to the
+     * new state becomes asynchronous: libvfio-user does not send a response to
+     * the client and does not process any new messages.  Transitioning to the
+     * new device state is completed by calling vfu_migr_done.
```
jlevon, line 510: should this explain what could happen if you call e.g. vfu_dma_read() ?
Or one day, with tcp-ip transport, vfu_irq_trigger() ?

include/libvfio-user.h
```diff
@@ -501,7 +501,13 @@ typedef struct {
     /*
      * Migration state transition callback.
      *
-     * The callback should return -1 on error, setting errno.
+     * The callback should return 0 on success and -errno on error.
+     *
+     * When operating in non-blocking mode (LIBVFIO_USER_FLAG_ATTACH_NB was
```
jlevon, line 506: is there a specific reason to require this?

lib/libvfio-user.c
```diff
@@ -1209,6 +1213,29 @@ exec_command(vfu_ctx_t *vfu_ctx, vfu_msg_t *msg)
     return ret;
 }
 
+static int
+do_reply(vfu_ctx_t *vfu_ctx, vfu_msg_t *msg, int ret)
```
jlevon, line 1217: now you factored this out, could we not re-use "ret", have the arg be "reply_ret" or something? This always caught me off guard

lib/libvfio-user.c
```diff
@@ -1943,4 +1966,20 @@ vfu_sg_is_mappable(vfu_ctx_t *vfu_ctx, dma_sg_t *sg)
     return dma_sg_is_mappable(vfu_ctx->dma, sg);
 }
 
+EXPORT void
+vfu_migr_done(vfu_ctx_t *vfu_ctx, int err)
+{
+    assert(vfu_ctx != NULL);
+    assert(vfu_ctx->migr_trans_pending);
+
+    if (vfu_ctx->migr_trans_msg != NULL) {
```
jlevon, line 1975: when can this be null ?

lib/migration.c
```diff
@@ -423,7 +428,7 @@ MOCK_DEFINE(migration_region_access_registers)(vfu_ctx_t *vfu_ctx, char *buf,
                 "migration: transition from state %s to state %s",
                  migr_states[old_device_state].name,
                  migr_states[*device_state].name);
-        } else {
+        } else if (errno != EBUSY) {
```
jlevon, line 431: seems worth a LOG_DEBUG?

lib/private.h
```diff
@@ -169,6 +169,8 @@ struct vfu_ctx {
     size_t                  client_max_data_xfer_size;
 
     struct migration        *migration;
+    bool                    migr_trans_pending;
+    vfu_msg_t               *migr_trans_msg;
```
jlevon, line 173: maybe this is in the followup but I'd prefer "pending_msg" or something?

test/py/test_migration.py
```diff
@@ -0,0 +1,144 @@
+#
```
jlevon, line 1: thanks for adding this!!

lib/libvfio-user.c
```diff
@@ -1376,6 +1396,9 @@ vfu_run_ctx(vfu_ctx_t *vfu_ctx)
     blocking = !(vfu_ctx->flags & LIBVFIO_USER_FLAG_ATTACH_NB);
 
     do {
+        if (vfu_ctx->migr_trans_pending) {
```
jlevon, line 1399: I think we need the same for at least vfu_dma_read/write()
For vfu_irq_trigger() we can (probably?) get away with it for now, until it needs the socket.

include/libvfio-user.h
```diff
@@ -501,7 +501,13 @@ typedef struct {
...
+     * passed to vfu_create_ctx) and -EBUSY is returned, transitioning to the
+     * new state becomes asynchronous: libvfio-user does not send a response to
+     * the client and does not process any new messages.  Transitioning to the
```
jlevon, line 509: Maybe give a little detail on why this is useful?

include/libvfio-user.h
tmakatos: I know, the problem is that setting `errno` in a Python callback when unit testing isn't visible in libvfio-user:
> The use_errno parameter, when set to true, enables a ctypes mechanism that allows accessing the system errno error number in a safe way. ctypes maintains a thread-local copy of the systems errno variable; if you call foreign functions created with use_errno=True then the errno value before the function call is swapped with the ctypes private copy, the same happens immediately after the function call.

Do you have any ideas how to solve this?

include/libvfio-user.h
tmakatos: Yes we should explicitly state that vfu_dma_read/write and vfu_irq_trigger are not allowed.

include/libvfio-user.h
tmakatos: Purely for simplicity since it's only SPDK that requires it and it doesn't use non-blocking mode. I haven't thought of the implications in allowing this in blocking mode and since we don't need it now I think it's best to postpone implementing this.

lib/migration.c
tmakatos: ok

lib/private.h
tmakatos: We do rename it to `pending_msg` in https://github.com/nutanix/libvfio-user/pull/609.

lib/libvfio-user.c
tmakatos: I've documented that it isn't allowed so I'll only add an `assert`.

include/libvfio-user.h
jlevon: I tested this, and you need to make sure your callback type *also* has "use_errno=True" e.g.

```
 vfu_region_access_cb_t = c.CFUNCTYPE(c.c_int, c.c_void_p, c.POINTER(c.c_char),
-                                     c.c_ulong, c.c_long, c.c_bool)
+                                     c.c_ulong, c.c_long, c.c_bool,
+                                     use_errno=True)
```

then you can use `c.set_errno()` inside the callback, and it should work OK

include/libvfio-user.h
jlevon: sure

include/libvfio-user.h
tmakatos: Well that was easy, thanks!

include/libvfio-user.h
tmakatos: Done, let me know if you want it rephrased.

lib/libvfio-user.c
```diff
@@ -1263,19 +1290,12 @@ process_request(vfu_ctx_t *vfu_ctx)
...
-
-            if (errno == ECONNRESET) {
-                vfu_reset_ctx(vfu_ctx, "reset");
-                errno = ENOTCONN;
-            } else if (errno == ENOMSG) {
-                vfu_reset_ctx(vfu_ctx, "closed");
-                errno = ENOTCONN;
-            }
+        if (vfu_ctx->migr_trans_pending) {
+            vfu_ctx->migr_trans_msg = msg;
```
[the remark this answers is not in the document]
tmakatos: sure

include/libvfio-user.h
tmakatos: I'll drop a comment for future reference.

lib/libvfio-user.c
tmakatos: When the no_reply bit is set in the source message.

jlevon: Hmm I think I'd rather always stash the msg, so we can refer to it, then respect the no_reply here as needed.

include/libvfio-user.h
```diff
@@ -503,6 +503,17 @@ typedef struct {
...
+     * When operating in non-blocking mode (LIBVFIO_USER_FLAG_ATTACH_NB was
+     * passed to vfu_create_ctx) and -1 is returned with errno set to EBUSY,
+     * transitioning to the new state becomes asynchronous: libvfio-user does
+     * not send a response to the client and does not process any new messages.
+     * Transitioning to the new device state is completed by calling
+     * vfu_migr_done. This behavior can be beneficial for devices whose
+     * threading model does not allow blocking.
+     *
+     * The user must not call functions vfu_dma_read or vfu_dma_write, doing so
+     * rresults in undefined behavior.
```
swapnili, line 515: typo: results

swapnili (commented):
Looks good to me, just minor nit. I will let John approve it, as he is still reviewing it.

include/libvfio-user.h
tmakatos: fixed

include/libvfio-user.h
```diff
@@ -569,6 +580,17 @@ typedef struct {
 
 } vfu_migration_callbacks_t;
 
+/*
+ * Completes a pending migration state transition. Calling this function when
+ * there is no pending migration state transition results in undefined
+ * behavior.
+ *
+ * @vfu_ctx: the libvfio-user context
+ * @err: 0 for success, -1 on error having set errno.
```
jlevon, line 589: this argument should surely just be the errno?

lib/libvfio-user.c
```diff
@@ -335,7 +335,18 @@ handle_region_access(vfu_ctx_t *vfu_ctx, vfu_msg_t *msg)
 
     ret = region_access(vfu_ctx, in_ra->region, buf, in_ra->count,
                         in_ra->offset, msg->hdr.cmd == VFIO_USER_REGION_WRITE);
-
+    if (ret == -1 && in_ra->region == VFU_PCI_DEV_MIGR_REGION_IDX
+        && errno == EBUSY && vfu_ctx->flags & LIBVFIO_USER_FLAG_ATTACH_NB) {
```
jlevon, line 339: sorry, small nit: brackets around the vfu_ctx->flags & LIBVFIO_USER_FLAG_ATTACH_NB?

lib/libvfio-user.c
```diff
@@ -1209,6 +1220,29 @@ exec_command(vfu_ctx_t *vfu_ctx, vfu_msg_t *msg)
     return ret;
 }
 
+static int
+do_reply(vfu_ctx_t *vfu_ctx, vfu_msg_t *msg, int reply_ret)
```
jlevon, line 1224: this arg should just the errno (or 0) too

lib/private.h
jlevon: sure

include/libvfio-user.h
tmakatos: ok

lib/libvfio-user.c
tmakatos: ok

lib/libvfio-user.c
tmakatos: ok

jlevon (approved):
Thanks!

Issue #598, opened 2021-08-25
rename vfu_map/unmap_sg()


jlevon:
In at least one way, this is really confusing naming - we don't actually map or unmap with these operations, they just convert an sg to an iov. Can we name better?

swapnili:
One of the suggestion is,
vfu_get_iovec() / vfu_put_iovec()

jlevon:
fixed a while back

Issue #595, opened 2021-08-19
mark dirty at unmap time


jlevon:
It's late but isn't this a bug:

![image](https://user-images.githubusercontent.com/21101340/130109494-663b9def-a28e-45d6-9862-e9cb63e68d68.png)

and can't we fix it simply by marking dirty in `vfu_unmap_sg()` ? If so, do we even need `->maps` ?

jlevon:
cc @tmakatos @swapnili

swapnili:
> can't we fix it simply by marking dirty in vfu_unmap_sg() 

Isn't the dirty page tracking is required only when device implementations maps the SG to do IO on it. It is not clear to me how marking dirty in vfu_unmap_sg() will achieve this.

jlevon:
> how marking dirty in vfu_unmap_sg() will achieve this.

By definition when the caller calls vfu_unmap_sg(), it's finished any possible writes to the mapped region.

The question is whether the final stage of live migration (downtime) can happen when there's still an active vfu_map_sg(), or will the device have finished and called vfu_unmap_sg() by that point? If so we can use unmap, and we don't have the bug above.

If that's the not the case, it seems fundamentally racy anyway.

Otherwise, presuming I'm right about the bug above, we will need the consumer to explicitly mark when a page has been written to.

jlevon:
closing as this isn't quite correct: there's no bug today because after clearing the bitmap, we immediately mark dirty all extant maps again

Issue #594, opened 2021-08-19
vfu_map_sg() concurrency


jlevon:
Currently, vfu_addr_to_sg(), vfu_map_sg() and vfu_unmap_sg() cannot be used in a concurrent manner. This is for a few reasons:

1. the (unused) region refcnt is not atomic
2. the ->maps linked list is not atomic
3. unless otherwise prevented, a concurrent `VFIO_IOMMU_DIRTY_PAGES_FLAG_GET_BITMAP` can race with dirty bitmap tracking of `vfu_map_sg()`
4. unless otherwise prevented, a concurrent `VFIO_USER_DMA_UNMAP` can race with region dereference in above calls.
5. same can happen in response to any context-using call in the case the socket client hangs up and we need to tear down the DMA controller - that will result in unmap callbacks

However SPDK would like to be able share libvfio-user context across reactors/pthreads. vfu_run_ctx can likely be pthread-protected, in the expectation there's only one poller on library state (so no contention), which leaves the above I/O handling code. Let's presume that SPDK can provide exclusion such that no active sg maps are present across a DMA unmap. The question becomes can be support multi-threaded use of the above three APIs.

1. We can remove the refcnt altogether.
2. we'd have to make the dirty bitmap writing atomic op
3. we'd have to use some lockless structure for ->maps or otherwise approach the problem
4. if a dirty pages get call comes in, that's currently not safe against a concurrent vfu_map_sg() marking things dirty

As a reminder, `->maps` exists so that if a consumer does `vfu_map_sg()` and *then* we are asked to enable dirty page tracking, we won't mark those pages as dirty, and will hence potentially lose data. Note that we actually only need that information during that transition, to catch vfu_map_sg() that happened just before dirty page tracking (and hence didn't mark dirty).

If we could ensure there were no inflight DMAs at the time of enabling dirty page tracking, we wouldn't need the map list at all.

Looking at this again, though, it looks a bit dubious anyway.

jraman567:
I'm wondering if a `pthread_rwlock` would be a fair compromise between performance between thread-safety.

Any thread that needs to perform vfu_map_sg() with vfu_ctx could take the pthread_rwlock_rdlock() - which can be taken by multiple threads concurrently.

Any thread that wants to perform the unmap operation could consider taking the `pthread_rwlock_wrlock()` - which prevent further rdlocks from being issues and waits for existing rdlocks to be returned.

jlevon:
[...]

PR #609, opened 2021-10-04
introduce device quiesce callback

tmakatos:

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -2,7 +2,7 @@ name: pull_request
 on: pull_request
 jobs:
   ubuntu-20:
-    timeout-minutes: 5
+    timeout-minutes: 10
     runs-on: ubuntu-20.04
     steps:
       - uses: actions/checkout@v2
@@ -13,7 +13,7 @@ jobs:
                                   python3-pytest debianutils flake8
           make pre-push VERBOSE=1
   ubuntu-18:
-    timeout-minutes: 5
+    timeout-minutes: 10
     runs-on: ubuntu-18.04
     steps:
       - uses: actions/checkout@v2
@@ -25,7 +25,7 @@ jobs:
                                   python3-pytest debianutils
           make pre-push VERBOSE=1
   centos-7:
-    timeout-minutes: 5
+    timeout-minutes: 10
     runs-on: ubuntu-latest
     container: centos:7
     steps:
@@ -37,15 +37,15 @@ jobs:
           yum -y install clang cmake json-c-devel libcmocka-devel \
               openssl-devel valgrind python36-pytest which
           make pre-push VERBOSE=1
-  fedora-34:
-    timeout-minutes: 5
+  fedora-35:
+    timeout-minutes: 10
     runs-on: ubuntu-latest
     container: fedora
     steps:
       - uses: actions/checkout@v2
       - name: pre-push
         run: |
-          dnf -y install --releasever=34 \
+          dnf -y install --releasever=35 \
               gcc make clang cmake json-c-devel libcmocka-devel openssl-devel \
               pciutils diffutils valgrind python3-pytest python3-flake8 which
           make pre-push VERBOSE=1
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -10,7 +10,7 @@ jobs:
         run: |
           sudo apt-get update
           sudo apt-get -y install libjson-c-dev libcmocka-dev clang valgrind \
-                                  python3-pytest debianutils
+                                  python3-pytest debianutils flake8
           make pre-push VERBOSE=1
   ubuntu-18:
     timeout-minutes: 5
@@ -19,6 +19,7 @@ jobs:
       - uses: actions/checkout@v2
       - name: pre-push
         run: |
+          # NB: no working flake8
           sudo apt update
           sudo apt-get -y install libjson-c-dev libcmocka-dev clang valgrind \
                                   python3-pytest debianutils
@@ -31,6 +32,7 @@ jobs:
       - uses: actions/checkout@v2
       - name: pre-push
         run: |
+          # NB: no working flake8
           yum -y install make gcc-4.8.5 epel-release pciutils
           yum -y install clang cmake json-c-devel libcmocka-devel \
               openssl-devel valgrind python36-pytest which
@@ -45,5 +47,5 @@ jobs:
         run: |
           dnf -y install --releasever=34 \
               gcc make clang cmake json-c-devel libcmocka-devel openssl-devel \
-              pciutils diffutils valgrind python3-pytest which
+              pciutils diffutils valgrind python3-pytest python3-flake8 which
           make pre-push VERBOSE=1
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -2,7 +2,7 @@ name: pull_request
 on: pull_request
 jobs:
   ubuntu-20:
-    timeout-minutes: 5
+    timeout-minutes: 10
     runs-on: ubuntu-20.04
     steps:
       - uses: actions/checkout@v2
@@ -13,7 +13,7 @@ jobs:
                                   python3-pytest debianutils flake8
           make pre-push VERBOSE=1
   ubuntu-18:
-    timeout-minutes: 5
+    timeout-minutes: 10
     runs-on: ubuntu-18.04
     steps:
       - uses: actions/checkout@v2
@@ -25,7 +25,7 @@ jobs:
                                   python3-pytest debianutils
           make pre-push VERBOSE=1
   centos-7:
-    timeout-minutes: 5
+    timeout-minutes: 10
     runs-on: ubuntu-latest
     container: centos:7
     steps:
@@ -38,7 +38,7 @@ jobs:
               openssl-devel valgrind python36-pytest which
           make pre-push VERBOSE=1
   fedora-34:
-    timeout-minutes: 5
+    timeout-minutes: 10
     runs-on: ubuntu-latest
     container: fedora
     steps:
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -2,7 +2,7 @@ name: pull_request
 on: pull_request
 jobs:
   ubuntu-20:
-    timeout-minutes: 10
+    timeout-minutes: 5
     runs-on: ubuntu-20.04
     steps:
       - uses: actions/checkout@v2
@@ -13,7 +13,7 @@ jobs:
                                   python3-pytest debianutils flake8
           make pre-push VERBOSE=1
   ubuntu-18:
-    timeout-minutes: 10
+    timeout-minutes: 5
     runs-on: ubuntu-18.04
     steps:
       - uses: actions/checkout@v2
@@ -25,7 +25,7 @@ jobs:
                                   python3-pytest debianutils
           make pre-push VERBOSE=1
   centos-7:
-    timeout-minutes: 10
+    timeout-minutes: 5
     runs-on: ubuntu-latest
     container: centos:7
     steps:
@@ -38,7 +38,7 @@ jobs:
               openssl-devel valgrind python36-pytest which
           make pre-push VERBOSE=1
   fedora-34:
-    timeout-minutes: 10
+    timeout-minutes: 5
     runs-on: ubuntu-latest
     container: fedora
     steps:
--- a/README.md
+++ b/README.md
@@ -92,14 +92,7 @@ run specific tests use the pytest `-k` option in `PYTESTCMD` in the Makefile.
 Coverity
 ========
 
-Build according to:
-
-	https://scan.coverity.com/download?tab=cxx
-
-E.g.:
-
-	../cov-analysis-linux64-2019.03/bin/cov-build --dir cov-int make
-
+`make coverity` automatically uploads a new coverity build.
 
 Supported features
 ==================
--- a/README.md
+++ b/README.md
@@ -77,6 +77,30 @@ The kernel headers are necessary because VFIO structs and defines are reused.
 
 Finally build your program and link with `libvfio-user.so`.
 
+
+Testing
+=======
+
+
+Debugging Test Errors
+---------------------
+
+Sometimes debugging Valgrind errors on Python unit tests can be tricky. To
+run specific tests use the pytest `-k` option in `PYTESTCMD` in the Makefile.
+
+
+Coverity
+========
+
+Build according to:
+
+	https://scan.coverity.com/download?tab=cxx
+
+E.g.:
+
+	../cov-analysis-linux64-2019.03/bin/cov-build --dir cov-int make
+
+
 Supported features
 ==================
 
--- a/README.md
+++ b/README.md
@@ -61,13 +61,10 @@ Build requirements:
  * `cmake` (v2 or above)
  * `apt install libjson-c-dev libcmocka-dev libssl-dev` *or*
  * `yum install json-c-devel libcmocka-devel openssl-devel`
- * `valgrind` for testing
 
 To build:
 
     make && make install
-    # optional
-    make test
 
 By default a debug build is created. To create a release build do:
 
@@ -292,6 +289,11 @@ backed by hugepages:
 
 Becasue SPDK must be run as root, either fix the vfio-user socket permissions
 or configure libvirt to run QEMU as root.
+
+Testing
+=======
+
+See [Testing](docs/testing.md).
 
 History
 =======
--- a/README.md
+++ b/README.md
@@ -74,18 +74,6 @@ The kernel headers are necessary because VFIO structs and defines are reused.
 
 Finally build your program and link with `libvfio-user.so`.
 
-
-Testing
-=======
-
-
-Debugging Test Errors
----------------------
-
-Sometimes debugging Valgrind errors on Python unit tests can be tricky. To
-run specific tests use the pytest `-k` option in `PYTESTCMD` in the Makefile.
-
-
 Coverity
 ========
 
--- a/include/libvfio-user.h
+++ b/include/libvfio-user.h
@@ -416,17 +416,17 @@ typedef void (vfu_dma_register_cb_t)(vfu_ctx_t *vfu_ctx, vfu_dma_info_t *info);
 
 /*
  * Function that is called when the guest unregisters a DMA region.
- * The callback should return 0 on success and -errno on failure.
+ * The callback should return 0 on success and -1 on failure setting errno.
  * (although unregister should not fail: this will not stop a guest from
  * unregistering the region).
  * This callback is required if you want to be able to access guest memory
  * directly via a mapping.
  *
  * The device must release all references to that region before the callback
- * returns. If the device returns -EBUSY then libvfio-user does not respond
- * to the client and does not process new messages until vfu_async_done is
- * called by the device, effectively making the DMA unmap operation
- * asynchronous.
+ * returns. If the device returns -1 and errno is set to EBUSY then
+ * libvfio-user does not respond to the client and does not process new
+ * messages until vfu_async_done is called by the device, effectively making
+ * the DMA unmap operation asynchronous.
  *
  * @vfu_ctx: the libvfio-user context
  * @info: the DMA info
--- a/include/libvfio-user.h
+++ b/include/libvfio-user.h
@@ -81,6 +81,10 @@ dma_sg_size(void);
  * Attaching to the transport is non-blocking.
  * The caller must then manually call vfu_attach_ctx(),
  * which is non-blocking, as many times as necessary.
+ *
+ * This also applies to vfu_run_ctx(). However, it's presumed that any actual
+ * reads or writes of the socket connection will not need to block, since both
+ * APIS are synchronous.
  */
 #define LIBVFIO_USER_FLAG_ATTACH_NB  (1 << 0)
 
--- a/include/libvfio-user.h
+++ b/include/libvfio-user.h
@@ -897,6 +897,25 @@ vfu_pci_find_next_capability(vfu_ctx_t *vfu_ctx, bool extended,
 bool
 vfu_sg_is_mappable(vfu_ctx_t *vfu_ctx, dma_sg_t *sg);
 
+/*
+ * Creates a new ioeventfd at the given setup memory region with @offset, @size,
+ * @fd, @flags and @datamatch.
+ *
+ * Returns 0 on success and -1 on failure with errno set.
+ *
+ * @vfu_ctx: the libvfio-user context
+ * @region_idx: The index of the memory region to set up the ioeventfd
+ * @fd: the value of the file descriptor
+ * @offset: The offset into the memory region
+ * @size: size of the ioeventfd
+ * @flags: Any flags to set up the ioeventfd
+ * @datamatch: sets the datamatch value
+ */
+int
+vfu_create_ioeventfd(vfu_ctx_t *vfu_ctx, uint32_t region_idx, int fd,
+                     size_t offset, uint32_t size, uint32_t flags,
+                     uint64_t datamatch);
+
 #ifdef __cplusplus
 }
 #endif
--- a/include/libvfio-user.h
+++ b/include/libvfio-user.h
@@ -409,27 +409,34 @@ typedef struct vfu_dma_info {
  * Called when a guest registers one of its DMA regions via a VFIO_USER_DMA_MAP
  * message.
  *
- * @vfu_ctx: the libvfio-user context
- * @info: the DMA info
- */
-typedef void (vfu_dma_register_cb_t)(vfu_ctx_t *vfu_ctx, vfu_dma_info_t *info);
-
-/*
- * Function that is called when the guest unregisters a DMA region.
- * The callback should return 0 on success and -1 on failure setting errno.
- * (although unregister should not fail: this will not stop a guest from
- * unregistering the region).
- * This callback is required if you want to be able to access guest memory
- * directly via a mapping.
- *
- * The device must release all references to that region before the callback
- * returns. If the device returns -1 and errno is set to EBUSY then
+ * If the callback returns -1 and errno is set to EBUSY then
  * libvfio-user does not respond to the client and does not process new
  * messages until vfu_async_done is called by the device, effectively making
- * the DMA unmap operation asynchronous.
+ * the DMA map operation asynchronous.
  *
  * @vfu_ctx: the libvfio-user context
  * @info: the DMA info
+ *
+ * @returns: 0 on success, -1 on failure with errno set.
+ */
+typedef int (vfu_dma_register_cb_t)(vfu_ctx_t *vfu_ctx, vfu_dma_info_t *info);
+
+/*
+ * Function that is called when the guest unregisters a DMA region.  The
+ * callback should return 0 on success and -1 on failure setting errno.
+ * (although unregister should not fail: this will not stop a guest from
+ * unregistering the region).  This callback is required if you want to be able
+ * to access guest memory directly via a mapping.
+ *
+ * The callback can be asynchronous, just like vfu_dma_register_cb_t.
+ *
+ * The device must release all references to that region before the callback
+ * returns.
+ *
+ * @vfu_ctx: the libvfio-user context
+ * @info: the DMA info
+
+ * @returns: 0 on success, -1 on failure with errno set.
  */
 typedef int (vfu_dma_unregister_cb_t)(vfu_ctx_t *vfu_ctx, vfu_dma_info_t *info);
 
--- a/include/libvfio-user.h
+++ b/include/libvfio-user.h
@@ -174,7 +174,7 @@ vfu_run_ctx(vfu_ctx_t *vfu_ctx);
  *
  * @vfu_ctx: the libvfio-user context to destroy
  */
-void
+int
 vfu_destroy_ctx(vfu_ctx_t *vfu_ctx);
 
 /**
@@ -342,6 +342,51 @@ typedef enum vfu_reset_type {
 } vfu_reset_type_t;
 
 /*
+ * Device callback for quiescing the device.
+ *
+ * livfio-user uses this callback to tell the device that it must quiesce its
+ * operation. After returning from this callback, the device can continue to
+ * operate however it cannot use any of the following libvfio-user functions:
+ * vfu_addr_to_sg, vfu_map_sg, vfu_unmap_sg, vfu_dma_read, and vfu_dma_write.
+ * The device must also be prepared for the following callbacks to be executed:
+ * vfu_dma_register_cb_t, vfu_unregister_cb_t, vfu_reset_cb_t, and transition.
+ * These callbacks only called after the device has been quiesced.
+ *
+ * If the device returns -1 with errno set to EBUSY, then the device is not
+ * quiesced until the device calls vfu_device_quiesced. During this time the
+ * device is allowed to use the aforementioned libvfio-user functions.
+ *
+ * @vfu_ctx: the libvfio-user context
+ *
+ * @returns: 0 on success, -1 on failure with errno set.
+ */
+typedef int (vfu_device_quiesce_cb_t)(vfu_ctx_t *vfu_ctx);
+
+/**
+ * Sets up the device quiesce callback.
+ *
+ * @vfu_ctx: the libvfio-user context
+ * @reset: device reset callback
+ */
+int
+vfu_setup_device_quiesce_cb(vfu_ctx_t *vfu_ctx,
+                            vfu_device_quiesce_cb_t *quiesce_cb);
+
+/*
+ * Called by the device to complete a pending quiesce operation. After the
+ * function returns the device is unquiesced.
+ *
+ * @vfu_ctx: the libvfio-user context
+ * @quiesce_errno: 0 for success or errno in case the device fails to quiesce,
+ *                 in which case the operation requiring the quiesce is failed
+ *                 and the device is reset.
+ *
+ * @returns 0 on success, or -1 on failure. Sets errno.
+ */
+int
+vfu_device_quiesced(vfu_ctx_t *vfu_ctx, int quiesce_errno);
+
+/*
  * Callback function that is called when the device must be reset.
  */
 typedef int (vfu_reset_cb_t)(vfu_ctx_t *vfu_ctx, vfu_reset_type_t type);
@@ -409,11 +454,6 @@ typedef struct vfu_dma_info {
  * Called when a guest registers one of its DMA regions via a VFIO_USER_DMA_MAP
  * message.
  *
- * If the callback returns -1 and errno is set to EBUSY then
- * libvfio-user does not respond to the client and does not process new
- * messages until vfu_async_done is called by the device, effectively making
- * the DMA map operation asynchronous.
- *
  * @vfu_ctx: the libvfio-user context
  * @info: the DMA info
  *
@@ -426,14 +466,8 @@ typedef int (vfu_dma_register_cb_t)(vfu_ctx_t *vfu_ctx, vfu_dma_info_t *info);
  * callback should return 0 on success and -1 on failure setting errno.
  * (although unregister should not fail: this will not stop a guest from
  * unregistering the region).  This callback is required if you want to be able
- * to access guest memory directly via a mapping.
- *
- * The callback can be asynchronous, just like vfu_dma_register_cb_t.
- *
- * If operating synchronously, the device must release all references to that
- * region before the callback returns. Otherwise, if operating asynchronously,
- * the device must release all references to that region before the device calls
- * vfu_async_done.
+ * to access guest memory directly via a mapping. The device must release all
+ * references to that region before the callback returns.
  *
  * @vfu_ctx: the libvfio-user context
  * @info: the DMA info
@@ -516,16 +550,6 @@ typedef struct {
      *
      * The callback should return -1 on error, setting errno.
      *
-     * When operating in non-blocking mode (LIBVFIO_USER_FLAG_ATTACH_NB was
-     * passed to vfu_create_ctx) and -1 is returned with errno set to EBUSY,
-     * transitioning to the new state becomes asynchronous: libvfio-user does
-     * not send a response to the client and does not process any new messages.
-     * Transitioning to the new device state is completed by calling
-     * vfu_async_done. This behavior can be beneficial for devices whose
-     * threading model does not allow blocking.
-     *
-     * The user must not call functions vfu_dma_read or vfu_dma_write, doing so
-     * results in undefined behavior.
      *
      * TODO rename to vfu_migration_state_transition_callback
      * FIXME maybe we should create a single callback and pass the state?
@@ -592,20 +616,6 @@ typedef struct {
     int (*data_written)(vfu_ctx_t *vfu_ctx, uint64_t count);
 
 } vfu_migration_callbacks_t;
-
-/*
- * Completes a pending migration state transition or DMA /mapunmap operation.
- * Calling this function when there is no pending migration state transition or
- * DMA map/unmap operation results in undefined behavior.
- *
- * @vfu_ctx: the libvfio-user context
- * @reply_errno: 0 for success or errno on error.
- *
- * @returns 0 on success, or -1 on failure. Sets errno.
- */
-int
-vfu_async_done(vfu_ctx_t *vfu_ctx, int reply_errno);
-
 
 #ifndef VFIO_DEVICE_STATE_STOP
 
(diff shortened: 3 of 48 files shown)


tmakatos:
Converted to draft as we're reworking dependee #608.

include/libvfio-user.h
```diff
@@ -415,14 +415,18 @@ typedef struct vfu_dma_info {
 typedef void (vfu_dma_register_cb_t)(vfu_ctx_t *vfu_ctx, vfu_dma_info_t *info);
```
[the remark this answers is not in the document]
tmakatos: right

lib/libvfio-user.c
```diff
@@ -817,8 +817,14 @@ handle_dma_unmap(vfu_ctx_t *vfu_ctx, vfu_msg_t *msg,
                                        vfu_ctx);
     if (ret < 0) {
         ret = errno;
+        if (errno == EBUSY) {
+            vfu_ctx->pending = VFU_CTX_PENDING_DMA_UNMAP;
+            vfu_ctx->pending_dma_unmap_addr = dma_unmap->addr;
```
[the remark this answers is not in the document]
tmakatos: Good point, that was when we didn't keep a copy of the original msg if the no_reply bit was set, I'll fix it.

lib/libvfio-user.c
```diff
@@ -1838,6 +1844,10 @@ vfu_map_sg(vfu_ctx_t *vfu_ctx, dma_sg_t *sg, struct iovec *iov, int cnt,
         return ERROR_INT(EINVAL);
     }
 
+    if (unlikely(vfu_ctx->pending == VFU_CTX_PENDING_DMA_UNMAP)) {
```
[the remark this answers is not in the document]
tmakatos: It's a simplification, otherwise dirty page tracking gets complicated. Currenlty, we generate the dirty pages bitmap before we call the callback, so if the user calls `vfu_sg_map` we'll miss dirty pages. Do you think it's worth fixing now?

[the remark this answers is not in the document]
jlevon: I think we can revisit in context of #594 and #598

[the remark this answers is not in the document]
jlevon: and maybe we should reconsider #595 there too

include/libvfio-user.h
```diff
@@ -409,23 +409,34 @@ typedef struct vfu_dma_info {
...
+ * Function that is called when the guest unregisters a DMA region.  The
+ * callback should return 0 on success and -1 on failure setting errno.
  * (although unregister should not fail: this will not stop a guest from
- * unregistering the region).
+ * unregistering the region).  This callback is required if you want to be able
+ * to access guest memory directly via a mapping.
+ *
+ * The callback can be asynchronous, just like vfu_dma_register_cb_t.
+ *
+ * The device must release all references to that region before the callback
```
jlevon, line 433: Should specify this is the case only if sync, and if async, then it must be true when calling vfu_async_done() ?

include/libvfio-user.h
```diff
@@ -581,15 +592,17 @@ typedef struct {
 } vfu_migration_callbacks_t;
 
 /*
- * Completes a pending migration state transition. Calling this function when
- * there is no pending migration state transition results in undefined
- * behavior.
+ * Completes a pending migration state transition or DMA unmap operation.
+ * Calling this function when there is no pending migration state transition or
+ * DMA unmap operation results in undefined behavior.
```
jlevon, line 597: map/unmap

lib/dma.c
```diff
@@ -168,12 +168,13 @@ MOCK_DEFINE(dma_controller_remove_region)(dma_controller_t *dma,
...
 
         err = dma_unregister == NULL ? 0 : dma_unregister(data, &region->info);
-        if (err != 0) {
-            err = errno;
-            vfu_log(dma->vfu_ctx, LOG_ERR,
-                   "failed to dma_unregister() DMA region [%p, %p): %m",
-                   region->info.iova.iov_base, iov_end(&region->info.iova));
-            return ERROR_INT(err);
+        if (err < 0) {
+            if (errno != EBUSY) {
```
jlevon, line 172: just had an awkward thought, what about remove_all_regions?

lib/libvfio-user.c
```diff
@@ -1976,19 +1993,30 @@ vfu_sg_is_mappable(vfu_ctx_t *vfu_ctx, dma_sg_t *sg)
...
 {
+    int ret = 0;
+
     assert(vfu_ctx != NULL);
-    assert(vfu_ctx->migr_trans_pending);
+    assert(vfu_ctx->pending != VFU_CTX_PENDING_NONE);
 
-    if (!vfu_ctx->migr_trans_msg->hdr.flags.no_reply) {
-        do_reply(vfu_ctx, vfu_ctx->migr_trans_msg, reply_errno);
+    if (vfu_ctx->pending == VFU_CTX_PENDING_DMA_UNMAP) {
```
jlevon, line 2004: what about map? isn't the point of async map that we don't touch the DMA region array until consumer tells us?

lib/libvfio-user.c
```diff
@@ -335,19 +336,11 @@ handle_region_access(vfu_ctx_t *vfu_ctx, vfu_msg_t *msg)
...
-         * because we don't have a use case yet, the only user of migration
-         * is SPDK and it operates in non-blocking mode. We don't know the
-         * implications of enabling this in blocking mode as we haven't looked
-         * at the details.
-         */
-        vfu_ctx->migr_trans_pending = true;
-        return 0;
-    }
     if (ret != in_ra->count) {
+        if (ret == -1 && errno == EBUSY
```
jlevon, line 340: one-line comment as to when this happens and clarify that the callback will have set pending state?

lib/libvfio-user.c
```diff
@@ -661,31 +654,13 @@ consume_fd(int *fds, size_t nr_fds, size_t index)
...
-    assert(msg != NULL);
-    assert(dma_map != NULL);
-
-    if (msg->in_size < sizeof(*dma_map) || dma_map->argsz < sizeof(*dma_map)) {
-        vfu_log(vfu_ctx, LOG_ERR, "bad DMA map region size=%zu argsz=%u",
-                msg->in_size, dma_map->argsz);
-        return ERROR_INT(EINVAL);
-    }
-
-    snprintf(rstr, sizeof(rstr), "[%#lx, %#lx) offset=%#lx flags=%#x",
```
jlevon, line 683: I think it's better to have this output at the point we're about to handle it, not the point that requested it. (see comment about quiesce below too)

lib/libvfio-user.c
```diff
@@ -713,51 +691,73 @@ handle_dma_map(vfu_ctx_t *vfu_ctx, vfu_msg_t *msg,
...
-    }
+    vfu_log(vfu_ctx, LOG_DEBUG,
+            "adding DMA region [%#lx, %#lx) offset=%#lx flags=%#x",
+            dma_map->addr, dma_map->addr + dma_map->size, dma_map->offset,
+            dma_map->flags);
 
-    snprintf(rstr, sizeof(rstr), "[%#lx, %#lx) flags=%#x",
-             dma_unmap->addr, dma_unmap->addr + dma_unmap->size, dma_unmap->flags);
+    if (vfu_ctx->quiesce != NULL) {
+        vfu_log(vfu_ctx, LOG_DEBUG, "quiescing device");
```
jlevon, line 737: given above comment, this should say *why* we want to quiesce "... due to DMA map request"

lib/libvfio-user.c
```diff
@@ -1976,19 +2083,71 @@ vfu_sg_is_mappable(vfu_ctx_t *vfu_ctx, dma_sg_t *sg)
...
-    vfu_ctx->migr_trans_pending = false;
+    vfu_log(vfu_ctx, LOG_DEBUG, "device quiesced with error=%d", quiesce_errno);
+
+    if (quiesce_errno == 0) {
+        switch (vfu_ctx->pending.state) {
+        case VFU_CTX_PENDING_MIGR:
+            ret = migr_trans_to_valid_state(vfu_ctx, vfu_ctx->migration,
+                                            vfu_ctx->pending.migr_dev_state, true);
+            break;
+        case VFU_CTX_PENDING_DMA_MAP:
```
jlevon, line 2109: Why not this:

```
 case VFU_CTX_PENDING_MSG:
    ret = exec_command(vfu_ctx, vfu_ctx.pending.msg);
    break;
```

and drop the map/unmap/device_reset_quiesced() variants altogether, along with .device_reset_reason. All you'd need to change would be don't call ->quiesce() when there's already ->pending.state (although I think I'd prefer an explicit "quiesced" state tracking anyway)

In fact, given my above point about FLR, I'm wondering - can we do this for *all* region access too. Instead of tracking sub-state for the region write (FLR or migration device state), peek ahead first:

```
 198 static ssize_t                                                                   
 199 region_access(vfu_ctx_t *vfu_ctx, size_t region_index, char *buf,                
 200               size_t count, uint64_t offset, bool is_write)                      
 201 {                                                                                

...
       if (access_needs_quiesce()) {
           ... call quiesce, handle EBUSY if needed ...
       }
...
```

which would check for the migration device state offset or the PCI FLR offset. I think this works out simpler.

lib/libvfio-user.c
```diff
@@ -713,51 +691,73 @@ handle_dma_map(vfu_ctx_t *vfu_ctx, vfu_msg_t *msg,
...
+            dma_map->flags);
 
-    snprintf(rstr, sizeof(rstr), "[%#lx, %#lx) flags=%#x",
-             dma_unmap->addr, dma_unmap->addr + dma_unmap->size, dma_unmap->flags);
+    if (vfu_ctx->quiesce != NULL) {
+        vfu_log(vfu_ctx, LOG_DEBUG, "quiescing device");
+        ret = vfu_ctx->quiesce(vfu_ctx);
+        if (ret < 0) {
+            if (errno == EBUSY) {
+                vfu_log(vfu_ctx, LOG_DEBUG, "device will quiesce asynchronously");
```
jlevon, line 741: why do we get debug for this but not the other quiesce users.

lib/libvfio-user.c
```diff
@@ -823,18 +823,61 @@ handle_dma_unmap(vfu_ctx_t *vfu_ctx, vfu_msg_t *msg,
                                        vfu_ctx->dma_unregister,
                                        vfu_ctx);
     if (ret < 0) {
-        ret = errno;
+        if (errno == EBUSY) {
```
jlevon, line 826: this shouldn't be here I think.

lib/libvfio-user.c
```diff
@@ -853,10 +896,29 @@ do_device_reset(vfu_ctx_t *vfu_ctx, vfu_reset_type_t reason)
...
+handle_device_reset_quiesced(vfu_ctx_t *vfu_ctx, vfu_reset_type_t reason)
+{
+    return do_device_reset(vfu_ctx, reason);
+}
+
 int
 handle_device_reset(vfu_ctx_t *vfu_ctx, vfu_reset_type_t reason)
 {
-    return do_device_reset(vfu_ctx, reason);
+    int ret;
```
jlevon, line 908: we need async FLR support too.

lib/libvfio-user.c
```diff
@@ -1976,19 +2083,71 @@ vfu_sg_is_mappable(vfu_ctx_t *vfu_ctx, dma_sg_t *sg)
...
+            break;
+        case VFU_CTX_PENDING_DMA_UNMAP:
+            ret = handle_dma_unmap_quiesced(vfu_ctx, vfu_ctx->pending.msg,
+                                            vfu_ctx->pending.msg->in_data);
+            break;
+        case VFU_CTX_PENDING_DEVICE_RESET:
+            ret = handle_device_reset_quiesced(vfu_ctx,
+                                               vfu_ctx->pending.device_reset_reason);
+            break;
+        case VFU_CTX_PENDING_CTX_RESET:
```
jlevon, line 2121: I think I now share your discomfort with vfu_reset_ctx() though. Will discuss on slack.

lib/libvfio-user.c
```diff
@@ -1976,19 +2083,71 @@ vfu_sg_is_mappable(vfu_ctx_t *vfu_ctx, dma_sg_t *sg)
...
+        default:
+            assert(false);
+        }
+    } else {
+        ret = -1;
+        errno = quiesce_errno;
+    }
+
+    /*
+     * We only reply when there is a message sent by the clinet to reply to
```
jlevon, line 2133: client

samples/client.c
```diff
@@ -1250,12 +1250,12 @@ int main(int argc, char *argv[])
     }
 
     /*
-     * Schedule an interrupt in 3 seconds from now in the old server and then
+     * Schedule an interrupt in 10 seconds from now in the old server and then
```
jlevon, line 1253: separate PR please?

Does this fix the constant CI hangs/timeouts?

samples/client.c
```diff
@@ -1250,12 +1250,12 @@ int main(int argc, char *argv[])
...
      * immediatelly migrate the device. The new server should deliver the
-     * interrupt. Hopefully 3 seconds should be enough for migration to finish.
+     * interrupt. Hopefully 10 seconds should be enough for migration to finish.
      * TODO make this value a command line option.
```
jlevon, line 1256: I think you can remove the TODO?

samples/gpio-pci-idio-16.c
```diff
@@ -128,12 +128,13 @@ migration_write_data(UNUSED vfu_ctx_t *vfu_ctx, void *buf,
 static void
 dma_register(UNUSED vfu_ctx_t *vfu_ctx, UNUSED vfu_dma_info_t *info)
 {
+    return;
```
jlevon, line 131: why? please drop

lib/libvfio-user.c
```diff
@@ -1454,15 +1513,48 @@ vfu_reset_ctx(vfu_ctx_t *vfu_ctx, const char *reason)
...
+static int
+vfu_reset_ctx(vfu_ctx_t *vfu_ctx, int reason)
+{
+    if (vfu_ctx->quiesce != NULL
+        && vfu_ctx->pending.state == VFU_CTX_PENDING_NONE) {
+        int ret = vfu_ctx->quiesce(vfu_ctx);
+        if (ret < 0) {
+            if (errno == EBUSY) {
+                vfu_ctx->pending.state = VFU_CTX_PENDING_CTX_RESET;
+                vfu_ctx->pending.ctx_reset_errno = reason;
```
jlevon, line 1525: seems a bit silly to stash the reason, *solely* to have something to print out on :1497 - can we print it out here instead?

Also,

include/libvfio-user.h
```diff
@@ -170,7 +170,8 @@ int
 vfu_run_ctx(vfu_ctx_t *vfu_ctx);
```
tmakatos, line 170: Need to document that `vfu_run_ctx` returns `EBUSY` while a quiesce is pending.

include/libvfio-user.h
```diff
@@ -341,6 +342,51 @@ typedef enum vfu_reset_type {
     VFU_RESET_PCI_FLR
 } vfu_reset_type_t;
 
+/*
+ * Device callback for quiescing the device.
+ *
+ * livfio-user uses this callback to tell the device that it must quiesce its
+ * operation. After returning from this callback, the device can continue to
+ * operate however it cannot use any of the following libvfio-user functions:
+ * vfu_addr_to_sg, vfu_map_sg, vfu_unmap_sg, vfu_dma_read, and vfu_dma_write.
```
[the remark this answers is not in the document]
tmakatos: I see your point.

The previous, main idea was that the device's quiesce callback is executed by `vfu_run_ctx`, and if it returns zero then `vfu_run_ctx` can proceed immediatelly. Therefore once `vfu_run_ctx` returns it doesn't require the device to be unquiesced, so the device knows it's no longer quiesced. Do you think this too complicated?

.github/workflows/pull_request.yml
```diff
@@ -2,7 +2,7 @@ name: pull_request
 on: pull_request
 jobs:
   ubuntu-20:
-    timeout-minutes: 5
+    timeout-minutes: 10
```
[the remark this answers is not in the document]
tmakatos: Because the `--read-inline-info=yes --read-var-info=yes --track-origins=yes` options added to Valgrind make it much slower.

README.md
```diff
@@ -77,6 +77,30 @@ The kernel headers are necessary because VFIO structs and defines are reused.
...
+
+
+Coverity
+========
+
+Build according to:
+
+	https://scan.coverity.com/download?tab=cxx
+
+E.g.:
```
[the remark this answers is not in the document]
tmakatos: ok

include/libvfio-user.h
```diff
@@ -170,7 +170,8 @@ int
 vfu_run_ctx(vfu_ctx_t *vfu_ctx);
 
 /**
- * Destroys libvfio-user context.
+ * Destroys libvfio-user context. During this call the device must be in
+ * quiesced state and the device callbacks can be called.
```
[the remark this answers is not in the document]
tmakatos: The quiesce callback is not called at all during destroy. I've reworded `vfu_destroy_ctx` for more clarity.

include/libvfio-user.h
```diff
@@ -341,6 +342,51 @@ typedef enum vfu_reset_type {
...
+ *
+ * @returns: 0 on success, -1 on failure with errno set.
+ */
+typedef int (vfu_device_quiesce_cb_t)(vfu_ctx_t *vfu_ctx);
+
+/**
+ * Sets up the device quiesce callback.
+ *
+ * @vfu_ctx: the libvfio-user context
+ * @reset: device reset callback
```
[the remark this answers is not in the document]
tmakatos: fixed

lib/libvfio-user.c
tmakatos: sure

lib/libvfio-user.c
tmakatos: OK, moved right before we call `dma_controller_add_region`.

lib/libvfio-user.c
tmakatos: sure

lib/libvfio-user.c
tmakatos: No good reason, I think I was debugging something. Shall I add this to all others? Or remove it from everywhere?

lib/libvfio-user.c
tmakatos: Sure we can print it here.

lib/libvfio-user.c
tmakatos: fixed

samples/client.c
tmakatos: Yes, OK will separate them.

samples/client.c
tmakatos: ok

lib/libvfio-user.c
jlevon: can we commonize at all?

lib/libvfio-user.c
jlevon: no - SPDK needs async reset, and this is one of the modes, so we need to do it

include/libvfio-user.h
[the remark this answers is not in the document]
jlevon: So I call vfu_run_ctx, and as part of that I get the quiesce callback. I'm OK to quiesce at this point, so I return 0 from that callback, and mark myself as quiesced.

Then I get back from vfu_run_ctx(). You're saying I should then check if I was quiesced, and if so, unquiesce myself?

This relies on vfu_run_ctx() returning EBUSY while not fully quiesced, right?

something like:

```
vfu_ctx_poller()


  ret = vfu_run_ctx();

  if (ret == EBUSY) {
    // quiescing still, don't do anything
    return IDLE;
  } else {
     if (ctrlr->quiesced) {
         vfio_user_unquiesce();
  }
```

is that what you're suggesting? I think that probably works with the comment you're going to add on EBUSY

lib/libvfio-user.c
tmakatos: I don't think I undertstand, we have two kinds of resets:

1. `handle_device_reset`, initiated by `VFIO_USER_DEVICE_RESET`, and
2. `vfu_reset_ctx`.

They can both be asynchronous. What am I missing?

jlevon: https://github.com/nutanix/libvfio-user/blob/master/lib/pci_caps.c#L286

include/libvfio-user.h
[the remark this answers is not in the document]
tmakatos: Yes, something like that, however note that `vfio_user_unquiesce` is _not_ a libvfio-user function, it's an internal SPDK function.
```
vfu_ctx_poller() {

  ret = vfu_run_ctx();

  if (ret == EBUSY) {
    // quiescing still, don't do anything
    return IDLE;
  } else {
     if (ctrlr->quiesced) {
         spdk_vfio_user_unquiesce();
         ctrlr->quiesced = false;
     }
  }
```

lib/libvfio-user.c
tmakatos: Note: FLR is initiated by writing to a PCI config space register (like migration state transition).

tmakatos: I've added a check in the relevant unit test that the quiesce callback must be called and this fails now, so we won't miss it.

lib/libvfio-user.c
tmakatos: The idea is to first check whether quiesce is required and if so quiesce the device and thern re-run `exec_command` once the device is quiesced.

tmakatos: @jlevon checking for the FLR offset isn't trivial as it's the relevant PCI capability can be anywhere past 0x40 in the PCI config space. The simplest way would be to call `cap_find_by_offset` and see whether a capability is returned and if it's the PCI Express capability then check whether it's FLR being written. This will make any PCI config space access a bit slower. Another way would be to store the offset of the PCI express capability in `vfu_ctx` so that we can easily look it up. This will, however, result in keeping redundant information that can be harder to maintain. Thoughts?

jlevon: Agreed, let's cache it

tmakatos: It just occured to me that if the user handles entirely the PCI config space then we won't know that relevant FLR register is written to so we won't tell the device to quiesce. One idea is to look for this capability in the PCI config space when `vfu_realize_ctx` is called. Otherwise we could just not handle this at all and leave it to the user.

jlevon: good point, let's file a ticket on that for now, and check with Ram later.

tmakatos: https://github.com/nutanix/libvfio-user/issues/629

test/py/libvfio_user.py
```diff
@@ -406,6 +424,7 @@ class vfio_user_region_io_fds_reply(Structure):
     ]
 
 
+# FIXME add a constructoer where argsz is set automatically
```
jlevon, line 427: remove fixme or clean up?

test/py/libvfio_user.py
```diff
@@ -595,18 +659,20 @@ def disconnect_client(ctx, sock):
...
-    assert c.get_errno() == errno.ENOTCONN
+    assert c.get_errno() == errno.ENOTCONN, os.strerror(c.get_errno())
 
 
 def get_reply(sock, expect=0):
     buf = sock.recv(4096)
     (msg_id, cmd, msg_size, flags, errno) = struct.unpack("HHIII", buf[0:16])
     assert (flags & VFIO_USER_F_TYPE_REPLY) != 0
-    assert errno == expect
+    assert errno == expect, "expect=%s, errno=%s" % (expect, errno)
```
jlevon, line 669: why? output already gives the values

test/py/libvfio_user.py
```diff
@@ -595,18 +659,20 @@ def disconnect_client(ctx, sock):
...
     return buf[16:]
 
 
-def msg(ctx, sock, cmd, payload, expect=0, fds=None, rsp=True):
+# TODO rename expect a it's confusing with expect_run_ctx_errno.
```
jlevon, line 673: fix the TODO?

test/py/libvfio_user.py
```diff
@@ -899,22 +1038,76 @@ def vfu_setup_device_dma(ctx, register_cb=None, unregister_cb=None):
                                                 vfu_dma_unregister_cb_t))
 
 
+# TODO group all callbacks in some class?
```
jlevon, line 1041: clean up?

jlevon (commented):
still most of test/py to review, but here's what I have so far, mostly nits. Thanks for this, it's much cleaner like this

test/py/test_dma_unmap.py
```diff
@@ -29,33 +29,45 @@
...
-    msg(ctx, sock, VFIO_USER_DMA_MAP, payload)
+def teardown_function(function):
+    global ctx, sock
+    disconnect_client(ctx, sock)
+    vfu_destroy_ctx(ctx)
 
 
-def test_dma_unmap_short_write():
+def with_dma_map(dma_maps=[(0x0, 0x1000)]):
+    """Function decorator that initializes the device as a PCI device."""
```
jlevon, line 53: why does this need to be a decorator?

test/py/test_migration.py
```diff
@@ -78,7 +58,39 @@ def stub():
...
 
-def test_migration_trans_sync():
+def teardown_function(function):
+    global ctx
+    vfu_destroy_ctx(ctx)
+
+
+@patch('libvfio_user.quiesce_cb')
+@patch('libvfio_user.migr_trans_cb')
+def test_migragion_bad_access(mock_trans, mock_quiesce):
```
jlevon, line 68: migration

test/py/test_pci_caps.py
```diff
@@ -27,80 +27,93 @@
...
+    ret = vfu_setup_device_quiesce_cb(ctx)
     assert ret == 0
 
 
+def teardown_function(function):
+    vfu_destroy_ctx(ctx)
+
+
+def with_pci(config_space=True, realize=False):
+    """Function decorator that initializes the device as a PCI device."""
```
jlevon, line 54: ditto, why a decorator

test/py/test_pci_caps.py
```diff
@@ -142,8 +157,19 @@ def test_add_caps():
...
+    __test_find_caps()
+
+    sock = connect_client(ctx)
+
+    __test_pci_cap_write_hdr(sock)
+    __test_pci_cap_readonly(sock)
+    __test_pci_cap_callback(sock)
+    __test_pci_cap_write_pmcs(sock)
+
+    # FIXME disconnect client?
```
jlevon, line 169: fix?

jlevon (commented):
actually, done now, thanks.

README.md
```diff
@@ -74,6 +74,23 @@ The kernel headers are necessary because VFIO structs and defines are reused.
 
 Finally build your program and link with `libvfio-user.so`.
 
+
+Testing
```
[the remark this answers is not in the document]
tmakatos: ok

include/libvfio-user.h
```diff
@@ -340,6 +343,59 @@ typedef enum vfu_reset_type {
...
+ *
+ * @returns: 0 on success, -1 on failure with errno set.
+ */
+typedef int (vfu_device_quiesce_cb_t)(vfu_ctx_t *vfu_ctx);
+
+/**
+ * Sets up the device quiesce callback.
+ *
+ * @vfu_ctx: the libvfio-user context
+ * @quiesce: device quiesce callback
```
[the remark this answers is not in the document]
tmakatos: ok

include/libvfio-user.h
```diff
@@ -340,6 +343,59 @@ typedef enum vfu_reset_type {
...
+ *      to execute and when vfu_run_ctx returns to the caller the device is
+ *      unquiesced.
+ * 2) -1 with errno set to EBUSY: this indicates that the device cannot
+ *      immediately quiesce. In this case, vfu_run_ctx returns -1 with errno
+ *      set to EBUSY and future calls to vfu_run_ctx return the same. Until the
+ *      device quiesces it can continue operate as normal. The device indicates
+ *      that it quiesced by calling vfu_device_quiesced. When
+ *      vfu_device_quiesced returns the device is no longer quiesced.
+ *
+ * A quiesced device should expected for any of the following callbacks to be
```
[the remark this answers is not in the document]
tmakatos: ok

lib/libvfio-user.c
```diff
@@ -195,6 +196,36 @@ dump_buffer(const char *prefix UNUSED, const char *buf UNUSED,
 #endif
 }
 
+static bool
```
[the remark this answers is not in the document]
tmakatos: ok

lib/libvfio-user.c
```diff
@@ -195,6 +196,36 @@ dump_buffer(const char *prefix UNUSED, const char *buf UNUSED,
...
+     * Writing to the migration state register with an unaligned access won't
+     * trigger this check but that's not a problem because
+     * migration_region_access_registers will fail the access.
+     */
+    return region_index == VFU_PCI_DEV_MIGR_REGION_IDX
+           && vfu_ctx->migration != NULL
+           && offset == offsetof(struct vfio_user_migration_info, device_state);
+}
+
+static bool
```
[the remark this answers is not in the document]
tmakatos: ok

lib/private.h
```diff
@@ -149,11 +149,26 @@ struct pci_dev {
...
+    VFU_CTX_PENDING_MSG,
+    VFU_CTX_PENDING_DEVICE_RESET,
+    VFU_CTX_PENDING_CTX_RESET
+};
+
+struct vfu_ctx_pending_info {
+    enum vfu_ctx_pending_state  state;
+    vfu_msg_t                   *msg;
+    /* when pending == VFU_CTX_PENDING_XXX_RESET */
+    union {
```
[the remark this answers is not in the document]
tmakatos: :D

test/py/libvfio_user.py
tmakatos: ok

test/py/libvfio_user.py
tmakatos: It doesn't for me:
```
    def test_dma_region_too_big():
        global ctx, sock

        payload = vfio_user_dma_map(argsz=len(vfio_user_dma_map()),
            flags=(VFIO_USER_F_DMA_REGION_READ |
                   VFIO_USER_F_DMA_REGION_WRITE),
            offset=0, addr=0x10000, size=MAX_DMA_SIZE + 4096)

>       msg(ctx, sock, VFIO_USER_DMA_MAP, payload, expect=123)

test/py/test_dma_map.py:65:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ test/py/libvfio_user.py:694: in msg
    return get_reply(sock, expect=expect)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
sock = <socket.socket fd=4, family=AddressFamily.AF_UNIX, type=SocketKind.SOCK_STREAM, proto=0, raddr=/tmp/vfio-user.sock.111900>, expect = 123

    def get_reply(sock, expect=0):
        buf = sock.recv(4096)
        (msg_id, cmd, msg_size, flags, errno) = struct.unpack("HHIII", buf[0:16])
        assert (flags & VFIO_USER_F_TYPE_REPLY) != 0
>       assert errno == expect
E       AssertionError

test/py/libvfio_user.py:671: AssertionError
```

jlevon: what's your version of pytest-3?

```
[john.levon@john-levon master]$ pytest-3 --version
This is pytest version 2.9.2, imported from /usr/lib/python3.6/site-packages/pytest.py
[john.levon@john-levon master]$ pytest-3  -rP --quiet test/py/test_dma_unmap.py 
...F.....
...

sock = <socket.socket fd=10, family=AddressFamily.AF_UNIX, type=SocketKind.SOCK_STREAM, proto=0, raddr=/tmp/vfio-user.sock.126582>, expect = 22

    def get_reply(sock, expect=0):
        buf = sock.recv(4096)
        (msg_id, cmd, msg_size, flags, errno) = struct.unpack("HHIII", buf[0:16])
        assert (flags & VFIO_USER_F_TYPE_REPLY) != 0
>       assert errno == expect
E       assert 95 == 22

test/py/libvfio_user.py:605: AssertionError
```

test/py/libvfio_user.py
tmakatos: ok

test/py/libvfio_user.py
tmakatos:
```
# pytest-3 --version
This is pytest version 3.10.1, imported from /usr/lib/python3/dist-packages/pytest.py
```

lib/libvfio-user.c
```diff
@@ -195,6 +196,36 @@ dump_buffer(const char *prefix UNUSED, const char *buf UNUSED,
...
+access_is_pci_cap_exp(const vfu_ctx_t *vfu_ctx, size_t region_index,
+                      uint64_t offset)
+{
+    return region_index == VFU_PCI_DEV_CFG_REGION_IDX
+           && offset == (size_t)vfu_ctx->pci_cap_exp_off + offsetof(struct pxcap, pxdc);
+}
+
+static bool
```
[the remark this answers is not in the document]
tmakatos: So far reading the migration state register doesn't require the device to be quiesced.

For FRL, why would the read have to be synchronous given that the reset callback is synchronous?

lib/libvfio-user.c
```diff
@@ -1156,6 +1178,32 @@ MOCK_DEFINE(should_exec_command)(vfu_ctx_t *vfu_ctx, uint16_t cmd)
...
 }
 
+static bool
+command_needs_quiesce(vfu_ctx_t *vfu_ctx, const vfu_msg_t *msg)
+{
+    if (vfu_ctx->quiesce == NULL) {
+        return false;
+    }
+
+    if (msg->hdr.cmd == VFIO_USER_DMA_MAP
```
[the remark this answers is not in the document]
tmakatos: ok

lib/libvfio-user.c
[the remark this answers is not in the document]
jlevon: I'd argue the other way around: why are you requiring a quiesce to read the FLR register?

test/py/libvfio_user.py
jlevon: I'd ask upstream somewhere, but I don't think we need code changes.

lib/libvfio-user.c
```diff
@@ -1255,53 +1306,68 @@ do_reply(vfu_ctx_t *vfu_ctx, vfu_msg_t *msg, int reply_errno)
  * possibly reply.
  */
 static int
-process_request(vfu_ctx_t *vfu_ctx)
+process_request(vfu_ctx_t *vfu_ctx, vfu_msg_t *msg)
```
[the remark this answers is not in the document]
tmakatos: Not sure I follow. I understand putting the body of the fist `if` statement into a function, however that's not perfect as there are places where we `return` and places where we `goto out`. Can you provide an outline of what you have in mind?

lib/libvfio-user.c
```diff
@@ -1761,6 +1850,14 @@ vfu_setup_device_reset_cb(vfu_ctx_t *vfu_ctx, vfu_reset_cb_t *reset)
     return 0;
 }
 
+EXPORT int
```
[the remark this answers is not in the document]
tmakatos: not for now

test/py/libvfio_user.py
tmakatos: Not sure it's worth fixing, I'm removing the TODO.

test/py/test_dma_unmap.py
tmakatos: It doesn't. It just keeps the "envrironmet" of the test separate from the body of the test, thus making reading the test easier (at least to me). Shall I make it a standard function?

test/py/test_migration.py
tmakatos: ok

test/py/test_pci_caps.py
tmakatos: See answer about `with_dma_map` decorator.

test/py/test_pci_caps.py
tmakatos: not worth it

test/py/test_pci_caps.py
```diff
@@ -305,49 +318,58 @@ def test_pci_cap_write_pmcs():
...
 def test_pci_cap_write_msix():
     # FIXME
     pass
 
 
+@with_pci(realize=True)
 def test_pci_cap_write_pxdc2():
     sock = connect_client(ctx)
+
+    # FIXME copied from test_pci_cap_write_px
```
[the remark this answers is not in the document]
tmakatos: ok

test/py/test_quiesce.py
```diff
@@ -0,0 +1,109 @@
...
+    ret = vfu_device_quiesced(ctx, 0)
+    assert ret == -1
+    assert c.get_errno() == errno.EINVAL
+    assert mock_quiesce.call_count == 0
+
+
+@patch('libvfio_user.quiesce_cb', side_effect=fail_with_errno(errno.ENOTTY))
+def test_device_quiesce_error(mock_quiesce):
+    """
+    Checks that if the device quiesce callbacks fails then the operation
```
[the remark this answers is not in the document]
tmakatos: ok

lib/libvfio-user.c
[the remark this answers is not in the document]
tmakatos: We only call `why are you requiring a quiesce to read the FLR register` if it's a write.

tmakatos (commented):
@jlevon I've addressed most of your comments, excluding the ongoing discussions.

lib/libvfio-user.c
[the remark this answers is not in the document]
jlevon: yep, I screwed up reading the code.

lib/libvfio-user.c
```diff
@@ -1156,6 +1178,32 @@ MOCK_DEFINE(should_exec_command)(vfu_ctx_t *vfu_ctx, uint16_t cmd)
...
+        || msg->hdr.cmd == VFIO_USER_DMA_UNMAP) {
+        return vfu_ctx->dma != NULL;
+    }
+
+    if (msg->hdr.cmd == VFIO_USER_DEVICE_RESET) {
+        return true;
+    }
+
+    if (msg->hdr.cmd == VFIO_USER_REGION_WRITE) {
+        struct vfio_user_region_access *reg = msg->in_data;
```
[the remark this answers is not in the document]
jlevon: just for example, did we check that the client actually provided `sizeof (struct vfio_user_region_access)` bytes in `->in_data` ?

lib/libvfio-user.c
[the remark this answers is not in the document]
jlevon: That complicates things, yeah. Let's merge this as is and maybe revisit.

test/py/test_dma_unmap.py
jlevon: yes please, unless it's going to wrap (prelude and postlude) it just obscures what's going IMO, and it's basically the same number of lines.

lib/pci_caps.c
```diff
@@ -843,4 +843,12 @@ vfu_pci_find_capability(vfu_ctx_t *vfu_ctx, bool extended, int cap_id)
     return vfu_pci_find_next_capability(vfu_ctx, extended, 0, cap_id);
 }
 
+bool
+access_is_pci_cap_exp(const vfu_ctx_t *vfu_ctx, size_t region_index,
+                      uint64_t offset)
+{
+    return region_index == VFU_PCI_DEV_CFG_REGION_IDX
+           && offset == (size_t)vfu_ctx->pci_cap_exp_off + offsetof(struct pxcap, pxdc);
```
jlevon, line 851: nit, line length?

lib/libvfio-user.c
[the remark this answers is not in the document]
tmakatos: Sorry, I mised the point of your original comment, I though it was an affirmation but now I realize it's a question. Will fix.

test/py/test_dma_unmap.py
tmakatos: ok

test/py/test_pci_caps.py
tmakatos: fixed

lib/pci_caps.c
tmakatos: ok

PR #611, opened 2021-10-19
fix coverity warnings

jlevon:


samples/client.c
```diff
@@ -1159,14 +1162,16 @@ int main(int argc, char *argv[])
      */
     nr_dma_regions = server_max_fds << 1;
 
-    if ((fp = tmpfile()) == NULL) {
-        err(EXIT_FAILURE, "failed to create DMA file");
+    if ((tmpfd = mkstemp(template)) == -1) {
```
swapnili, line 1165: Coverity will not complain about missing 'close(tmpfd)' at the end?

PR #612, opened 2021-10-20
add coverity scan support in github action

jlevon:


Makefile
```diff
@@ -135,6 +136,13 @@ pre-push: realclean
 	make pytest-valgrind
 	make install DESTDIR=tmp.install
 
+coverity: realclean
```
swapnili, line 139: As this makefile target will not work locally, it expects env variables (github secrets), I think instead of doing the scan here better is to do it in github workflow. This target wont be of use for developers of libvfio-user.

PR #619, opened 2021-11-11
Fix reply of VFIO_USER_DEVICE_GET_REGION_INFO

swapnili:
Set VFIO_REGION_INFO_FLAG_CAPS flag only if caps are part of the reply.


test/py/test_device_get_region_info.py
```diff
@@ -136,8 +136,7 @@ def test_device_get_region_info_small_argsz_caps():
     assert info.argsz == 80
     assert info.flags == (VFIO_REGION_INFO_FLAG_READ |
```
jlevon, line 137: can we get a comment here? i.e. there are caps, but they don't fit, so we don't expect the flag to be set.

jlevon:
is this the issue changpeng had?

swapnili:
> is this the issue changpeng had?

Yes I think so this should solve that

jlevon (approved):
thank you

PR #623, opened 2021-11-12
support AFL++ fuzzing

jlevon:


samples/server.c
```diff
@@ -461,7 +462,11 @@ int main(int argc, char *argv[])
         err(EXIT_FAILURE, "failed to register signal handler");
     }
 
-    vfu_ctx = vfu_create_ctx(VFU_TRANS_SOCK, argv[optind], 0, &server_data,
+    if (strcmp(argv[optind], "pipe") == 0) {
```
tmakatos, line 465: Why do we support the pipe transport on the server but not on the client?

tmakatos (commented):
Why can't the existing, UNIX domain socket transport be used for fuzzing?

I see some duplication between tran_sock and tran_pipe, could we avoid that by moving the tran_pipe stuff into tran_sock?

Or maybe implement a simple Python/bash script that relays messages receives over the UNIX domain socket to the pipe and vice versa?

samples/server.c
jlevon: afl is the client

docs/testing.md
```diff
@@ -15,3 +15,38 @@ The master branch is run through [Coverity](scan.coverity.com) when a new PR
...
+docker run -ti -v /path/to/libvfio-user:/src aflplusplus/aflplusplus
+```
+
+Set up and build:
+
+```
+apt update
+apt-get -y install libjson-c-dev libcmocka-dev clang valgrind python3-pytest debianutils flake8 libssl-dev cmake
+
+cd /src
```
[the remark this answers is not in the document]
jlevon: did you pass the right -v argument to docker?

jlevon:
> Why can't the existing UNIX domain socket transport be used for fuzzing?
> 

AFL expects to be able to start a process and write to it for fuzzing, so going through unix domain socket doesn't work (easily).

> I see some duplication between tran_sock and tran_pipe, could we avoid that by moving the tran_pipe stuff into tran_sock?

Yeah, at least negotiate has a ton of duplication, let me take a look at fixing this.

>

jlevon:
Thanos: I refactored to share the negotiate code. I think trying to go any further would actually complicate things more - I don't want to add too much more to tran API.

swapnili:
I do not like the fact that this code around VFU_TRANS_PIPE will be part of regular builds although used only for testing. 
Can we somehow extract it out and build only in case for AFL testing?

swapnili:
> I don't want this to regress (not build) as we make changes though. I could add a build option, and only set it for "make pre-push" (or if explicitly given) ? Would that be better?

Yes that would be better to build only in case if explicitly asked or pre-push. Thanks

jlevon:
@swapnili @tmakatos ok now?

lib/tran_pipe.c
```diff
@@ -0,0 +1,472 @@
...
+    tran_pipe_t *tp = NULL;
+    int ret = 0;
+
+    assert(vfu_ctx != NULL);
+
+    tp = calloc(1, sizeof(tran_pipe_t));
+
+    if (tp == NULL) {
+        ret = errno;
+        goto out;
```
swapnili, line 254: This goto is not required you can simply free() and return from here?

swapnili (approved):
Looks good with one minor suggestion. Thanks

lib/tran_pipe.c
```diff
@@ -250,14 +250,12 @@ tran_pipe_init(vfu_ctx_t *vfu_ctx)
     tp = calloc(1, sizeof(tran_pipe_t));
 
     if (tp == NULL) {
-        ret = errno;
-        goto out;
+        return -1;
```
swapnili, line 253: return ERROR_INT(errno)

lib/tran_sock.c
```diff
@@ -375,8 +375,7 @@ tran_sock_init(vfu_ctx_t *vfu_ctx)
     ts = calloc(1, sizeof(tran_sock_t));
 
     if (ts == NULL) {
-        ret = errno;
-        goto out;
+        return -1;
```
swapnili, line 378: ditto

lib/tran_pipe.c
jlevon: why? What's here is the exact same.

swapnili: Just to be consistent with other returns. But I am fine with this also.

PR #628, opened 2021-11-22
check ->argsz for VFIO_USER_DIRTY_PAGES

jlevon:


lib/libvfio-user.c
```diff
@@ -894,18 +894,19 @@ handle_dirty_pages_get(vfu_ctx_t *vfu_ctx, vfu_msg_t *msg)
...
                 msg->in_size, dirty_pages_in->argsz);
         return ERROR_INT(EINVAL);
     }
 
     range_in = msg->in_data + sizeof(*dirty_pages_in);
 
-    /* NB: this is bound by MAX_DMA_SIZE. */
-    argsz = sizeof(*dirty_pages_out) + sizeof(*range_out) +
-            range_in->bitmap.size;
+    argsz = satadd_u64(sizeof(*dirty_pages_out) + sizeof(*range_out),
```
tmakatos, line 907: So if there's overflow we'll silently truncate `argsz`? Shouldn't we just fail?

jlevon: we'll fail later on when the dma controller checks bitmap.size but we don't have that info here. So we just need something safe against overflow.

tmakatos: Is this what the `MAX_DMA_SIZE` comment is about? It's a bit unclear.

jlevon: It was, but that comment was misleading. Let me fix

tmakatos (approved):

PR #642, opened 2022-01-28
ignore writes to RO MSI-X registers

tmakatos:

### DIFF
--- a/lib/pci_caps.c
+++ b/lib/pci_caps.c
@@ -171,7 +171,7 @@ cap_write_msix(vfu_ctx_t *vfu_ctx, struct pci_cap *cap, char *buf,
                size_t count, loff_t offset)
 {
     struct msixcap *msix = cap_data(vfu_ctx, cap);
-    struct msixcap new_msix = { { 0 } };
+    struct msixcap new_msix = *msix;
 
     memcpy((char *)&new_msix + offset - cap->off, buf, count);
 
--- a/lib/pci_caps.c
+++ b/lib/pci_caps.c
@@ -167,49 +167,36 @@ cap_write_pm(vfu_ctx_t *vfu_ctx, struct pci_cap *cap, char * buf,
 }
 
 static ssize_t
-handle_mxc_write(vfu_ctx_t *vfu_ctx, struct msixcap *msix,
-                 const struct mxc *const mxc)
+cap_write_msix(vfu_ctx_t *vfu_ctx, struct pci_cap *cap, char *buf,
+               size_t count, loff_t offset)
 {
-    assert(msix != NULL);
-    assert(mxc != NULL);
+    struct msixcap *msix = cap_data(vfu_ctx, cap);
+    struct msixcap new_msix = { { 0 } };
 
-    if (mxc->mxe != msix->mxc.mxe) {
-        vfu_log(vfu_ctx, LOG_DEBUG, "%s MSI-X",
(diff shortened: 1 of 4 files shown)


lib/pci_caps.c
```diff
@@ -167,49 +167,36 @@ cap_write_pm(vfu_ctx_t *vfu_ctx, struct pci_cap *cap, char * buf,
...
-    assert(mxc != NULL);
+    struct msixcap *msix = cap_data(vfu_ctx, cap);
+    struct msixcap new_msix = { { 0 } };
 
-    if (mxc->mxe != msix->mxc.mxe) {
-        vfu_log(vfu_ctx, LOG_DEBUG, "%s MSI-X",
-                mxc->mxe ? "enable" : "disable");
-        msix->mxc.mxe = mxc->mxe;
-    }
+    memcpy((char *)&new_msix + offset - cap->off, buf, count);
```
jlevon, line 176: don't we need to start with the old msix here, then copy over the new data?
what if we write to a byte that doesn't affect fm/mxe - won't that accidentally see them as zero?

tmakatos: IIUC the only parts of the MSI-X capability structure that can be modified are the function mask and enable bits, the rest must ignored. In L190 and L196 we only store those bits.

jlevon: If the client does a single byte write at offset 0, that should indeed be ignored.
But then new_msix.mxc.fm and new_msix.mxe are zero, so you'll disable them incorrectly.

can add a test for this too.

test/py/test_pci_caps.py
```diff
@@ -349,9 +349,55 @@ def test_pci_cap_write_px(mock_quiesce, mock_reset):
...
+
 def test_pci_cap_write_msix():
-    # FIXME
-    pass
+    setup_pci_dev(realize=True)
+    sock = connect_client(ctx)
+
+    flags = PCI_MSIX_FLAGS_MASKALL | PCI_MSIX_FLAGS_ENABLE
+    pos = vfu_pci_add_capability(ctx, pos=0, flags=0,
+              data=struct.pack("ccHII", to_byte(PCI_CAP_ID_MSIX), b'\0', 0, 0,
```
jlevon, line 362: nit, odd indentation?

test/py/test_pci_caps.py
```diff
@@ -349,9 +349,55 @@ def test_pci_cap_write_px(mock_quiesce, mock_reset):
...
+    data = b'\x00\x00'
+    write_region(ctx, sock, VFU_PCI_DEV_CFG_REGION_IDX,
+                 offset=offset + PCI_MSIX_FLAGS,
+                 count=len(data), data=data)
+    expected = to_bytes_le(PCI_CAP_ID_MSIX) + b'\x00\x00'
+    payload = read_region(ctx, sock, VFU_PCI_DEV_CFG_REGION_IDX, offset=offset,
+                          count=len(expected))
+    assert expected == payload
+
+    # write 2 bytes to Message Control + 1: mask all vectors and enable MSI-X
```
jlevon, line 390: worth mentioning that some versions of qemu do this?

tmakatos: sure

lib/pci_caps.c
tmakatos: ah you're right

jlevon (approved):
thanks

PR #647, opened 2022-02-24
check for allowed operations in quiesce state 

tmakatos:


tmakatos (commented):
@jlevon I implemented this to ensure SPDK wasn't doing the wrong thing (it wasn't), as we don't seem to have a way to detect if this happens, at least in debug builds. It's a bit cumbersome the way it's implemented now, if you think it's useful then we could introduce a function through which we'execute callbacks (the callback will be passed as an argument to that function) so that we'll only have to do the `in_cb = true`; ... `in_cb = false`; in that single place.

lib/libvfio-user.c
```diff
@@ -1426,6 +1453,12 @@ vfu_run_ctx(vfu_ctx_t *vfu_ctx)
             err = handle_request(vfu_ctx, msg);
             free_msg(vfu_ctx, msg);
             reqs_processed++;
+#ifdef DEBUG
```
jlevon, line 1456: why isn't this at the end of vfu_device_quiesced() ?

include/libvfio-user.h
```diff
@@ -347,7 +347,7 @@ typedef enum vfu_reset_type {
  * Device callback for quiescing the device.
  *
  * vfu_run_ctx uses this callback to request from the device to quiesce its
- * operation. A quiesced device must call the following functions:
+ * operation. A quiesced device must not call the following functions:
```
jlevon, line 350: separate PR for the comment changes?

lib/dma.c
```diff
@@ -167,7 +167,13 @@ MOCK_DEFINE(dma_controller_remove_region)(dma_controller_t *dma,
         }
 
         if (dma_unregister != NULL) {
+#ifdef DEBUG
```
jlevon, line 170: seems useful, but:

 - no reason to be debug only
 - we could track for debugging purposes which cb we're in (make it an enum)
 - let's add an inline helper in private.h or something

lib/dma.c
```diff
@@ -542,7 +556,27 @@ dma_controller_dirty_page_logging_stop(dma_controller_t *dma)
         dma->regions[i].dirty_bitmap = NULL;
     }
     dma->dirty_pgsize = 0;
+
+    vfu_log(dma->vfu_ctx, LOG_DEBUG, "dirty pages: stopped logging");
+}
+
+
+#ifdef DEBUG
```
jlevon, line 564: separate PR?

lib/libvfio-user.c
```diff
@@ -1963,6 +2002,10 @@ vfu_addr_to_sg(vfu_ctx_t *vfu_ctx, vfu_dma_addr_t dma_addr,
         return ERROR_INT(EINVAL);
     }
 
+#ifdef DEBUG
```
jlevon, line 2005: actually, this is only a bug depending on the client: they might have exclusion separately, or just be single-threaded.

We should also check if there's a quiesce callback set.

include/libvfio-user.h
tmakatos: right, that was accidental

lib/dma.c
tmakatos: right

lib/dma.c
tmakatos: > no reason to be debug only
We document this, do we need to police the user of the library for breaking contract?

> we could track for debugging purposes which cb we're in (make it an enum)
If we keep this in debug, the asserts will produce a stack trace so it should be clear which cb we're in.

> let's add an inline helper in private.h or something
Not sure I understand, are you referring to this?
>  if you think it's useful then we could introduce a function through which we'execute callbacks (the callback will be passed as an argument to that function)

lib/libvfio-user.c
tmakatos: sure

lib/dma.c
jlevon: > asserts will produce a stack trace

Only showing vfu_map_sg() etc. It won't show other situations where we are in callbacks. I'm saying, since you're tracking that user is in a callback, why not add a little extra  info so we know which one?

jlevon: >inline helper
I'm saying instead of open coding the assert, but maybe that's not really necessary

lib/libvfio-user.c
tmakatos: Hang on, it _is_ at the end of vfu_device_quiesced(), but that's not enough because the device might immediatelly quiesce as you say.

lib/libvfio-user.c
tmakatos: > We should also check if there's a quiesce callback set.

good point

> actually, this is only a bug depending on the client: they might have exclusion separately, or just be single-threaded.

I don't follow, can you elaborate?

lib/libvfio-user.c
jlevon: this is vfu_run_ctx() !

lib/libvfio-user.c
jlevon: If the client is single threaded, then by definition, it can't be executing vfu_map_sg() or friends at the same time as handling a dma unregister. So it's safe. Also true if it happens to take its own pthread lock around those. But, presumably, if they have a quiesce cb, then neither of those are true, and it makes sense to double check them.

lib/libvfio-user.c
tmakatos: I mean `vfu_ctx->quiesced = false` is already at the end of `vfu_device_quiesced()`.

lib/dma.c
tmakatos: > Only showing vfu_map_sg() etc. It won't show other situations where we are in callbacks.

good point, we do need the extra info you suggest

lib/libvfio-user.c
jlevon: sure, but that's fine, no? Doesn't harm anything to immediately mark it unquiesced there

lib/dma.c
tmakatos: Just realized I misunderstood something:
> we could track for debugging purposes which cb we're in (make it an enum)

These functions are allowed to be called even in quiesced state provided that they're called by a callback, so why track which callback we're in? If the call to e.g. `vfu_map_sg()` is not allowed, then it's guaranteed that it won't be call by a callback. Do you want the callback type printed each type we call these type of functions?

lib/libvfio-user.c
tmakatos: Right, makes sense. Is there something you want done about this, e.g. a comment in the code?

lib/libvfio-user.c
jlevon: I'm asking for some comments on this, as it's not immediately obvious why we need the check in vfu_run_ctx() too. At least I - clearly - always forget this

lib/dma.c
jlevon: From this:

```
  vfu_ctx->in_cb = true;
```

to this:

```
  vfu_ctx->in_cb = VFU_CTX_CB_DMA_UNREGISTER;
```

tmakatos: I get that, my point is why track the callback type? How are we going to use it? Accessing function like `vfu_map_sg()` and friends is allowed in done by a callback.

jlevon: just for debugging purposes if we have a core. it's no big deal.

lib/libvfio-user.c
jlevon: Yes, I'm suggesting you only assert if there's a quiesce callback registered

lib/libvfio-user.c
tmakatos: I've added a comment here: https://github.com/nutanix/libvfio-user/pull/647/commits/f7d2096ad94438a391244a056e7a8fccdf8cfa8a

lib/libvfio-user.c
tmakatos: oh sure

lib/dma.c
```diff
@@ -167,7 +167,13 @@ MOCK_DEFINE(dma_controller_remove_region)(dma_controller_t *dma,
         }
 
         if (dma_unregister != NULL) {
+#ifdef DEBUG
+            dma->vfu_ctx->in_cb = CB_DMA_UNREGISTER;
+#endif
             dma_unregister(data, &region->info);
+#ifdef DEBUG
```
jlevon, line 174: still think you can drop the DEBUG here

lib/dma.c
```diff
@@ -203,7 +209,13 @@ dma_controller_remove_all_regions(dma_controller_t *dma,
                 region->info.mapping.iov_base, iov_end(&region->info.mapping));
 
         if (dma_unregister != NULL) {
+#ifdef DEBUG
+            dma->vfu_ctx->in_cb = true;
```
jlevon, line 213: CB_DMA_UNREGISTER?

lib/dma.c
```diff
@@ -203,7 +209,13 @@ dma_controller_remove_all_regions(dma_controller_t *dma,
...
+#endif
             dma_unregister(data, &region->info);
+#ifdef DEBUG
+            dma->vfu_ctx->in_cb = false;
```
jlevon, line 217: CB_NONE

lib/libvfio-user.c
```diff
@@ -1432,6 +1454,17 @@ vfu_run_ctx(vfu_ctx_t *vfu_ctx)
             err = handle_request(vfu_ctx, msg);
             free_msg(vfu_ctx, msg);
             reqs_processed++;
+#ifdef DEBUG
+            /*
+             * get_request might call the quiesce callback which might
+             * immediatelly quiesce the device, vfu_device_quiesced won't
```
jlevon, line 1460: immediately

lib/libvfio-user.c
```diff
@@ -2128,6 +2187,9 @@ vfu_device_quiesced(vfu_ctx_t *vfu_ctx, int quiesce_errno)
     }
 
     vfu_log(vfu_ctx, LOG_DEBUG, "device quiesced with error=%d", quiesce_errno);
+#ifdef DEBUG
```
jlevon, line 2190: ditto

lib/migration.c
```diff
@@ -155,8 +155,16 @@ MOCK_DEFINE(migr_trans_to_valid_state)(vfu_ctx_t *vfu_ctx, struct migration *mig
 {
     if (notify) {
         int ret;
+#ifdef DEBUG
+        assert(!vfu_ctx->in_cb);
+        vfu_ctx->in_cb = true;
```
jlevon, line 160: fix up type?

lib/migration.c
```diff
@@ -155,8 +155,16 @@ MOCK_DEFINE(migr_trans_to_valid_state)(vfu_ctx_t *vfu_ctx, struct migration *mig
...
+#endif
         ret = state_trans_notify(vfu_ctx, migr->callbacks.transition,
                                  device_state);
+#ifdef DEBUG
+        vfu_ctx->in_cb = false;
```
jlevon, line 165: ditto

lib/dma.c
tmakatos: right

lib/dma.c
tmakatos: right

lib/libvfio-user.c
tmakatos: ok

lib/migration.c
tmakatos: right

lib/migration.c
tmakatos: right

lib/dma.c
tmakatos: Just to be clear, you want the #ifdef dropped from everywhere we track the callback type, right? If so, what do we do with `quiesce_check_allowed`? Do we still abort even in non-debug mode?

jlevon: no, I think that makes sense to be DEBUG only still?

jlevon (approved):
thanks

PR #658, opened 2022-04-04
use struct iovec for grouping buffer and length

tmakatos:
This make it tidier and easier to pass to function the buffer and
length, instead of passing the whole msg.

### DIFF
--- a/lib/private.h
+++ b/lib/private.h
@@ -60,10 +60,10 @@
 /*
  * Structure used to hold an in-flight request+reply.
  *
- * Incoming request body and fds are stored in in_*.
+ * Incoming request body and fds are stored in in.*.
  *
- * Outgoing requests are either stored in out_data, or out_iovecs. In the latter
- * case, the iovecs refer to data that should not be freed.
+ * Outgoing requests are either stored in out.iov.iov_base, or out_iovecs. In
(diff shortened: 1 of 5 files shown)


jlevon:
This change (and your suggested TODO) look like a nice cleanup on their own, but I don't like the idea of "only passing the iovecs" at all. Introducing the msg struct made the request handling code *much* easier to understand. Why do you want to pass just iovecs around?

tmakatos:
> I don't like the idea of "only passing the iovecs" at all. Introducing the msg struct made the request handling code _much_ easier to understand. Why do you want to pass just iovecs around?

The reason I suggested this is because I've added a new function in migration v2 where I only need the out iovec, and I also checked that in other, existing places we only use a subset of the message. I didn't think of it from the angle you state, so we can leave as it is.

lib/private.h
```diff
@@ -60,28 +60,22 @@
...
-    int *out_fds;
-    size_t nr_out_fds;
-
-    void *out_data;
-    size_t out_size;
+    struct {
+        int *fds;
+        size_t nr_fds;
+        struct iovec iov;
+    } in, out;
```
jlevon, line 78: MIND BLOWN. I had no idea this was allowed!

jlevon (approved):
LGTM thanks

PR #665, opened 2022-04-21
test library linkage

jlevon:
Add a cheesy test for identifying functions in the public header that
are not exported.


jlevon:
cc @berrange

berrange:
Ah, clever approach. I see it correctly reports missing dma_sg_size method export

PR #666, opened 2022-05-05
Replace CMake with Meson as the build system

berrange:
This proposal switches the build system to use Meson (https://mesonbuild.com/) , which is what QEMU, libvirt & various other virt related projects have switched to in recent time. 

I updated docs in the series, but here are some usage illustrations.

Normal build + test

```
meson build --prefix $HOME/usr/vfio-user
meson compile -C build   (can be skipped as done implicitly by the following 'test' and 'install' commands)
meson test -C build
meson install -C build
```

Tests run without valgrind by default, to enable use of valgrind

```
meson build -Dvalgrind=true
meson test -C build  (run tests both directly and under valgrind)
meson test -C build --suite=valgrind   (run tests only under valgrind)
meson test -C build --no-suite=valgrind  (run tests only directly)
```

Tests with ASAN

```
meson build -Db_sanitize=address
meson test -C build
```

Code coverage

```
meson build -Db_coverage=true
meson test -C build
ninja -C build coverage
```

The 'make pre-pull' command is replaced by a script `.github/workflows/pull_request.sh`, likewise for coverity.

Note on RHEL-7 there is no `meson` but there is `ninja-build` at least. Thus we must get meson using `pip install meson`. This can also be done if the OS has a meson older than the declared min 0.56.0


jlevon:
Thanks for this series! I'll review closely soon but
 - it's not clear how we do debug versus release builds now, and the pre-push replacement seems to lost building both ways. I presume we just need to pass -Dbuildtype as well -Ddebug_logs
 - I'd be inclined to leave a small Makefile around for convenience, but that's something I can add back in after, for sure.

berrange:
> Thanks for this series! I'll review closely soon but
> 
>     * it's not clear how we do debug versus release builds now, and the pre-push replacement seems to lost building both ways. I presume we just need to pass -Dbuildtype as well -Ddebug_logs

Yes, that should do the trick. If I lost this in the pull_request.sh script that's a mistake I'll lookat fixing.

>     * I'd be inclined to leave a small Makefile around for convenience, but that's something I can add back in after, for sure.

I can respin with that if you say which kind of targets you'd like to see - was it just the 'pre-push' and 'coverity' targets, or something more comprehensive ?


And I've clearly got some debugging todo on the pull request CI jobs....

jlevon:
I personally use "pre-push", "coverity" and also just a "make" that would correspond to -Ddebug_logs and a non-opt build. Thanks!

berrange:
> Also, do you know how awful it would be to not have to require a newer meson? Not a show-stopper but if centos7's was at all workable it would be nice

I thought it was hard, but actually now I look I found 0.55 is in EPEL7 after all. This has changed since we first dealt with this in libvirt/qemu.

FWIW, both libvirt and QEMU upstream have explicitly dropped support for RHEL-7 vintage distro entirely.

jlevon:
> 0.55 is in EPEL7

Yes that would work

> explicitly dropped support for RHEL-7 vintage distro entirely

We are very aware of this :)

lib/meson.build
```diff
@@ -0,0 +1,46 @@
+
+lib_include_dir = [include_directories('.')]
+
+libvfio_user_cflags = cc.get_supported_arguments([
+  '-fvisibility=hidden',
+])
```
eli-schwartz, line 6: Please consider passing the `gnu_symbol_visibility: 'hidden'` kwarg to the library instead. It's available since Meson 0.48.0.

eli-schwartz (changes requested):

CMakeLists.txt
```diff
@@ -1,80 +0,0 @@
...
-include_directories(${CMAKE_SOURCE_DIR}/include)
-include_directories(${CMAKE_SOURCE_DIR}/lib)
-include_directories(${JSON_INCLUDE_DIRS})
-
-set(CMAKE_C_FLAGS_DEBUG "-O0 -ggdb")
-#
-# NB: this intentionally does not define NDEBUG: we want our assert()s even in a
-# release build.
-#
-set(CMAKE_C_FLAGS_RELEASE "-O3")
```
jlevon, line 57: it looks like the non-NDEBUG thing happens by default with meson?

docs/meson.build
```diff
@@ -0,0 +1,10 @@
+
+if rstlint.found()
+  test(
```
jlevon, line 3: nit, here, generally, is 2-space-indent a meson standard? if not could we get 4-spaces to match local style?

docs/testing.md
```diff
@@ -1,26 +1,66 @@
...
-`valgrind` installed.
+The tests in libvfio-user are organized into a number of suites
 
-Running `make pre-push` runs the above builds and tests in different configurations: GCC,
-clang, and with ASAN enabled.
+* native - old tests that use native binaries written in C
+* python - modern tests written in python
+* valgrind - tests that can run under valgrind (only enabled
+             if `-Dvalgrind=true` to be passed when `meson`
+	     was first invoked)
```
jlevon, line 10: should this mention "style" suite?

docs/testing.md
```diff
@@ -1,26 +1,66 @@
...
+To run with ASAN enabled, pass the `-Db_sanitize=address`
+option to meson. Note, this is incompatible with enabling
+valgrind
+
+```
+meson build -Db_sanitize=address
+meson test -C build
+```
+
+The `.github/workflows/pull_request.sh` script run a
```
jlevon, line 36: nit, runs

docs/testing.md
```diff
@@ -38,11 +78,14 @@ Set up and build:
 
 ```
 apt update
-apt-get -y install libjson-c-dev libcmocka-dev clang valgrind python3-pytest debianutils flake8 libssl-dev cmake
+apt-get -y install libjson-c-dev libcmocka-dev clang valgrind \
+                   python3-pytest debianutils flake8 libssl-dev\
```
jlevon, line 82: nit, missing space before backslash?

docs/testing.md
```diff
@@ -1,26 +1,66 @@
...
+request.
 
 The master branch is run through [Coverity](scan.coverity.com) when a new PR
 lands.
 
-You can also run `make gcov` to get code coverage reports.
+Coverage reports can be enabled via meson
+
+```
+meson build -Db_coverage=true
```
jlevon, line 48: I had to install gcovr (OK), but I can't find out how to get the detailed annotated output with meson? It only gives me summaries.

lib/meson.build
```diff
@@ -0,0 +1,46 @@
...
+  json_c_dep,
+]
+
+libvfio_user = library(
+  'vfio-user',
+  sources: libvfio_user_sources,
+  c_args: libvfio_user_cflags + common_cflags,
+  dependencies: libvfio_user_deps,
+  include_directories: public_include_dir,
+  soversion: 0,
```
jlevon, line 34: could we please re-instate the old comment that we're not yet providing ABI so this doesn't get changed?

meson.build
```diff
@@ -0,0 +1,76 @@
+project(
+  'libvfio-user',
+  'c',
+  version: '0.1',
```
jlevon, line 4: should this maybe be 0.0.1?

meson.build
```diff
@@ -0,0 +1,76 @@
...
+flake8 = find_program('flake8', required: false)
+rstlint = find_program('restructuredtext-lint', required: false)
+valgrind = find_program('valgrind', required: opt_valgrind)
+
+
+common_cflags = [
+  '-D_GNU_SOURCE',
+]
+
+if opt_debug_logs
```
jlevon, line 59: this should get set by default unless we're in release mode I think?

test/meson.build
```diff
@@ -0,0 +1,106 @@
...
+  'unit_tests',
+  unit_tests,
+  suite: 'native',
+)
+
+if opt_valgrind
+  test(
+    'valgrind-unit_tests',
+    valgrind,
+    is_parallel : false,
```
jlevon, line 55: I'm not clear why we need this?

include/meson.build
```diff
@@ -0,0 +1,15 @@
...
+
+libvfio_user_includes = files(
+  'libvfio-user.h',
+  'pci_defs.h',
+  'vfio-user.h'
+)
+
+install_headers(
+  libvfio_user_includes,
+  install_dir: includedir / 'vfio-user'
```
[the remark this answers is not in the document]
berrange: Oh, I didn't realize that opt existed.

lib/meson.build
berrange: Yep, that works and is more portable

lib/meson.build
berrange: Sure, will so

meson.build
berrange: Sure, whatever you want the versioning number to say.

test/meson.build
berrange: valgrind is pretty slow when the tests ran it parallel I found the tests were taking long enough that they hit the test timeout. I guess I coud just increase the test timeout instead actually, as that'll probably still be faster overall.

test/meson.build
```diff
@@ -0,0 +1,106 @@
...
+    suite: 'valgrind',
+    args: valgrind_native_args + [unit_tests],
+  )
+endif
```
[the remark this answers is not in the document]
berrange: Oh that's an interesting feature I didn't know about. It certainly looks like a better  approach to this problem

.github/workflows/pull_request.sh
```diff
@@ -0,0 +1,41 @@
...
+
+set -e
+set -v
+
+test -z "$BUILD" && BUILD=build
+
+rm -rf $BUILD
+
+# ASAN build with clang
+CC=clang meson $BUILD/asan -Db_sanitize=address -Db_lundef=false
```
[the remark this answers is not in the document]
berrange: Its the ninja command that needs the '-v', but yes, I'll add it

.github/workflows/pull_request.sh
```diff
@@ -0,0 +1,41 @@
...
+ninja -C $BUILD/clang-plain
+meson test -C $BUILD/clang-plain --no-suite style --print-errorlogs
+
+# debug build with gcc
+CC=gcc meson build/gcc-debug -Dtran-pipe=true
+ninja -C $BUILD/gcc-debug
+meson test -C $BUILD/gcc-debug --no-suite style --print-errorlogs
+
+# plain build with gcc
+CC=gcc meson build/gcc-plain -Dtran-pipe=true
```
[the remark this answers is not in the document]
berrange: No, its a mistake - should have passed the build type option

docs/meson.build
berrange: I don't think meson itself has a documented standard, but 2-space indent is what the Emacs meson-mode gives me

jlevon: ok, do you mind following local style (4 spaces) then please

test/meson.build
jlevon: So that's not related to meson right?

test/meson.build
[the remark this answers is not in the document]
berrange: This add_test_setup approach indeed works very nicely

docs/testing.md
berrange: Also the 'libssl-dev' was recently dropped, so i'll fix that mistake too

docs/meson.build
berrange: Have switched to 4-space

CMakeLists.txt
berrange: Meson has a '-Db_ndebug' option that a user can specify, which will enable NDEBUG; by default NDEBUG is unset, which is good as asserts are worth keeping.

https://mesonbuild.com/Builtin-options.html#base-options

docs/testing.md
jlevon: OK, I kind of prefer the old way I could load into vim, but this will do indeed, thanks.

CMakeLists.txt
jlevon: thanks for explanation.

meson.build
berrange: Yes, good point, will do so

jlevon:
I've tested coverity myself. Will do a last pass review tomorrow, thanks.

jlevon (approved):

jlevon:
Thanks for your work here! Hopefully this will simplify inclusion into qemu (and SPDK)

berrange:
That's great, thanks for being receptive to the switch to meson!

PR #667, opened 2022-05-05
drop usage of MD5_*()

jlevon:
These functions from openssl are deprecated, and hence break builds with
openssl 3.0, which now has a compiler warning for them. We only use them
to check buffer contents; replace them with CRC code from DPDK instead,
and entirely drop use of openssl.


swapnili:
Instead of copying rte_hash_crc.h, why not simply use EVP_*. There is also wrapper function EVP_Digest() which makes it way easier than using Init, Update and FInal.

jlevon:
It seemed silly to have to have the openssl dependency just for some test code, and this way we don't need to worry about openssl breaking us again.

tmakatos (approved):

swapnili (approved):

PR #669, opened 2022-05-09
build PR runs in release mode

jlevon:
Instead of "plain", which doesn't specify any optimization, use
"release", which both turns off -DDEBUG and enables -O3.


jlevon:
cc @berrange

jlevon:
Yep, there's a bug in CPython somehow related to calling into optimized libraries - I need to move our default back to unoptimized

swapnili (approved):

PR #671, opened 2022-05-10
require quiesce for VFIO_USER_DIRTY_PAGES

jlevon:
If we require a quiesce for these calls, we can be sure that it will not race
with any usage of vfu_*_sg() calls, as a first step towards concurrency.

This is not ideal for VFIO_IOMMU_DIRTY_PAGES_FLAG_GET_BITMAP, which can
potentially be called multiple times during pre-copy phase, but that's something
we can fix later.

Signed-off-by: John Levon <john.levon@<a href="nutanix.com">nutanix.com</a>>

---

**Stack**:
- #677
- #676
- #675
- #674
- #673
- #672
- #671 ⬅

⚠️ *Part of a stack created by [spr](https://github.com/ejoffe/spr). Do not merge manually using the UI - doing so may have unexpected results.*

### DIFF
--- a/test/py/test_dirty_pages.py
+++ b/test/py/test_dirty_pages.py
@@ -28,11 +28,13 @@
 #
 
 from libvfio_user import *
+import ctypes as c
 import errno
 import mmap
 import tempfile
 
 ctx = None
+quiesce_errno = 0
 
 
 @vfu_dma_register_cb_t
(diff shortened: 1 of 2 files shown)


test/py/test_dirty_pages.py
```diff
@@ -271,7 +283,13 @@ def test_dirty_pages_get_unmodified():
     assert br.bitmap.size == 8
 
 
-def get_dirty_page_bitmap():
+def get_dirty_page_bitmap_result(resp):
+    dirty_pages, resp = vfio_user_dirty_pages.pop_from_buffer(resp)
```
tmakatos, line 287: nit: since you're changing this code anyway, we don't seem to use `dirty_pages` and `br` so can we replace them with `_`?

tmakatos (approved):

PR #677, opened 2022-05-10
allow concurrent dirty bitmap get

jlevon:
Use atomic operations to allow concurrent bitmap updates with
VFIO_IOMMU_DIRTY_PAGES_FLAG_GET_BITMAP operations.

Dirtying clients can race against each other, so we must use atomic or
when marking dirty: we do this byte-by-byte.

When reading the dirty bitmap, we must be careful to not race and lose
any set bits within the same byte. If we miss an update, we'll catch it
the next time around, presuming that before the final pass we'll have
quiesced all I/O.

Signed-off-by: John Levon <john.levon@<a href="nutanix.com">nutanix.com</a>>

### DIFF
--- a/lib/dma.c
+++ b/lib/dma.c
@@ -31,7 +31,6 @@
 #include <sys/types.h>
 #include <sys/stat.h>
 #include <unistd.h>
-#include <stdio.h>
 #include <sys/param.h>
 
 #include <stddef.h>
@@ -281,7 +280,8 @@ dirty_page_logging_start_on_region(dma_memory_region_t *region, size_t pgsize)
     if (size < 0) {
         return size;
     }
-    region->dirty_bitmap = calloc(size, sizeof(char));
+
+    region->dirty_bitmap = calloc(size, 1);
     if (region->dirty_bitmap == NULL) {
         return ERROR_INT(errno);
     }
@@ -553,10 +553,11 @@ dma_controller_dirty_page_get(dma_controller_t *dma, vfu_dma_addr_t addr,
                               uint64_t len, size_t pgsize, size_t size,
                               char *bitmap)
 {
-    int ret;
+    dma_memory_region_t *region;
     ssize_t bitmap_size;
     dma_sg_t sg;
-    dma_memory_region_t *region;
+    size_t i;
+    int ret;
 
     assert(dma != NULL);
     assert(bitmap != NULL);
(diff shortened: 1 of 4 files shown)


lib/dma.c
```diff
@@ -599,11 +600,32 @@ dma_controller_dirty_page_get(dma_controller_t *dma, vfu_dma_addr_t addr,
         return ERROR_INT(EINVAL);
     }
 
-    memcpy(bitmap, region->dirty_bitmap, size);
+    for (i = 0; i < (size_t)bitmap_size; i++) {
```
tmakatos, line 603: Not required for this patch, but should we consider in the future using QWORD instead of bytes for performance?

lib/dma.c
```diff
@@ -599,11 +600,32 @@ dma_controller_dirty_page_get(dma_controller_t *dma, vfu_dma_addr_t addr,
...
+         * atomic or in _dma_mark_dirty(), this cannot lose set bits - we might
+         * miss a bit being set after, but again, we'll catch that next time
+         * around.
+         */
+        if (val == 0) {
+            *outp = 0;
+        } else {
+            uint8_t zero = 0;
+            __atomic_exchange(&region->dirty_bitmap[i], &zero,
+                              outp, __ATOMIC_SEQ_CST);
```
tmakatos, line 622: Wouldn't `__ATOMIC_RELAXED` be enough in this case, since there are no other dependent operations?

tmakatos (approved):

lib/dma.h
```diff
@@ -140,22 +141,64 @@ _dma_addr_sg_split(const dma_controller_t *dma,
...
+    range_to_pages(sg->offset, sg->length, dma->dirty_pgsize,
+                   &pgstart, &pgend);
+
+    index = bit_to_u8(pgstart);
+    end = bit_to_u8(pgend) + !!(bit_to_u8off(pgend));
+
+    for (i = index; i < end; i++) {
+        uint8_t bm = ~0;
+
+        /* Mask off any pages in the first u8 that aren't in the range. */
```
raphael-nutanix, line 191: why not do this after the loop?

lib/dma.h
```diff
@@ -140,22 +141,64 @@ _dma_addr_sg_split(const dma_controller_t *dma,
...
+        if (i == index && bit_to_u8off(pgstart) != 0) {
+            bm &= ~((1 << bit_to_u8off(pgstart)) - 1);
+        }
+
+        /* Mask off any pages in the last u8 that aren't in the range. */
```
raphael-nutanix, line 196: ditto - I think this would read a lot more cleanly if this were before the loop.

lib/dma.c
```diff
@@ -599,11 +600,32 @@ dma_controller_dirty_page_get(dma_controller_t *dma, vfu_dma_addr_t addr,
         return ERROR_INT(EINVAL);
     }
 
-    memcpy(bitmap, region->dirty_bitmap, size);
+    for (i = 0; i < (size_t)bitmap_size; i++) {
+        uint8_t val = region->dirty_bitmap[i];
```
raphael-nutanix, line 604: I don't think this is safe. What if we read a part of the bitmap some bits set, say:

00011111

Then another thread calls mark dirty on the some more bits from the byte, so that it becomes:

01111111

Then we do our atomic exchange with zero. We have val as 00011111 and have zeroed out the byte but pages 2 and 3 are still dirty.

Sadly I don't see how we can avoid the atomic exchange.

raphael-nutanix (changes requested):
The _dma_mark_dirty code looks good. I'm a little concerned about the optimization you have in dma_controller_dirty_page_get.

lib/dma.c
tmakatos: fair enough

lib/dma.c
tmakatos: We don't use `val` for the atomic exchange, we use `region->dirty_bitmap[i]`, see L621.

lib/dma.c
tmakatos: We can probably leave this as a TODO, we've seen how tricky getting memory consistency right can be :).

tmakatos (approved):

lib/dma.c
jlevon: goign to leave this for now

lib/dma.h
jlevon: ack

lib/dma.h
jlevon: ack

lib/dma.c
raphael-nutanix: Apologies - bad read on my part. This looks good.

raphael-nutanix (approved):
Sorry for the misread. I still think the loop in dma_controller_dirty_page_get could read a little cleaner, but it looks correct to me.

jlevon:
> could read a little cleaner

you could well be correct, but I've struggled to do so - do you have a suggestion?

jlevon:
in the meantime I will merge

PR #680, opened 2022-05-12
run scan-build in CI

jlevon:
Yet another static analyzer pass, this one is used by SPDK, and as it
did detect some minor issues, it's worth running.


samples/client.c
```diff
@@ -78,17 +78,17 @@ vfu_log(UNUSED vfu_ctx_t *vfu_ctx, UNUSED int level,
 static int
 init_sock(const char *path)
 {
-    int ret, sock;
 	struct sockaddr_un addr = {.sun_family = AF_UNIX};
+    int sock;
```
tmakatos, line 82: nit: identation

lib/tran_pipe.c
```diff
@@ -73,6 +73,10 @@ tran_pipe_send_iovec(int fd, uint16_t msg_id, bool is_reply,
 
     ret = writev(fd, iovecs, nr_iovecs);
 
+    /* Quieten static analysis. */
```
tmakatos, line 76: Curious, what's this about?

tmakatos (approved):

samples/client.c
jlevon: some tabs snuck in

lib/tran_pipe.c
jlevon: iovec[0] refers to stack, and clang thinks we might try to use it.

PR #681, opened 2022-05-12
re-instate -Werror

jlevon:
This got lost in the translation to meson; add -Werror for debug builds.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -49,6 +49,10 @@ common_cflags = [
     '-D_GNU_SOURCE',
 ]
 
+if opt_debug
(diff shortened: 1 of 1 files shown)


jlevon:
cc @jraman567

swapnili (approved):

eli-schwartz:
Meson has a global option for this, `meson setup builddir -Dwerror=true` to go with the `-Ddebug=true`.

jlevon:
I don't see how I would use that in the way I did, i.e. we need this on automatically if debug is set, rather than opt-in at setup time?

PR #683, opened 2022-05-13
fix fedora CI

jlevon:
Presumably Fedora 36 came out, as our docker label moved, and we can no
longer install. Fix the docker label to fedora:35 to fix.

commit-id:ae8f8c5c


jlevon:
cc @jraman567

swapnili (approved):

PR #686, opened 2022-05-16
add a diagram to README.md

jlevon:


tmakatos (approved):
Looks good, though why are we using cloud-hypervisor specifically? Shouldn't it be QEMU, as it's far more common, or VMM, which is generic?

jlevon:
> Looks good, though why are we using cloud-hypervisor specifically? Shouldn't it be QEMU, as it's far more common, or VMM, which is generic?

I just re-used the one we had from kvm forum. Since there's nothing qemu-specific it seemed to useful to indicate other possibilities exist...

PR #687, opened 2022-05-16
meson: skip tests for subproject

jraman567:
when libvfio-user is building as subproject, the test environment of the
master project sometimes is not suitable to run libvfio-user tests.

For example, some of the containers used for QEMU don't have the lspci
command, as such test-lspci.sh fails. test-linkage.sh also fails because
it searches for header files in the master project's include directory.
As such, skipping libvfio-user tests from the master project test suite.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -70,5 +70,8 @@ endif
 subdir('include')
 subdir('lib')
 subdir('samples')
-subdir('test')
 subdir('docs')
+
+if not meson.is_subproject()
+    subdir('test')
+endif


jraman567:
Example issue: https://gitlab.com/jraman/qemu/-/jobs/2461903934

We can see more issues by running **make check** with QEMU.

eli-schwartz:
It's probably better to add an option in case people actually do want to test their subprojects as well.

On the other hand, the `meson test --suite XXXX` command can select a suite based on the name of the project, so users can choose to only run the tests for the superproject.

jlevon:
Hmm, was this discussed with qemu team? It would be nice to have the tests running, the reason being that qemu is being tested/run under many more setups than we do, and *not* testing could mean we have undiscovered bugs in those environments. Is it really that hard to set up qemu env so we can pass the tests?

Alternatively, can we make things like lspci skippable instead of failing if the binary doesn't exist?

jraman567:
I didn't discuss it with QEMU about this - I can certainly check.

I'm also seeing the following error:
```
153/173 libvfio-user:functional / test-linkage.sh                              FAIL            0.01s   exit status 1
>>> MALLOC_PERTURB_=153 /home/qemu-separation/qemu-split/subprojects/libvfio-user/test/test-linkage.sh /home/qemu-separation/qemu-split /home/qemu-separation/qemu-split/build 'cc -m64 -mcx16'
―――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――― ✀  ――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――
stderr:
grep: /home/qemu-separation/qemu-split/include/libvfio-user.h: No such file or directory
cc: error: /home/qemu-separation/qemu-split/build/lib/libvfio-user.so: No such file or directory
――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――
```

I could make try making the failing tests as skippable.

PR #688, opened 2022-05-17
tests: add checks to confirm test environment

jraman567:
test-lspci.sh: some test platforms don't include the lspci command, as
such skip this test if lspci is not found

test-linkage.sh: specify the source and build root paths of the subproject
instead of the root paths of the master project


jlevon (approved):
thanks!

PR #689, opened 2022-05-19
libvfio-user.h: sync VFIO_DEVICE_STATE* definitions with upstream

jraman567:
sync VFIO_DEVICE_STATE* definitions with latest ones in upstream

### DIFF
--- a/include/libvfio-user.h
+++ b/include/libvfio-user.h
@@ -671,7 +671,19 @@ typedef struct {
 
 } vfu_migration_callbacks_t;
 
-#ifndef VFIO_DEVICE_STATE_STOP
+/*
+ * The VFIO protocol is evolving w.r.t migration. It is moving away
+ * from a VFIO region based migration. The following keeps our headers
+ * in sync with latest kernel VFIO header.
+ */
+#if defined(VFIO_DEVICE_STATE_V1_STOP)
+
+#define VFIO_DEVICE_STATE_STOP      VFIO_DEVICE_STATE_V1_STOP
+#define VFIO_DEVICE_STATE_RUNNING   VFIO_DEVICE_STATE_V1_RUNNING
+#define VFIO_DEVICE_STATE_SAVING    VFIO_DEVICE_STATE_V1_SAVING
+#define VFIO_DEVICE_STATE_RESUMING  VFIO_DEVICE_STATE_V1_RESUMING
+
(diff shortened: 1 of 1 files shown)


jraman567:
@tmakatos

include/libvfio-user.h
```diff
@@ -671,15 +671,27 @@ typedef struct {
...
 
-#ifndef VFIO_DEVICE_STATE_STOP
+/*
+ * The VFIO protocol is evolving w.r.t migration. It is moving away
+ * from a VFIO region based migration. The following keeps our headers
+ * in sync with latest kernel VFIO header.
+ */
+#if defined(VFIO_DEVICE_STATE_V1_STOP)
+
+#define VFIO_DEVICE_STATE_STOP      VFIO_DEVICE_STATE_V1_STOP
```
tmakatos, line 681: I'm concerned that this might lead to confusion: if `VFIO_REGION_TYPE_MIGRATION_DEPRECATED ` is defined then we're defining `VFIO_DEVICE_STATE_STOP`, but `VFIO_DEVICE_STATE_STOP` already exists in `enum vfio_device_mig_state`.

Can we instread use XXX_V1_XXX everywhere and only define those if `VFIO_REGION_TYPE_MIGRATION_DEPRECATED ` is not defined?

tmakatos (changes requested):

include/libvfio-user.h
```diff
@@ -671,15 +671,27 @@ typedef struct {
 
 } vfu_migration_callbacks_t;
 
-#ifndef VFIO_DEVICE_STATE_STOP
+/*
+ * The VFIO protocol is evolving w.r.t migration. It is moving away
+ * from a VFIO region based migration. The following keeps our headers
+ * in sync with latest kernel VFIO header.
+ */
+#if defined(VFIO_DEVICE_STATE_V1_STOP)
```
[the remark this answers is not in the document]
jraman567: OK

include/libvfio-user.h
jraman567: OK.

Do you mean confusing to the readers or the compiler? The compiler would expand the macros at preprocessing though. But I agree it could cause confusion when surveying the code when a macro with the same name as symbol is defined.

tmakatos: I mean it would confuse the reader.

PR #690, opened 2022-05-20
libvfio-user.h: sync VFIO_DEVICE_STATE_XXXX definitions with upstream

jraman567:
Rename VFIO_DEVICE_STATE_XXXX defines as VFIO_DEVICE_STATE_V1_XXXX.
Upstream renamed these variable to be of the XXXX_V1_XXXX format and
switched an enum for VFIO_DEVICE_STATE_XXXX.


jraman567:
@tmakatos made the changes you suggested. Sorry, ended up creating new pull request.

include/libvfio-user.h
```diff
@@ -671,15 +671,24 @@ typedef struct {
...
-#ifndef VFIO_DEVICE_STATE_STOP
+#if defined(VFIO_DEVICE_STATE_STOP)
 
-#define VFIO_DEVICE_STATE_STOP (0)
-#define VFIO_DEVICE_STATE_RUNNING (1 << 0)
-#define VFIO_DEVICE_STATE_SAVING (1 << 1)
-#define VFIO_DEVICE_STATE_RESUMING (1 << 2)
-#define VFIO_DEVICE_STATE_MASK ((1 << 3) - 1)
+/* We are on a system that uses older VFIO header files */
+#define VFIO_DEVICE_STATE_V1_STOP         VFIO_DEVICE_STATE_STOP
```
tmakatos, line 677: We don't need this at all, if `VFIO_REGION_TYPE_MIGRATION_DEPRECATED` then just define `VFIO_DEVICE_STATE_V1_STOP` etc. as you do below.

tmakatos (changes requested):
nit: I think email address in SOB should be enclosed in '<' '>'.

include/libvfio-user.h
jraman567: @tmakatos The pre-push tests in this repo. fails if we just depend on  `VFIO_REGION_TYPE_MIGRATION_DEPRECATED`. The ubuntu-22 system is the one showing the failure. The failure message was redefinition of `VFIO_DEVICE_STATE_MASK`.

The "linux/vfio.h" file on this system has  VFIO_DEVICE_STATE_STOP defined, whereas it does not have VFIO_REGION_TYPE_MIGRATION_DEPRECATED. So if we rely on just VFIO_REGION_TYPE_MIGRATION_DEPRECATED, we would hit the VFIO_DEVICE_STATE_MASK redefinition failure.

We could separately have checks for VFIO_DEVICE_STATE_MASK and define is only when it's not previously defined, but that would not be accurate - VFIO_DEVICE_STATE_MASK should be based on the macros we define in this file.

jraman567:
> nit: I think email address in SOB should be enclosed in '<' '>'.

Hey, that's weird. It was enclosed within arrow brackets when I pushed. Lemme hunt the arrow bracket monster.

include/libvfio-user.h
jraman567: > 



> > nit: I think email address in SOB should be enclosed in '<' '>'.
> 
> Hey, that's weird. It was enclosed within arrow brackets when I pushed. Lemme hunt the arrow bracket monster.

It does show the arrow brackets when we look-up the commit:
https://github.com/jraman567/libvfio-user/commit/70d51e5b04610b61e677414566ac09993e69b021

I think it's an issue with the viewer for pull requests?

tmakatos:
> > nit: I think email address in SOB should be enclosed in '<' '>'.
> 
> Hey, that's weird. It was enclosed within arrow brackets when I pushed. Lemme hunt the arrow bracket monster.

Indeed that's the PR message, the commit itself is fine, nothing to be fixed here.

include/libvfio-user.h
tmakatos: I see the problem, I guess your solution is correct then. Can we have a comment in the code explaining the rationale begin this? Also, as a sanity check, can we have `_Static_assert(VFIO_DEVICE_STATE_STOP == 0)` etc. for this case?

jraman567: OK, will add a comment and static assert

I see, you're relying on VFIO_DEVICE_STATE_STOP / VFIO_DEVICE_STATE_V1_STOP starting at 0?

JSYK, not connected, the new device state enums start with VFIO_DEVICE_STATE_ERROR at zero, and _STOP is one.

include/libvfio-user.h
```diff
@@ -671,15 +671,33 @@ typedef struct {
 
 } vfu_migration_callbacks_t;
 
-#ifndef VFIO_DEVICE_STATE_STOP
+/**
+ * The definition for VFIO_DEVICE_STATE_XXX differs with the version of vfio
+ * header file used. Some old systems wouldn't have these definitions. Some
+ * other newer systems would have be using region based migration, and not
```
jlevon, line 677: "would be using"

jlevon (commented):
Please also fix the python code in test/py/

do we need an update to docs/vfio-user.rst too?

jlevon (approved):

PR #692, opened 2022-05-31
dma: Clean up _dma_mark_dirty

raphael-nutanix:
Low priority - just the suggestion I had here: https://github.com/nutanix/libvfio-user/pull/677#issuecomment-1140871862

### DIFF
--- a/lib/dma.h
+++ b/lib/dma.h
@@ -165,14 +165,43 @@ bit_to_u8off(size_t val)
 }
 
 static inline void
+u8_set_all_bits(uint8_t *bm)
+{
+    *bm = ~0;
+}
+
+static inline void
+mask_any_lower_bits(uint8_t *byte, size_t u8off)
+{
+    if (bit_to_u8off(u8off) != 0) {
+        *byte &= ~((1 << bit_to_u8off(u8off)) - 1);
+    }
+}
+
+static inline void
+mask_any_higher_bits(uint8_t *byte, size_t u8off)
+{
+    if (bit_to_u8off(u8off) != 0) {
+        *byte &= ((1 << bit_to_u8off(u8off)) - 1);
+    }
+}
+
+static inline void
+u8_atomic_or(uint8_t *bitmap, uint8_t byte_mask)
+{
+    __atomic_or_fetch(bitmap, byte_mask, __ATOMIC_RELAXED);
+}
+
+static inline void
 _dma_mark_dirty(const dma_controller_t *dma, const dma_memory_region_t *region,
                 dma_sg_t *sg)
 {
-    size_t index;
-    size_t end;
+    size_t bm_start_byte;
+    size_t bm_end_byte;
+    size_t middle_bytes;
     size_t pgstart;
     size_t pgend;
-    size_t i;
+    uint8_t bm;
 
     assert(dma != NULL);
     assert(region != NULL);
@@ -182,24 +211,47 @@ _dma_mark_dirty(const dma_controller_t *dma, const dma_memory_region_t *region,
     range_to_pages(sg->offset, sg->length, dma->dirty_pgsize,
                    &pgstart, &pgend);
 
-    index = bit_to_u8(pgstart);
-    end = bit_to_u8(pgend) + !!(bit_to_u8off(pgend));
+    bm_start_byte = bit_to_u8(pgstart);
+    bm_end_byte = bit_to_u8(pgend) + !!(bit_to_u8off(pgend));
 
-    for (i = index; i < end; i++) {
-        uint8_t bm = ~0;
+    assert(bm_start_byte <= bm_end_byte);
 
-        /* Mask off any pages in the first u8 that aren't in the range. */
-        if (i == index && bit_to_u8off(pgstart) != 0) {
-            bm &= ~((1 << bit_to_u8off(pgstart)) - 1);
-        }
-
-        /* Mask off any pages in the last u8 that aren't in the range. */
-        if (i == end - 1 && bit_to_u8off(pgend) != 0) {
-            bm &= ((1 << bit_to_u8off(pgend)) - 1);
-        }
-
-        __atomic_or_fetch(&region->dirty_bitmap[i], bm, __ATOMIC_SEQ_CST);
+    u8_set_all_bits(&bm);
+    /* If only one byte of the bitmap has dirty pages, mark them and return. */
+    if (bm_start_byte >= bm_end_byte - 1) {
+        mask_any_lower_bits(&bm, pgstart);
+        mask_any_higher_bits(&bm, pgend);
+        u8_atomic_or(&region->dirty_bitmap[bm_start_byte], bm);
+        /*
+         * Since we are using relaxed consistency when marking the bitmap, we
+         * add a write barrier to ensure all writes complete before returning.
+         *
+         * This protects us from missing pages on the final dirty bitmap read
+         * before proceeding to 'stop-and-copy'.
+         */
+        vfu_smp_wmb();
+        return;
     }
+
+    /* Mask off any pages in the first u8 that aren't in the range. */
+    mask_any_lower_bits(&bm, pgstart);
+    u8_atomic_or(&region->dirty_bitmap[bm_start_byte], bm);
+
+    /* Set 1s for all of the pages in the middle u8s of the range. */
+    u8_set_all_bits(&bm);
+    middle_bytes = bm_end_byte - bm_start_byte - 2;
+    memset(&region->dirty_bitmap[bm_start_byte + 1], (int)bm, middle_bytes);
+
+    /* Mask off any pages in the last u8 that aren't in the range. */
+    mask_any_higher_bits(&bm, pgend);
+    u8_atomic_or(&region->dirty_bitmap[bm_end_byte - 1], bm);
+
+    /*
+     * Like the prior vfu_smp_wmb(), we add one here to ensure that on the
+     * final iteration all of the writes to the bitmap persist before returning
+     * from _dma_mark_dirty().
+     */
+    vfu_smp_wmb();
 }
 
 static inline int
--- a/lib/private.h
+++ b/lib/private.h
@@ -58,6 +58,22 @@
                              sizeof(struct vfio_user_region_access))
 
 /*
+ * Based on QEMU's atomic primitives.
+ *
+ * vfu_barrier() prevents compiler re-ordering. The other vfu_smp* barriers
+ * guard against out-of-order execution with the same semantics as the smp_*
+ * barriers in the Linux kernel. See the kernel memory_barriers.txt docs for
+ * detailed explanations.
+ *
+ * These are required by the dma code to make marking and clearing the dirty
+ * bitmap threadsafe.
+ */
+#define vfu_barrier() ({ asm volatile("" ::: "memory"); (void)0; })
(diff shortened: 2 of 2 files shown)


jlevon:
@raphael-nutanix as I mentioned, this implementation won't work as is, you should be able to reproduce with "make pytest"

raphael-nutanix:
Haven't debugged the timeout. I'm seeing it on my local with the prior commit checked out so not sure it's me.

jlevon:
timeout isn't you, we have a long-standing race in client/server code.

lib/dma.h
```diff
@@ -164,15 +164,44 @@ bit_to_u8off(size_t val)
     return val % (CHAR_BIT);
 }
 
+static inline void
+u8_set_all(uint8_t *bm)
```
jlevon, line 168: u8_set_all_bits() ?

lib/dma.h
```diff
@@ -164,15 +164,44 @@ bit_to_u8off(size_t val)
...
+static inline void
+mask_any_higher_bits(uint8_t *byte, size_t end_idx)
+{
+    if (bit_to_u8off(end_idx) != 0) {
+        *byte &= ((1 << bit_to_u8off(end_idx)) - 1);
+    }
+}
+
+static inline void
+apply_byte_mask(uint8_t *bitmap, uint8_t byte_mask)
```
jlevon, line 190: it's not a mask if we're or'ing. Can we just call this u8_atomic_or() to match other naming?

lib/dma.h
```diff
@@ -164,15 +164,44 @@ bit_to_u8off(size_t val)
...
 }
 
+static inline void
+u8_set_all(uint8_t *bm)
+{
+    *bm = ~0;
+}
+
+static inline void
+mask_any_lower_bits(uint8_t *byte, size_t start_idx)
```
jlevon, line 174: nit, not a fan of "start_idx" here, how about u8off to match bit_to_u8off?

lib/dma.h
```diff
@@ -164,15 +164,44 @@ bit_to_u8off(size_t val)
...
+{
+    if (bit_to_u8off(start_idx) != 0) {
+        *byte &= ~((1 << bit_to_u8off(start_idx)) - 1);
+    }
+}
+
+static inline void
+mask_any_higher_bits(uint8_t *byte, size_t end_idx)
```
jlevon, line 182: same

lib/dma.h
```diff
@@ -164,15 +164,44 @@ bit_to_u8off(size_t val)
...
+    __atomic_or_fetch(bitmap, byte_mask, __ATOMIC_SEQ_CST);
+}
+
 static inline void
 _dma_mark_dirty(const dma_controller_t *dma, const dma_memory_region_t *region,
                 dma_sg_t *sg)
 {
-    size_t index;
-    size_t end;
+    size_t bm_sg_start;
```
jlevon, line 199: I'm not quite following what the "sg" part of this is - we already converted from sg in the range_to_pages() call.

lib/dma.h
```diff
@@ -182,24 +211,33 @@ _dma_mark_dirty(const dma_controller_t *dma, const dma_memory_region_t *region,
...
-            bm &= ~((1 << bit_to_u8off(pgstart)) - 1);
-        }
+    bm_sg_start = bit_to_u8(pgstart);
+    bm_sg_end = bit_to_u8(pgend) + !!(bit_to_u8off(pgend));
+
+    assert(bm_sg_start <= bm_sg_end);
+    
+    u8_set_all(&bm);
+    /* If only one byte of the bitmap has dirty pages, mark them and return. */
+    if (pgstart == pgend || bm_sg_start == bm_sg_end - 1) {
```
jlevon, line 221: what do we need the first pgstart == pgend check here?

lib/dma.h
```diff
@@ -182,24 +211,33 @@ _dma_mark_dirty(const dma_controller_t *dma, const dma_memory_region_t *region,
...
-        }
+    /* Mask off any pages in the first u8 that aren't in the range. */
+    mask_any_lower_bits(&bm, pgstart);
+    apply_byte_mask(&region->dirty_bitmap[bm_sg_start], bm);
 
-        __atomic_or_fetch(&region->dirty_bitmap[i], bm, __ATOMIC_SEQ_CST);
+    /* Set 1s for all of the pages in the middle u8s of the range. */
+    u8_set_all(&bm);
+    for (i = bm_sg_start + 1; i < bm_sg_end - 1; i++) {
+        apply_byte_mask(&region->dirty_bitmap[i], bm);
```
jlevon, line 235: No need for a loop, this can be a memset() now!

lib/dma.h
```diff
@@ -182,24 +211,33 @@ _dma_mark_dirty(const dma_controller_t *dma, const dma_memory_region_t *region,
...
+        return;
+    }
 
-        /* Mask off any pages in the last u8 that aren't in the range. */
-        if (i == end - 1 && bit_to_u8off(pgend) != 0) {
-            bm &= ((1 << bit_to_u8off(pgend)) - 1);
-        }
+    /* Mask off any pages in the first u8 that aren't in the range. */
+    mask_any_lower_bits(&bm, pgstart);
+    apply_byte_mask(&region->dirty_bitmap[bm_sg_start], bm);
```
jlevon, line 230: can use relaxed consistency now it's split up, and add a barrier at the end

lib/dma.h
```diff
@@ -182,24 +211,33 @@ _dma_mark_dirty(const dma_controller_t *dma, const dma_memory_region_t *region,
...
-        __atomic_or_fetch(&region->dirty_bitmap[i], bm, __ATOMIC_SEQ_CST);
+    /* Set 1s for all of the pages in the middle u8s of the range. */
+    u8_set_all(&bm);
+    for (i = bm_sg_start + 1; i < bm_sg_end - 1; i++) {
+        apply_byte_mask(&region->dirty_bitmap[i], bm);
     }
+
+    /* Mask off any pages in the last u8 that aren't in the range. */
+    mask_any_higher_bits(&bm, pgend);
+    apply_byte_mask(&region->dirty_bitmap[bm_sg_end - 1], bm);
```
jlevon, line 240: can either make this relaxed too, or - probably better - have an explicit barrier at the end after?

jlevon (commented):
thanks, this is probably easier to understand, and could be a bit faster than before, in theory anyway

lib/dma.h
raphael-nutanix: ACK

lib/dma.h
raphael-nutanix: ACK

lib/dma.h
raphael-nutanix: ACK

lib/dma.h
raphael-nutanix: ACK

lib/dma.h
raphael-nutanix: Agreed - not a great name. Maybe we call it bm_start_byte? I think it's worth emphasizing that this is the first byte of the sg list.

lib/dma.h
raphael-nutanix: ACK RE relaxed. Why do we need the barrier?

lib/dma.h
raphael-nutanix: AH yes, I think it can!

lib/dma.h
raphael-nutanix: again, why do we need the barrier here?

lib/dma.h
raphael-nutanix: so if pgstart == pgend && pgstart % 8 == 0 then we have bm_sg_start == bm_sg_end. The condition could be rewritten:

if (bm_sg_start == bm_sg_end || bm_sg_start == bm_sg_end - 1)

I think that's a bit easier on the eyes so maybe I'll do that and add a comment.

lib/dma.h
jlevon: works for me

lib/dma.h
jlevon: ok, but you can do that with a single less than /more than comparison

lib/dma.h
jlevon: See previously PR - not sure we do, really, it was paranoia as we had to say something.

raphael-nutanix: I assume this would be a smp_wmb()? I'd see no reason for a read barrier since if dma_controller_dirty_page_get() missed the write we would pick it up on the next pass, and concurrent calls to _dma_mark_dirty() won't read it outside of an atomic operation.

The argument that the memset() should be safe for the middle bytes should imply that the order in which the byte writes to the bitmap persist shouldn't matter, right? Why would it be different for the first or last byte, especially considering it is an atomic-or? Therefore I don't think we need a smp_wmb().

> it was paranoia as we had to say something.

I'm struggling to find the discussion. Was there a failure? I'd rather not add barriers we can't fully justify but I'll defer to your judgement.

jlevon: https://github.com/nutanix/libvfio-user/pull/677#discussion_r880374347

So, the question is on the very last pass - is it possible for our writes here not to show up prior to the final grab of bitmap data? I think we might need a full barrier at the point we read the bitmap instead? I haven't thought about it properly, hence the previous seq consistency changes.

> especially considering it is an atomic-or

we still need to pick a consistency flag

jlevon: If you mean smb_wmb() and smb_rmb(), and using relaxed consistency in all atomic ors, then right, that's what I was sugessting. But I also think smb_mb() before dirty_page_Get would work on its own, but again, haven't "done the maths" on that.

lib/dma.c
```diff
@@ -601,8 +601,11 @@ dma_controller_dirty_page_get(dma_controller_t *dma, vfu_dma_addr_t addr,
     }
 
     for (i = 0; i < (size_t)bitmap_size; i++) {
-        uint8_t val = region->dirty_bitmap[i];
-        uint8_t *outp = (uint8_t *)&bitmap[i];
+        uint8_t val, *outp;
+
+        outp = (uint8_t *)&bitmap[i];
+        dma_smp_rmb();
```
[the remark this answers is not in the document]
jlevon: also, every barrier should have a comment explaining it please.
Are we sure we need this given the seq cst in line :624?

lib/dma.h
```diff
@@ -82,6 +82,12 @@
 
 #define iov_end(iov) ((iov)->iov_base + (iov)->iov_len)
 
+/* Based on QEMU's atomic primitives. */
```
jlevon, line 85: move to lib/private.h with suitable prefix?

Also, I think at least having the comment as to why we need dma_barrier() is worth having too

raphael-nutanix: ACK

lib/dma.c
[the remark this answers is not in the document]
raphael-nutanix: Agreed - I think it's completely unnecessary. Removing it.

The barrier for the 'or' on the first byte is similarly unnecessary. I'm removing that too.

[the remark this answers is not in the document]
raphael-nutanix: Done

lib/dma.h
raphael-nutanix: Done

tmakatos:
Is this PR still relevant?

raphael-nutanix:
It's as relevant as it ever was. I still think it's a nice cleanup and possibly comes with a perf boost but I'll leave it to you guys to decide.

jlevon:
Closing this as there's nothing particularly wrong with the existing code, that we're aware of.

PR #700, opened 2022-07-21
disable client-server test by default

jlevon:
This test is flaky: there is some kind of race that causes the test to
hang. Now we are run as part of qemu CI, we need to disable this by
default, until we can find time to fix the test.


jlevon:
cc @berrange

swapnili (approved):
Should we open an issue to track fixing of this test?

jlevon:
> Should we open an issue to track fixing of this test?

yes

PR #702, opened 2022-08-03
delete socket on vfu_ctx_destroy

### LINKED ISSUE #660: vfu_destroy_ctx() doesn't delete the socket file that created via vfu_create_ctx()
As described in the title, the socket file is still there after calling vfu_destroy_ctx().

tmakatos:
fixes #660

### DIFF
--- a/lib/tran_sock.c
+++ b/lib/tran_sock.c
@@ -414,6 +414,7 @@ tran_sock_init(vfu_ctx_t *vfu_ctx)
     ret = listen(ts->listen_fd, 0);
     if (ret < 0) {
         ret = errno;
+        (void) unlink(vfu_ctx->uuid);
     }
 
 out:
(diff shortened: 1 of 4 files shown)


lib/tran_sock.c
```diff
@@ -651,10 +652,13 @@ tran_sock_fini(vfu_ctx_t *vfu_ctx)
 
     ts = vfu_ctx->tran_data;
 
-    if (ts != NULL && ts->listen_fd != -1) {
-        // FIXME: handle EINTR
-        (void) close(ts->listen_fd);
-        ts->listen_fd = -1;
+    if (ts != NULL) {
+        (void) unlink(vfu_ctx->uuid);
```
swapnili, line 656: unlink() can also happen under previous check,
 if (ts != NULL && ts->listen_fd != -1) { }
Not a big gain, but will save one less, useless call to unlink()

jlevon (approved):

lib/tran_sock.c
swapnili: Right I missed the fact that we dont set  tran_data in case of error in tran_sock_init(). thanks

swapnili (approved):

PR #704, opened 2022-08-16
support VFIO_USER_REGION_WRITE_MULTI

jlevon:


tmakatos:
Why do we need this?

PR #709, opened 2022-09-02
fix compilation for i386 and ppc64

tmakatos:


jlevon:
I really hate PRix64 and friends, they make code so unreadable. Can't we just use %llx and cast the arg instead?

Also, if we want to support these (I'm surprised i686 is still relevant), we need some CI to at least build + test them.

tmakatos:
> I really hate PRix64 and friends, they make code so unreadable. Can't we just use %llx and cast the arg instead?
> 
Sure.

> Also, if we want to support these (I'm surprised i686 is still relevant), we need some CI to at least build + test them.
Indeed.

jlevon:
BTW, someone pointed me to:
https://lore.kernel.org/lkml/alpine.DEB.2.22.394.2208251435370.104368@digraph.polyomino.org.uk/

Looking forward to that!!

lib/libvfio-user.c
```diff
@@ -718,7 +726,8 @@ handle_dma_map(vfu_ctx_t *vfu_ctx, vfu_msg_t *msg,
         }
     }
 
-    ret = dma_controller_add_region(vfu_ctx->dma, (void *)dma_map->addr,
+    ret = dma_controller_add_region(vfu_ctx->dma,
+                                    (vfu_dma_addr_t)(uintptr_t)dma_map->addr,
```
tmakatos, line 730: Perhaps we should use `uinptr_t` instead of `void *` for `vfu_dma_addr_t`?

jlevon: I think uintptr_t is better yeah

jlevon:
why do we need our own vfio_user_region_info?

tmakatos:
> why do we need our own vfio_user_region_info?

We don't, that was from the first version where I was working around the different typedef for `u64` between x86_64 and i386.

lib/libvfio-user.c
tmakatos: We'll need a lot of casts because we compare `struct iovec.iov_base` to `vfu_dma_addr_t`, not sure it's worth it. Unless we drop `struct iovec.iov_base`, which I'm not sure it's nice either.

tmakatos:
> > Also, if we want to support these (I'm surprised i686 is still relevant), we need some CI to at least build + test them.
> 
> Indeed.

From a quick look I had at https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners, it doesn't look like GitHub provides a 32-bit VM. Or we'd have to enable cross-compilation some how.

tmakatos:
Next, I'll test on PPC and then will mark PR ready for review.

jlevon:
Cross-compilation doesn't help too much, since we couldn't run tests.

test/py/libvfio_user.py
```diff
@@ -89,10 +89,10 @@
 VFIO_DEVICE_FLAGS_RESET = (1 << 0)
 VFIO_DEVICE_FLAGS_PCI = (1 << 1)
 
-VFIO_REGION_INFO_FLAG_READ = (1 << 0)
```
jlevon, line 92: do you still need/want these changes?

lib/dma.c
```diff
@@ -300,8 +300,8 @@ MOCK_DEFINE(dma_controller_add_region)(dma_controller_t *dma,
 
     assert(dma != NULL);
 
-    snprintf(rstr, sizeof(rstr), "[%p, %p) fd=%d offset=%#lx prot=%#x",
-             dma_addr, (char *)dma_addr + size, fd, offset, prot);
+    snprintf(rstr, sizeof(rstr), "[%p, %p) fd=%d offset=%#llx prot=%#x",
+             dma_addr, dma_addr + size, fd, (unsigned long long)offset, prot);
```
jlevon, line 304: how about a helper macro for all these, LLX() or some such?

test/py/libvfio_user.py
tmakatos: We don't need them, I forgot to remove them.

lib/dma.c
tmakatos: I'm not clear, you want the format string and the arguments replaced by a macro?

tmakatos:
I've added a few more fixes

- Still need to fix tests on ppc64.
- Need to fix pylint errors (see faile PR run).
- Need to fix QEMU using libvfio-user (https://urldefense.proofpoint.com/v2/url?u=https-3A__koji.fedoraproject.org_koji_taskinfo-3FtaskID-3D92194687&d=DwQFaQ&c=s883GpUCOChKOHiocYtGcg&r=XTpYsh5Ps2zJvtw6ogtti46atk736SI4vgsJiUKIyDE&m=5dCi1oG4l96CCRmCk7LuGrfPLj-WoDoa5qecaQ6aWiBBzNML2A1Ph29cM0wyvdQp&s=lYo_Ew6-Rkcr_nFoFAaatVaP1nXXuBS4cXOYuW8GRM8&e=)

lib/dma.c
tmakatos: Let's go with the latter.

PR #727, opened 2022-12-30
allow -1 file descriptor for ioregionfd

tmakatos:
This allows setting up a shadow ioeventfd without an eventfd. This can be useful when the ioeventfd doesn't need to be kicked but the value still needs to be copied e.g. NVMe completion queue head doorbells.

### DIFF
--- a/lib/libvfio-user.c
+++ b/lib/libvfio-user.c
@@ -483,7 +483,6 @@ vfu_create_ioeventfd(vfu_ctx_t *vfu_ctx, uint32_t region_idx, int fd,
     vfu_reg_info_t *vfu_reg;
 
     assert(vfu_ctx != NULL);
-    assert(fd >= 0);
 
 #ifndef SHADOW_IOEVENTFD
     if (shadow_fd != -1) {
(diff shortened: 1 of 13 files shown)


lib/libvfio-user.c
```diff
@@ -551,6 +550,10 @@ add_fd_index(int *out_fds, size_t *nr_out_fds, int fd_search)
     assert(out_fds != NULL);
     assert(nr_out_fds != NULL);
 
+    if (fd_search < 0) {
```
jlevon, line 553: add behaviour to comment above? Also check for -1 explicitly maybe? And assert for any other negative value?

jlevon (commented):
can we add a test?

lib/libvfio-user.c
tmakatos: ok

tmakatos:
> can we add a test?

done

lib/libvfio-user.c
```diff
@@ -542,6 +541,15 @@ free_regions(vfu_ctx_t *vfu_ctx)
  * This function is used to add fd's to the fd return array and gives you back
  * the index of the fd that has been added. If the fd is already present it will
  * return the index to that duplicate fd to reduce the number of fd's sent.
+ * The fd must be a valid fd or -1, any other negative value is not permitted.
+ *
+ * out_fds: an array where the fd is stored
+ * nr_our_fds: pointer to memory that contains the size of the array
```
jlevon, line 547: nr_out_fds

jlevon (approved):

PR #728, opened 2023-01-01
test: use `grep -E` instead of `egrep`

scop:
`egrep` has been deprecated in GNU grep since 2007, and since 3.8 it emits obsolescence warnings:
https://git.savannah.gnu.org/cgit/grep.git/commit/?id=a9515624709865d480e3142fd959bccd1c9372d1


jlevon:
Thanks for your fix, do you mind adding a Signed-off-by ?

scop:
Done.

jlevon (approved):

jlevon:
Thank you.

Issue #387, opened 2021-03-08
no way to trigger an err or req IRQ


jlevon:
```
496 /**                                                                              
497  * Triggers an interrupt.                                                        
498  *                                                                               
499  * libvfio-user takes care of using the correct IRQ type (IRQ index: INTx or     
500  * MSI/X), the caller only needs to specify the sub-index.                       
501  *                                                                               
502  * @vfu_ctx: the libvfio-user context to trigger interrupt                       
503  * @subindex: vector subindex to trigger interrupt on                            
504  *                                                                               
505  * @returns 0 on success, or -1 on failure. Sets errno.                          
506  */                                                                              
507 int                                                                              
508 vfu_irq_trigger(vfu_ctx_t *vfu_ctx, uint32_t subindex);                          
509                                                                                  
510 /**                                                                              
511  * Sends message to client to trigger an interrupt.                              
512  *                                                                               
513  * libvfio-user takes care of using the IRQ type (INTx, MSI/X), the caller only  
514  * needs to specify the sub-index.                                               
515  * This api can be used to trigger interrupt by sending message to client.       
516  *                                                                               
517  * @vfu_ctx: the libvfio-user context to trigger interrupt                       
518  * @subindex: vector subindex to trigger interrupt on                            
519  *                                                                               
520  * @returns 0 on success, or -1 on failure. Sets errno.                          
521  */                                                                              
522 int                                                                              
523 vfu_irq_message(vfu_ctx_t *vfu_ctx, uint32_t subindex);                          
```

This API means that it's not possible to trigger an ERR or REQ interrupt. Need to understand the semantics of these interrupts, and make sure we have this covered (perhaps by allowing specification of irq type)

pogobanane:
The same mechanism for err and req should also be used for INTx:

I believe INTx also may be fired by devices, even when MSI(-X) is active. While software must restrict itself to either MSI or MSI-X according to pci spec, i cannot find any such statement regarding INTx and MSI(-X). Instead there is a dedicated CSR bit to indicate the ability to generate INTx (`Command Register (Offset 06h)` bit 10, `#define  PCI_COMMAND_INTX_DISABLE 0x400 /* INTx Emulation Disable */`).

PR #731, opened 2023-04-23
fix err/req irq fd issue

limiao-intel:
When handle_device_set_irqs set err irq/req irq, fd will be filled in vfu_ctx->irqs->efds[] rather than vfu_ctx->irqs->err_efd/vfu_ctx->irqs->req_efd. So, err/req irq fd will be mistakenly used as msi/x irq fd, which will cause program stuck.


lib/irq.c
```diff
@@ -189,7 +189,12 @@ irqs_set_data_none(vfu_ctx_t *vfu_ctx, struct vfio_irq_set *irq_set)
     eventfd_t val;
 
     for (i = irq_set->start; i < (irq_set->start + irq_set->count); i++) {
-        efd = vfu_ctx->irqs->efds[i];
+        if (irq_set->index == VFIO_PCI_ERR_IRQ_INDEX)
```
jlevon, line 192: please make this be a common helper - and use switch() ?

lib/irq.c
```diff
@@ -243,21 +253,38 @@ irqs_set_data_eventfd(vfu_ctx_t *vfu_ctx, struct vfio_irq_set *irq_set,
...
+                vfu_ctx->irqs->efds[i] = -1;
         }
         assert(data[j] >= 0);
         /*
          * We've already checked in handle_device_set_irqs that
          * nr_fds == irq_set->count.
          */
-        vfu_ctx->irqs->efds[i] = consume_fd(data, irq_set->count, j);
-        vfu_log(vfu_ctx, LOG_DEBUG, "event fd[%d]=%d", i, vfu_ctx->irqs->efds[i]);
+        if (irq_set->index == VFIO_PCI_ERR_IRQ_INDEX) {
```
jlevon, line 278: please just consume_fd() once into a local var, then set as appropriate

lib/irq.c
```diff
@@ -217,7 +222,12 @@ irqs_set_data_bool(vfu_ctx_t *vfu_ctx, struct vfio_irq_set *irq_set, void *data)
 
     for (i = irq_set->start, d8 = data; i < (irq_set->start + irq_set->count);
          i++, d8++) {
-        efd = vfu_ctx->irqs->efds[i];
+        if (irq_set->index == VFIO_PCI_ERR_IRQ_INDEX)
```
jlevon, line 225: Were you able to test any of these changes? I'm guessing not?

lib/irq.c
limiao-intel: If irq_set->index is equal to VFIO_PCI_ERR_IRQ_INDEX/VFIO_PCI_REQ_IRQ_INDEX, irq_set->count must be 1. So consume_fd() only is called once. I think this check can be added in device_set_irqs_validate?

lib/irq.c
limiao-intel: I test irqs_set_data_eventfd in my case. It works.

jlevon: You mean, you did a unit test? Is there any way we can add it to the python component tests?

lib/irq.c
jlevon: What I mean is, you are calling it on line :279, line :282 and line :285. I'm saying it would be simpler if you did this:

```
    efd = consume_fd(data, irq_set->count, j);
    switch (irq_set->index) {
    case VFIO_PCI_ERR_IRQ_INDEX:
           vfu_ctx->irqs->err_efd = efd;
           vfu_log(...
           break;
    ...
```

lib/irq.c
limiao-intel: Ok, I will replace all if else with switch().

lib/irq.c
limiao-intel: Ok, sorry for the misunderstanding. I will modify it.

lib/irq.c
limiao-intel: I did not a unit test. I just used it in my emulated device implementation. We can test it by running test_device_set_irqs.py.

jlevon: Does this mean you have some other changes pending for supporting your device? Since you can't currently actually generate these types of IRQs with the current API

limiao-intel: Actually, my device will only use msi/x irq. But qemu will set err irq and my device will use err irq as msi/x irq mistakenly. After adding this patch, this bug will be fixed.

lib/irq.c
tmakatos: I guess for `ERR` and `REQ` case, if wouldn't make sense for `irq_set->count` not to be 1? I suppose it might be technically correct, was wondering whether we could simplify the code because of this.

lib/irq.c
tmakatos: @jlevon do you mean that ` vfu_irq_trigger()` would also have to take the `index`?

tmakatos (commented):
@limiao-intel is the device you're implementing something you could share?

lib/irq.c
jlevon: https://github.com/nutanix/libvfio-user/issues/387

Yes. But if qemu client side is doing it, maybe that doesn't matter much.

jlevon:
@limiao-intel hey, are you planning on updating this for the code review comments above? Thanks.

limiao-intel:
@jlevon @tmakatos Hi, I update the patch according to your comments. Thanks!

lib/irq.c
```diff
@@ -189,7 +189,17 @@ irqs_set_data_none(vfu_ctx_t *vfu_ctx, struct vfio_irq_set *irq_set)
     eventfd_t val;
 
     for (i = irq_set->start; i < (irq_set->start + irq_set->count); i++) {
-        efd = vfu_ctx->irqs->efds[i];
+        switch (irq_set->index) {
```
tmakatos, line 192: Please convert this into a function and use it here and below.

lib/irq.c
```diff
@@ -236,28 +256,46 @@ static int
...
+            break;
+        case VFIO_PCI_REQ_IRQ_INDEX:
+            efd = &vfu_ctx->irqs->req_efd;
+            break;
+        default:
+            efd = &vfu_ctx->irqs->efds[i];
+            break;
+        }
+        if (*efd >= 0) {
+            if (close(*efd) == -1)
```
tmakatos, line 278: missing curly brackets

tmakatos (commented):
Please rebase on master.
Also, when addressing review comments, don't force push as we have to review the whole patch again.

Thanks

lib/irq.c
limiao-intel: ok

lib/irq.c
tmakatos: Yes, this is libvfio-user's coding style.

lib/irq.c
tmakatos: Not needed, just a separate commit which we'll squash when merging this PR.

lib/irq.c
limiao-intel: Ok.

lib/irq.c
```diff
@@ -322,6 +344,13 @@ device_set_irqs_validate(vfu_ctx_t *vfu_ctx, vfu_msg_t *msg)
         line = __LINE__;
         goto invalid;
     }
+    //count must be 0 or 1 for ERR/REQ
```
tmakatos, line 347: nit: space after `//`

tmakatos (commented):
Looks good, minor nit. Also, the commit message should be clearer, e.g. what is broken and how is it being fixed.

lib/irq.c
```diff
@@ -180,19 +180,32 @@ irqs_set_state(vfu_ctx_t *vfu_ctx, struct vfio_irq_set *irq_set)
     cb(vfu_ctx, irq_set->start, irq_set->count, mask);
 }
 
+static inline int*
```
jlevon, line 183: "int *" please. Also no need for "inline" keyword.

lib/irq.c
```diff
@@ -322,6 +344,13 @@ device_set_irqs_validate(vfu_ctx_t *vfu_ctx, vfu_msg_t *msg)
         line = __LINE__;
         goto invalid;
     }
+    //count must be 0 or 1 for ERR/REQ
+    if (((irq_set->index == VFIO_PCI_ERR_IRQ_INDEX) ||
+         (irq_set->index == VFIO_PCI_REQ_IRQ_INDEX)) &&
```
jlevon, line 349: thanks for adding this, but I think you can definitely add a pytest for this case, let us know if it's not obvious?

jlevon (commented):
thanks, looking good, just a couple of minor comments.

lib/irq.c
limiao-intel: ok

lib/irq.c
limiao-intel: ok

lib/irq.c
limiao-intel: ok, I will add it in test_device_set_irqs.py.

jlevon (approved):
thanks!

PR #735, opened 2023-05-19
test: don't leave global pointing to stack memory

jimharris:
test_device_is_stopped_and_copying points the global vfu_ctx structure to a local stack-allocated data
structure.  This is fine while the function is
executing, but newer gcc complains that the
pointer is left there after it returns.

So clear the pointer to NULL before returning.

Reported-by: Kamil Godzwon <kamilx.godzwon@intel.com>

### DIFF
--- a/test/unit-tests.c
+++ b/test/unit-tests.c
@@ -601,6 +601,7 @@ test_device_is_stopped_and_copying(UNUSED void **state)
             assert_false(r);
         }
     }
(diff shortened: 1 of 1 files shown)


jimharris:
@jlevon @tmakatos when might this patch be merged?  This is just one of a few blockers to get SPDK CI using Fedora 38.

PR #737, opened 2023-06-17
Fix some python tests

brvtalcake:
Some of the tests didn't work, due to small syntax errors (at least for the last one). The second test didn't work either, but I don't know why, I didn't really dig further.

### DIFF
--- a/test/py/test_device_get_region_info.py
+++ b/test/py/test_device_get_region_info.py
@@ -128,7 +128,7 @@ def test_device_get_region_info_larger_argsz():
 
     result = msg(ctx, sock, VFIO_USER_DEVICE_GET_REGION_INFO, payload)
 
-    assert(len(result) == argsz)
+    assert len(result) == argsz
 
     info, _ = vfio_region_info.pop_from_buffer(result)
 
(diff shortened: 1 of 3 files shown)


tmakatos:
The current syntax is correct, can you explain what you mean by "Some of the tests didn't work, due to small syntax errors"?

brvtalcake:
I do not remember the errors (and I'm not currently at home), but when running the tests python3 was unhappy about asserts with parentheses (I think I have 3.11). I do not know python to be honest, but removing the parentheses solved the problem for one of the tests (maybe python was actually not understanding why parentheses were attached to the asserts, thus corresponding to a function call ?). The only thing I can say for sure is that wasn't an assert failure, but a syntax error.

I'll be soon at home (≈ 45 minutes) to show you the error logs

brvtalcake:
Here are the logs :

tmakatos:
This seems legitimate, I think it's more related to your flake8 version rather than Python itself, can you state both versions please?

brvtalcake:
My flake8 version is 5.0.4 (mccabe: 0.7.0, pycodestyle: 2.10.0, pyflakes: 2.5.0)

brvtalcake:
Oh actually yes it's for sure a flake8 "error". Then my bad I really thought that was python :/

PR #740, opened 2023-06-30
Fix address calculation for message-based DMA

mnissler-rivos:
The correct DMA address is formed by adding base and offset - the latter was accidentally missing.

### DIFF
--- a/lib/libvfio-user.c
+++ b/lib/libvfio-user.c
@@ -2198,7 +2198,7 @@ vfu_dma_transfer(vfu_ctx_t *vfu_ctx, enum vfio_user_command cmd,
     while (remaining > 0) {
         int ret;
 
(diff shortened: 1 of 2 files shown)


jlevon (commented):
Oops, thanks. It's a shame we don't have any python tests for this, but a bit unreasonable to ask you to add them.

However, looks like we could easily change samples server/client to uncover this bug and confirm the fix.

Would you mind adding that? Basically changing do_dma_io()?

Note that you'd need a manual run of the client/server test, it's currently disabled from CI due to a timing issue. It does usually pass though.

mnissler-rivos:
Updated the test and verified that it fails without and passes with the fix.

jlevon (approved):
thank you!

jlevon:
@mnissler-rivos sorry - do you mind adding Signed-off-by ?

PR #741, opened 2023-07-06
Adapt to new (v2) VFIO migration interface

### LINKED ISSUE #654: adapt to new (v2) VFIO migration interface
Alex Williamson says:

> +
> +The only device-specific region type and subtype supported by vfio-user is
> +``VFIO_REGION_TYPE_MIGRATION`` (3) and ``VFIO_REGION_SUBTYPE_MIGRATION`` (1).

These should be considered deprecated from the kernel interface.  I
hope there are plans for vfio-user to adopt the new interface that's
currently available in linux-next and intended for v5.18.

Which is this:

```
commit b042b27868c00142da1a7e31f4740a5fbd5628ed
Merge: cfb92440ee71 88faa5e8ead6
Author: Alex Williamson <alex.williamson@redhat.com>
Date:   Thu Mar 3 09:51:03 2022 -0700

    Merge tag 'mlx5-vfio-v10' of https://git.kernel.org/pub/scm/linux/kernel/git/mellanox/linux into v5.18/vfio/next/mlx5-migration-v10

    Add mlx5 live migration driver and v2 migration protocol
```

Need to investigate and possibly adopt these changes.

w-henderson:
This PR will (when done):

- [x] Update to version of the spec as submitted to [qemu-devel](https://lists.nongnu.org/archive/html/qemu-devel/2023-06/msg06567.html)
- [x] Add new migration protocol based on VFIO migration v2, replacing the migration region with a more intuitive stream-like mechanism for transferring migration data (alongside other changes)
- [ ] Implement the new protocol in libvfio-user
- [ ] Update and add more tests for the new protocol
- [ ] Update the samples for the new protocol

Closes #654

### DIFF
--- a/docs/vfio-user.rst
+++ b/docs/vfio-user.rst
@@ -4,7 +4,7 @@ vfio-user Protocol Specification
 ********************************
 
 --------------
-Version_ 0.9.1
+Version_ 0.9.2
 --------------
 
 .. contents:: Table of Contents
@@ -366,6 +366,9 @@ Name                                    Command    Request Direction
 ``VFIO_USER_DMA_WRITE``                 12         server -> client
 ``VFIO_USER_DEVICE_RESET``              13         client -> server
 ``VFIO_USER_REGION_WRITE_MULTI``        15         client -> server
+``VFIO_USER_DEVICE_FEATURE``            16         client -> server
+``VFIO_USER_MIG_DATA_READ``             17         client -> server
+``VFIO_USER_MIG_DATA_WRITE``            18         client -> server
 ======================================  =========  =================
 
 Header
@@ -515,18 +518,7 @@ Capabilities:
 |                    |         | are supported if the value is ``true``.        |
 +--------------------+---------+------------------------------------------------+
 
-The migration capability contains the following name/value pairs:
-
-+-----------------+--------+--------------------------------------------------+
-| Name            | Type   | Description                                      |
-+=================+========+==================================================+
-| pgsize          | number | Page size of dirty pages bitmap. The smallest    |
-|                 |        | between the client and the server is used.       |
-+-----------------+--------+--------------------------------------------------+
-| max_bitmap_size | number | Maximum bitmap size in ``VFIO_USER_DIRTY_PAGES`` |
-|                 |        | and ``VFIO_DMA_UNMAP`` messages.  Optional,      |
-|                 |        | with a default value of 256MB.                   |
-+-----------------+--------+--------------------------------------------------+
+The migration capability is currently an empty object.
 
 Reply
 ^^^^^
@@ -1468,6 +1460,292 @@ Reply
 
 * *wr_cnt* is the number of device writes completed.
 
+``VFIO_USER_DEVICE_FEATURE``
+----------------------------
+
+This command is analogous to ``VFIO_DEVICE_FEATURE``. It is used to get, set, or
+probe feature data of the device.
+
+Request
+^^^^^^^
+
+The request payload for this message is a structure of the following format.
+
++-------+--------+--------------------------------+
+| Name  | Offset | Size                           |
++=======+========+================================+
+| argsz | 0      | 4                              |
++-------+--------+--------------------------------+
+| flags | 4      | 4                              |
++-------+--------+--------------------------------+
+|       | +---------+---------------------------+ |
+|       | | Bit     | Definition                | |
+|       | +=========+===========================+ |
+|       | | 0 to 15 | VFIO_DEVICE_FEATURE_MASK  | |
+|       | +---------+---------------------------+ |
+|       | | 16      | VFIO_DEVICE_FEATURE_GET   | |
+|       | +---------+---------------------------+ |
+|       | | 17      | VFIO_DEVICE_FEATURE_SET   | |
+|       | +---------+---------------------------+ |
+|       | | 18      | VFIO_DEVICE_FEATURE_PROBE | |
+|       | +---------+---------------------------+ |
++-------+--------+--------------------------------+
+| data  | 8      | variable                       |
++-------+--------+--------------------------------+
+
+* *argsz* is the size of the above structure.
+
+* *flags* defines the action to be performed by the server and upon which
+  feature:
+
+  * ``VFIO_DEVICE_FEATURE_MASK`` is the 16-bit feature index of the relevant
+    feature.
+  
+  * ``VFIO_DEVICE_FEATURE_GET`` instructs the server to get the feature data.
+
+  * ``VFIO_DEVICE_FEATURE_SET`` instructs the server to set the feature data to
+    that given in the ``data`` field of the payload.
+
+  * ``VFIO_DEVICE_FEATURE_PROBE`` instructs the server to probe for feature
+    support. If ``VFIO_DEVICE_FEATURE_GET/SET`` are also set, the probe will
+    only return success if all of the indicated methods are supported.
+
+  ``VFIO_DEVICE_FEATURE_GET`` and ``VFIO_DEVICE_FEATURE_SET`` are mutually
+  exclusive, except for use with ``VFIO_DEVICE_FEATURE_PROBE``.
+
+* *data* is specific to the particular feature. It is not used for probing.
+
+This part of the request is analogous to VFIO's ``struct vfio_device_feature``.
+
+Reply
+^^^^^
+
+The reply payload must be the same as the request payload for setting or
+probing a feature. For getting a feature's data, the data is added in the data
+section and its length is added to ``argsz``.
+
+Device Features
+^^^^^^^^^^^^^^^
+
+The only device features supported by vfio-user are those related to migration.
+
+``VFIO_DEVICE_FEATURE_MIGRATION``
+"""""""""""""""""""""""""""""""""
+
+This feature indicates that the device can support the migration API through
+``VFIO_DEVICE_FEATURE_MIG_DEVICE_STATE``. If ``GET`` succeeds, the ``RUNNING``
+and ``ERROR`` states are always supported. Support for additional states is
+indicated via the flags field; at least ``VFIO_MIGRATION_STOP_COPY`` must be
+set.
+
+There is no data field of the request message.
+
+The data field of the reply message is structured as follows:
+
++-------+--------+---------------------------+
+| Name  | Offset | Size                      |
++=======+========+===========================+
+| flags | 0      | 8                         |
++-------+--------+---------------------------+
+|       | +-----+--------------------------+ |
+|       | | Bit | Definition               | |
+|       | +=====+==========================+ |
+|       | | 0   | VFIO_MIGRATION_STOP_COPY | |
+|       | +-----+--------------------------+ |
+|       | | 1   | VFIO_MIGRATION_P2P       | |
+|       | +-----+--------------------------+ |
+|       | | 2   | VFIO_MIGRATION_PRE_COPY  | |
+|       | +-----+--------------------------+ |
++-------+--------+---------------------------+
+
+``VFIO_DEVICE_FEATURE_MIG_DEVICE_STATE``
+""""""""""""""""""""""""""""""""""""""""
+
+Upon ``VFIO_DEVICE_FEATURE_SET``, execute a migration state change on the VFIO
+device. The new state is supplied in ``device_state``. The state transition must
+fully complete before the reply is sent.
+
+The data field of the reply message, as well as the ``SET`` request message, is
+structured as follows:
+
++--------------+--------+------+
+| Name         | Offset | Size |
++==============+========+======+
+| device_state | 0      | 4    |
++--------------+--------+------+
+| data_fd      | 4      + 4    |
++--------------+--------+------+
+
+* *device_state* is the current state of the device (for ``GET``) or the
+  state to transition to (for ``SET``). It is defined by the
+  ``vfio_device_mig_state`` enum as detailed below. These states are the states
+  of the device migration Finite State Machine.
+
++--------------------------------+-------+---------------------------------------------------------------------+
+| Name                           | State | Description                                                         |
++================================+=======+=====================================================================+
+| VFIO_DEVICE_STATE_ERROR        | 0     | The device has failed and must be reset.                            |
++--------------------------------+-------+---------------------------------------------------------------------+
+| VFIO_DEVICE_STATE_STOP         | 1     | The device does not change the internal or external state.          |
++--------------------------------+-------+---------------------------------------------------------------------+
+| VFIO_DEVICE_STATE_RUNNING      | 2     | The device is running normally.                                     |
++--------------------------------+-------+---------------------------------------------------------------------+
+| VFIO_DEVICE_STATE_STOP_COPY    | 3     | The device internal state can be read out.                          |
++--------------------------------+-------+---------------------------------------------------------------------+
+| VFIO_DEVICE_STATE_RESUMING     | 4     | The device is stopped and is loading a new internal state.          |
++--------------------------------+-------+---------------------------------------------------------------------+
+| VFIO_DEVICE_STATE_RUNNING_P2P  | 5     | (not used in vfio-user)                                             |
++--------------------------------+-------+---------------------------------------------------------------------+
+| VFIO_DEVICE_STATE_PRE_COPY     | 6     | The device is running normally but tracking internal state changes. |
++--------------------------------+-------+---------------------------------------------------------------------+
+| VFIO_DEVICE_STATE_PRE_COPY_P2P | 7     | (not used in vfio-user)                                             |
++--------------------------------+-------+---------------------------------------------------------------------+
+
+* *data_fd* is unused in vfio-user, as the ``VFIO_USER_MIG_DATA_READ`` and
+  ``VFIO_USER_MIG_DATA_WRITE`` messages are used instead for migration data
+  transport.
+
+Direct State Transitions
+""""""""""""""""""""""""
+
+The device migration FSM is a Mealy machine, so actions are taken upon the arcs
+between FSM states. The following transitions need to be supported by the
+server, a subset of those defined in ``<linux/vfio.h>``
+(``enum vfio_device_mig_state``).
+
+* ``RUNNING -> STOP``, ``STOP_COPY -> STOP``: Stop the operation of the device. The ``STOP_COPY`` arc terminates the data transfer session.
+
+* ``RESUMING -> STOP``: Terminate the data transfer session. Complete processing
+  of the migration data. Stop the operation of the device. If the delivered data
+  is found to be incomplete, inconsistent, or otherwise invalid, fail the
+  ``SET`` command and optionally transition to the ``ERROR`` state.
+
+* ``PRE_COPY -> RUNNING``: Terminate the data transfer session. The device is
+  now fully operational.
+
+* ``STOP -> RUNNING``: Start the operation of the device.
+
+* ``RUNNING -> PRE_COPY``, ``STOP -> STOP_COPY``: Begin the process of saving
+  the device state. The device operation is unchanged, but data transfer begins.
+  ``PRE_COPY`` and ``STOP_COPY`` are referred to as the "saving group" of
+  states.
+
+* ``PRE_COPY -> STOP_COPY``: Continue to transfer migration data, but stop 
+  device operation.
+
+* ``STOP -> RESUMING``: Start the process of restoring the device state. The
+  internal device state may be changed to prepare the device to receive the
+  migration data.
+
+The ``STOP_COPY -> PRE_COPY`` transition is explicitly not allowed and should
+return an error if requested.
+
+``ERROR`` cannot be specified as a device state, but any transition request can
+be failed and then move the state into ``ERROR`` if the server was unable to
+execute the requested arc AND was unable to restore the device into any valid
+state. To recover from ``ERROR``, ``VFIO_USER_DEVICE_RESET`` must be used to
+return back to ``RUNNING``.
+
+If ``PRE_COPY`` is not supported, arcs touching it are removed.
+
+Complex State Transitions
+"""""""""""""""""""""""""
+
+The remaining possible transitions are to be implemented as combinations of the
+above FSM arcs. As there are multiple paths, the path should be selected based
+on the following rules:
+
+* Select the shortest path.
+
+* The path cannot have saving group states as interior arcs, only start/end
+  states.
+
+``VFIO_USER_MIG_DATA_READ``
+---------------------------
+
+This command is used to read data from the source migration server while it is
+in a saving group state (``PRE_COPY`` or ``STOP_COPY``).
+
+This command, and ``VFIO_USER_MIG_DATA_WRITE``, are used in place of the
+``data_fd`` file descriptor in ``<linux/vfio.h>``
+(``struct vfio_device_feature_mig_state``) to enable all data transport to use
+the single already-established UNIX socket. Hence, the migration data is
+treated like a stream, so the client must continue reading until no more
+migration data remains.
+
+Request
+^^^^^^^
+
+The request payload for this message is a structure of the following format.
+
++-------+--------+------+
+| Name  | Offset | Size |
++=======+========+======+
+| argsz | 0      | 4    |
++-------+--------+------+
+| size  | 4      | 8    |
++-------+--------+------+
+
+* *argsz* is the size of the above structure.
+
+* *size* is the size of the migration data to read.
+
+Reply
+^^^^^
+
+The reply payload for this message is a structure of the following format.
+
++-------+--------+----------+
+| Name  | Offset | Size     |
++=======+========+==========+
+| argsz | 0      | 4        |
++-------+--------+----------+
+| size  | 4      | 8        |
++-------+--------+----------+
+| data  | 12     | variable |
++-------+--------+----------+
+
+* *argsz* is the size of the above structure.
+
+* *size* indicates the size of returned migration data. If this is less than the
+  requested size, there is no more migration data to read.
+
+* *data* contains the migration data.
+
+``VFIO_USER_MIG_DATA_WRITE``
+----------------------------
+
+This command is used to write data to the destination migration server while it
+is in the ``RESUMING`` state.
+
+As above, this replaces the ``data_fd`` file descriptor for transport of
+migration data, and as such, the migration data is treated like a stream.
+
+Request
+^^^^^^^
+
+The request payload for this message is a structure of the following format.
+
++-------+--------+----------+
+| Name  | Offset | Size     |
++=======+========+==========+
+| argsz | 0      | 4        |
++-------+--------+----------+
+| size  | 4      | 8        |
++-------+--------+----------+
+| data  | 12     | variable |
++-------+--------+----------+
+
+* *argsz* is the size of the above structure.
+
+* *size* is the size of the migration data to be written.
+
+* *data* contains the migration data.
+
+Reply
+^^^^^
+
+There is no reply payload for this message.
 
 Appendices
 ==========
--- a/docs/vfio-user.rst
+++ b/docs/vfio-user.rst
@@ -1528,6 +1528,22 @@ Device Features
 ^^^^^^^^^^^^^^^
 
 The only device features supported by vfio-user are those related to migration.
+They are a subset of those supported in the VFIO implementation of the Linux
+kernel.
+
++----------------------------------------+-------+
+| Name                                   | Value |
++========================================+=======+
+| VFIO_DEVICE_FEATURE_MIGRATION          | 1     |
++----------------------------------------+-------+
+| VFIO_DEVICE_FEATURE_MIG_DEVICE_STATE   | 2     |
++----------------------------------------+-------+
+| VFIO_DEVICE_FEATURE_DMA_LOGGING_START  | 6     |
++----------------------------------------+-------+
+| VFIO_DEVICE_FEATURE_DMA_LOGGING_STOP   | 7     |
++----------------------------------------+-------+
+| VFIO_DEVICE_FEATURE_DMA_LOGGING_REPORT | 8     |
++----------------------------------------+-------+
 
 ``VFIO_DEVICE_FEATURE_MIGRATION``
 """""""""""""""""""""""""""""""""
@@ -1604,6 +1620,91 @@ structured as follows:
 * *data_fd* is unused in vfio-user, as the ``VFIO_USER_MIG_DATA_READ`` and
   ``VFIO_USER_MIG_DATA_WRITE`` messages are used instead for migration data
   transport.
+  
+``VFIO_DEVICE_FEATURE_DMA_LOGGING_START`` / ``VFIO_DEVICE_FEATURE_DMA_LOGGING_STOP``
+""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""
+
+Upon ``VFIO_DEVICE_FEATURE_SET``, start/stop DMA logging. These features can
+also be probed to determine whether the device supports DMA logging.
+
+When DMA logging is started, a range of IOVAs to monitor is provided and the
+device can optimize its logging to cover only the IOVA range given. Only DMA
+writes are logged.
+
+The data field of the ``SET`` request is structured as follows:
+
++------------+--------+----------+
+| Name       | Offset | Size     |
++============+========+==========+
+| page_size  | 0      | 8        |
++------------+--------+----------+
+| num_ranges | 8      | 4        |
++------------+--------+----------+
+| reserved   | 12     | 4        |
++------------+--------+----------+
+| ranges     | 16     | variable |
++------------+--------+----------+
+
+* *page_size* hints what tracking granularity the device should try to achieve.
+  If the device cannot do the hinted page size then it's the driver's choice
+  which page size to pick based on its support. On output the device will return
+  the page size it selected.
+
+* *num_ranges* is the number of IOVA ranges to monitor. A value of zero
+  indicates that all writes should be logged.
+
+* *ranges* is an array of ``vfio_user_device_feature_dma_logging_range``
+  entries:
+
++--------+--------+------+
+| Name   | Offset | Size |
++========+========+======+
+| iova   | 0      | 8    |
++--------+--------+------+
+| length | 8      | 8    |
++--------+--------+------+
+
+  * *iova* is the base IO virtual address
+  * *length* is the length of the range to log
+
+Upon success, the response data field will be the same as the request, unless
+the page size was changed, in which case this will be reflected in the response.
+
+``VFIO_DEVICE_FEATURE_DMA_LOGGING_REPORT``
+""""""""""""""""""""""""""""""""""""""""""
+
+Upon ``VFIO_DEVICE_FEATURE_GET``, returns the dirty bitmap for a specific IOVA
+range. This operation is only valid if logging of dirty pages has been
+previously started by setting ``VFIO_DEVICE_FEATURE_DMA_LOGGING_START``.
+
+The data field of the request is structured as follows:
+
++-----------+--------+------+
+| Name      | Offset | Size |
++===========+========+======+
+| iova      | 0      | 8    |
++-----------+--------+------+
+| length    | 8      | 8    |
++-----------+--------+------+
+| page_size | 16     | 8    |
++-----------+--------+------+
+
+* *iova* is the base IO virtual address
+
+* *length* is the length of the range
+
+* *page_size* is the unit of granularity of the bitmap, and must be a power of
+  two. It doesn't have to match the value given to
+  ``VFIO_DEVICE_FEATURE_DMA_LOGGING_START`` because the driver will format its
+  internal logging to match the reporting page size possibly by replicating bits
+  if the internal page size is lower than requested
+
+The data field of the response is identical, except with the bitmap added on
+the end at offset 24.
+
+The mapping of IOVA to bits is given by:
+
+``bitmap[(addr - iova)/page_size] & (1ULL << (addr % 64))``
 
 Direct State Transitions
 """"""""""""""""""""""""
--- a/docs/vfio-user.rst
+++ b/docs/vfio-user.rst
@@ -1,5 +1,4 @@
 .. include:: <isonum.txt>
-
 ********************************
 vfio-user Protocol Specification
 ********************************
@@ -322,9 +321,9 @@ usual ``msg_size`` field in the header, not the ``argsz`` field.
 
 In a reply, the server sets ``argsz`` field to the size needed for a full
 payload size. This may be less than the requested maximum size. This may be
-larger than the requested maximum size: in that case, the payload reply header
-is returned, but the ``argsz`` field in the reply indicates the needed size,
-allowing a client to allocate a larger buffer for holding the reply before
+larger than the requested maximum size: in that case, the full payload is not
+included in the reply, but the ``argsz`` field in the reply indicates the needed
+size, allowing a client to allocate a larger buffer for holding the reply before
 trying again.
 
 In addition, during negotiation (see  `Version`_), the client and server may
@@ -337,8 +336,9 @@ Protocol Specification
 
 To distinguish from the base VFIO symbols, all vfio-user symbols are prefixed
 with ``vfio_user`` or ``VFIO_USER``. In this revision, all data is in the
-little-endian format, although this may be relaxed in future revisions in cases
-where the client and server are both big-endian.
+endianness of the host system, although this may be relaxed in future
+revisions in cases where the client and server run on different hosts
+with different endianness.
 
 Unless otherwise specified, all sizes should be presumed to be in bytes.
 
@@ -365,7 +365,7 @@ Name                                    Command    Request Direction
 ``VFIO_USER_DMA_READ``                  11         server -> client
 ``VFIO_USER_DMA_WRITE``                 12         server -> client
 ``VFIO_USER_DEVICE_RESET``              13         client -> server
-``VFIO_USER_DIRTY_PAGES``               14         client -> server
+``VFIO_USER_REGION_WRITE_MULTI``        15         client -> server
 ======================================  =========  =================
 
 Header
@@ -488,30 +488,45 @@ format:
 
 Capabilities:
 
-+--------------------+--------+------------------------------------------------+
-| Name               | Type   | Description                                    |
-+====================+========+================================================+
-| max_msg_fds        | number | Maximum number of file descriptors that can be |
-|                    |        | received by the sender in one message.         |
-|                    |        | Optional. If not specified then the receiver   |
-|                    |        | must assume a value of ``1``.                  |
-+--------------------+--------+------------------------------------------------+
-| max_data_xfer_size | number | Maximum ``count`` for data transfer messages;  |
-|                    |        | see `Read and Write Operations`_. Optional,    |
-|                    |        | with a default value of 1048576 bytes.         |
-+--------------------+--------+------------------------------------------------+
-| migration          | object | Migration capability parameters. If missing    |
-|                    |        | then migration is not supported by the sender. |
-+--------------------+--------+------------------------------------------------+
++--------------------+---------+------------------------------------------------+
+| Name               | Type    | Description                                    |
++====================+=========+================================================+
+| max_msg_fds        | number  | Maximum number of file descriptors that can be |
+|                    |         | received by the sender in one message.         |
+|                    |         | Optional. If not specified then the receiver   |
+|                    |         | must assume a value of ``1``.                  |
++--------------------+---------+------------------------------------------------+
+| max_data_xfer_size | number  | Maximum ``count`` for data transfer messages;  |
+|                    |         | see `Read and Write Operations`_. Optional,    |
+|                    |         | with a default value of 1048576 bytes.         |
++--------------------+---------+------------------------------------------------+
+| pgsizes            | number  | Page sizes supported in DMA map operations     |
+|                    |         | or'ed together. Optional, with a default value |
+|                    |         | of supporting only 4k pages.                   |
++--------------------+---------+------------------------------------------------+
+| max_dma_maps       | number  | Maximum number DMA map windows that can be     |
+|                    |         | valid simultaneously.  Optional, with a        |
+|                    |         | value of 65535 (64k-1).                        |
++--------------------+---------+------------------------------------------------+
+| migration          | object  | Migration capability parameters. If missing    |
+|                    |         | then migration is not supported by the sender. |
++--------------------+---------+------------------------------------------------+
+| write_multiple     | boolean | ``VFIO_USER_REGION_WRITE_MULTI`` messages      |
+|                    |         | are supported if the value is ``true``.        |
++--------------------+---------+------------------------------------------------+
 
 The migration capability contains the following name/value pairs:
 
-+--------+--------+-----------------------------------------------+
-| Name   | Type   | Description                                   |
-+========+========+===============================================+
-| pgsize | number | Page size of dirty pages bitmap. The smallest |
-|        |        | between the client and the server is used.    |
-+--------+--------+-----------------------------------------------+
++-----------------+--------+--------------------------------------------------+
+| Name            | Type   | Description                                      |
++=================+========+==================================================+
+| pgsize          | number | Page size of dirty pages bitmap. The smallest    |
+|                 |        | between the client and the server is used.       |
++-----------------+--------+--------------------------------------------------+
+| max_bitmap_size | number | Maximum bitmap size in ``VFIO_USER_DIRTY_PAGES`` |
+|                 |        | and ``VFIO_DMA_UNMAP`` messages.  Optional,      |
+|                 |        | with a default value of 256MB.                   |
++-----------------+--------+--------------------------------------------------+
 
 Reply
 ^^^^^
@@ -607,56 +622,18 @@ The request payload for this message is a structure of the following format:
 +--------------+--------+------------------------+
 | flags        | 4      | 4                      |
 +--------------+--------+------------------------+
-|              | +-----+-----------------------+ |
-|              | | Bit | Definition            | |
-|              | +=====+=======================+ |
-|              | | 0   | get dirty page bitmap | |
-|              | +-----+-----------------------+ |
-|              | | 1   | unmap all regions     | |
-|              | +-----+-----------------------+ |
-+--------------+--------+------------------------+
 | address      | 8      | 8                      |
 +--------------+--------+------------------------+
 | size         | 16     | 8                      |
 +--------------+--------+------------------------+
 
 * *argsz* is the maximum size of the reply payload.
-* *flags* contains the following DMA region attributes:
-
-  * *get dirty page bitmap* indicates that a dirty page bitmap must be
-    populated before unmapping the DMA region. The client must provide a
-    `VFIO Bitmap`_ structure, explained below, immediately following this
-    entry.
-  * *unmap all regions* indicates to unmap all the regions previously
-    mapped via `VFIO_USER_DMA_MAP`. This flag cannot be combined with
-    *get dirty page bitmap* and expects *address* and *size* to be 0.
-
+* *flags* is unused in this version.
 * *address* is the base DMA address of the DMA region.
 * *size* is the size of the DMA region.
 
 The address and size of the DMA region being unmapped must match exactly a
-previous mapping. The size of request message depends on whether or not the
-*get dirty page bitmap* bit is set in Flags:
-
-* If not set, the size of the total request message is: 16 + 24.
-
-* If set, the size of the total request message is: 16 + 24 + 16.
-
-.. _VFIO Bitmap:
-
-VFIO Bitmap Format
-""""""""""""""""""
-
-+--------+--------+------+
-| Name   | Offset | Size |
-+========+========+======+
-| pgsize | 0      | 8    |
-+--------+--------+------+
-| size   | 8      | 8    |
-+--------+--------+------+
-
-* *pgsize* is the page size for the bitmap, in bytes.
-* *size* is the size for the bitmap, in bytes, excluding the VFIO bitmap header.
+previous mapping.
 
 Reply
 ^^^^^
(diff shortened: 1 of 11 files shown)


jlevon:
Hi William, thanks! Looks like a great start.
Let's split things up from your branch into individual PRs - this will let us review each logical change on its own. Some context here: https://movementarian.org/blog/posts/github-prs/

As for the first change here, this looks good, but could you add some context to the commit message? In particular:
 - mention *where* you're syncing from
 - explain that while you're removing LM in the spec, the code will remain in the library until updated to v2 by later changes
 - mention that we don't implement all of the other new changes in the spec update - but we are otherwise compatible
 - as a useful side quest, would be good to file issues on any of those changes. For example we already have a PR for multi-write, but we definitely don't have issues filed for some of the new capabilities during negotiation

include/vfio-user.h
```diff
@@ -67,6 +67,7 @@ enum vfio_user_command {
     VFIO_USER_DMA_WRITE                 = 12,
     VFIO_USER_DEVICE_RESET              = 13,
     VFIO_USER_DIRTY_PAGES               = 14,
+    VFIO_USER_DEVICE_FEATURE            = 15,
```
jlevon: In your commit message, it'll be helpful if you can describe where we differ from vfio, if we do, and why, so we can understand context a little more.

lib/libvfio-user.c
```diff
@@ -1052,6 +1052,45 @@ handle_dirty_pages(vfu_ctx_t *vfu_ctx, vfu_msg_t *msg)
...
+static int
+handle_device_feature(vfu_ctx_t *vfu_ctx, vfu_msg_t *msg)
+{
+    struct vfio_user_device_feature *req = msg->in.iov.iov_base;
+
+    if (vfu_ctx->migration == NULL) {
+        return -EINVAL;
+    }
+
+    if (!migration_feature_supported(req->flags)) {
```
jlevon, line 1037: Is it possible for msg->in.iov.iov_len to be smaller than the size of vfio_user_device_feature?

lib/migration.c
```diff
@@ -534,6 +534,60 @@ migration_region_access(vfu_ctx_t *vfu_ctx, char *buf, size_t count,
     return count;
 }
 
+bool
+migration_feature_supported(uint32_t flags) {
+    switch (flags & VFIO_DEVICE_FEATURE_MASK) {
+        case VFIO_DEVICE_FEATURE_MIGRATION:
+            return !(flags & VFIO_DEVICE_FEATURE_SET); // not supported for set
+        case VFIO_DEVICE_FEATURE_MIG_DEVICE_STATE:
```
jlevon, line 179: what does it mean to set this flag.

lib/libvfio-user.c
```diff
@@ -1052,6 +1052,45 @@ handle_dirty_pages(vfu_ctx_t *vfu_ctx, vfu_msg_t *msg)
...
+        // FIXME what error code to return? we really want "not supported"
+        //   instead of "not permitted"?
+        return -EINVAL;
+    }
+
+    ssize_t ret;
+
+    if (req->flags & VFIO_DEVICE_FEATURE_PROBE) {
```
jlevon, line 1045: presumably it's invalid to set any other flag its in this case; should you check for that?

lib/libvfio-user.c
```diff
@@ -1052,6 +1052,45 @@ handle_dirty_pages(vfu_ctx_t *vfu_ctx, vfu_msg_t *msg)
...
+        msg->out.iov.iov_base = msg->in.iov.iov_base;
```
jlevon: This implementation doesn't make much sense to me: it has no effect?

lib/migration.c
```diff
@@ -534,6 +534,60 @@ migration_region_access(vfu_ctx_t *vfu_ctx, char *buf, size_t count,
     return count;
 }
 
+bool
+migration_feature_supported(uint32_t flags) {
```
jlevon, line 175: migration.c shouldn't be dealing with DEVICE_FEATURE encoding: you should pass in the masked out flags section only.

lib/migration.c
```diff
@@ -534,6 +534,60 @@ migration_region_access(vfu_ctx_t *vfu_ctx, char *buf, size_t count,
...
+    switch (flags & VFIO_DEVICE_FEATURE_MASK) {
+        case VFIO_DEVICE_FEATURE_MIGRATION:
+            return !(flags & VFIO_DEVICE_FEATURE_SET); // not supported for set
```
jlevon, line 178: and this should either be handled in migration_feature_set(), or the handler above.

lib/migration.c
```diff
@@ -534,6 +534,60 @@ migration_region_access(vfu_ctx_t *vfu_ctx, char *buf, size_t count,
...
+    };
+}
+
+ssize_t
+migration_feature_get(vfu_ctx_t *vfu_ctx, uint32_t flags, void *buf)
+{
+    struct vfio_user_device_feature_migration *res;
+    struct vfio_user_device_feature_mig_state *state;
+
+    switch (flags & VFIO_DEVICE_FEATURE_MASK) {
```
jlevon, line 192: same comment re the mask, caller should do this.

lib/libvfio-user.c
```diff
@@ -1052,6 +1052,45 @@ handle_dirty_pages(vfu_ctx_t *vfu_ctx, vfu_msg_t *msg)
...
+        msg->out.iov.iov_len = msg->in.iov.iov_len;
+
+        ret = 0;
+    } else if (req->flags & VFIO_DEVICE_FEATURE_GET) {
+        msg->out.iov.iov_base = calloc(8, 1);
```
jlevon, line 1053: 1, 8 is more typical. You don't check for allocation failure.
If you set iov_len first, you can then use that value and not erpeat yourself.
Finally, I'd use sizeof(vfio_user_device_feature_migration), with a comment that currently all supported outgoing data is the same size as this struct (8 bytes).

lib/libvfio-user.c
```diff
@@ -1052,6 +1052,45 @@ handle_dirty_pages(vfu_ctx_t *vfu_ctx, vfu_msg_t *msg)
...
+        msg->out.iov.iov_len = 8;
+
+        ret = migration_feature_get(vfu_ctx, req->flags,
+                                    msg->out.iov.iov_base);
+    } else if (req->flags & VFIO_DEVICE_FEATURE_SET) {
+        msg->out.iov.iov_base = msg->in.iov.iov_base;
```
jlevon: is it in the spec to return the data back in this case??

include/vfio-user.h
```diff
@@ -228,6 +229,44 @@ struct vfio_user_bitmap_range {
...
+	uint8_t  	data[];
+} __attribute__((packed));
+
+/* Analogous to vfio_device_feature_migration */
+struct vfio_user_device_feature_migration {
+    uint64_t flags;
+#define VFIO_MIGRATION_STOP_COPY    (1 << 0)
+#define VFIO_MIGRATION_P2P          (1 << 1)
+#define VFIO_MIGRATION_PRE_COPY     (1 << 2)
+} __attribute__((packed));
```
jlevon, line 245: since we require this to be 8-bytes in size, you should add a static assert.

include/vfio-user.h
```diff
@@ -228,6 +229,44 @@ struct vfio_user_bitmap_range {
...
+#define VFIO_DEVICE_FEATURE_MIGRATION 1
+
+/* Analogous to vfio_device_feature_mig_state */
+struct vfio_user_device_feature_mig_state {
+    uint32_t    device_state;
+    uint32_t    data_fd;
+} __attribute__((packed));
```
jlevon, line 252: ditto

lib/migration.c
```diff
@@ -534,6 +534,60 @@ migration_region_access(vfu_ctx_t *vfu_ctx, char *buf, size_t count,
...
+        case VFIO_DEVICE_FEATURE_MIGRATION:
+            res = buf;
+            // FIXME are these always supported? Can we consider to be
+            //   "supported" if said support is just an empty callback?
+            res->flags = VFIO_MIGRATION_STOP_COPY | VFIO_MIGRATION_PRE_COPY;
+
```
jlevon, line 198: mega-nit, un-needed blank line, and below.

lib/migration.c
```diff
@@ -534,6 +534,60 @@ migration_region_access(vfu_ctx_t *vfu_ctx, char *buf, size_t count,
...
+ssize_t
+migration_feature_set(UNUSED vfu_ctx_t *vfu_ctx, uint32_t flags,
+                      UNUSED void *buf)
+{
+    if (flags & VFIO_DEVICE_FEATURE_MIG_DEVICE_STATE) {
+        // struct vfio_user_device_feature_mig_state *res = buf;
+
+        // TODO migrate to res->device_state state if allowed to
+
+        return 0;
```
jlevon: no problem with phasing stuff, but if it doesn't do anything, let's make it return -ENOTSUP until later in the patch series.

lib/migration.c
```diff
@@ -534,6 +534,60 @@ migration_region_access(vfu_ctx_t *vfu_ctx, char *buf, size_t count,
...
+{
+    struct vfio_user_device_feature_migration *res;
+    struct vfio_user_device_feature_mig_state *state;
+
+    switch (flags & VFIO_DEVICE_FEATURE_MASK) {
+        case VFIO_DEVICE_FEATURE_MIGRATION:
+            res = buf;
+            // FIXME are these always supported? Can we consider to be
+            //   "supported" if said support is just an empty callback?
+            res->flags = VFIO_MIGRATION_STOP_COPY | VFIO_MIGRATION_PRE_COPY;
```
jlevon, line 197: Thanos?

test/py/libvfio_user.py
```diff
@@ -124,6 +124,16 @@
 VFIO_DEVICE_STATE_V1_RESUMING = (1 << 2)
 VFIO_DEVICE_STATE_MASK = ((1 << 3) - 1)
 
+VFIO_DEVICE_FEATURE_MIGRATION = (1)
```
jlevon, line 127: I didn't check - does it make sense to add tests here too, or too early/tricky?

docs/vfio-user.rst
```diff
@@ -1468,6 +1460,292 @@ Reply
...
+---------------------------
+
+This command is used to read data from the source migration server while it is
+in a saving group state (``PRE_COPY`` or ``STOP_COPY``).
+
+This command, and ``VFIO_USER_MIG_DATA_WRITE``, are used in place of the
+``data_fd`` file descriptor in ``<linux/vfio.h>``
+(``struct vfio_device_feature_mig_state``) to enable all data transport to use
+the single already-established UNIX socket. Hence, the migration data is
+treated like a stream, so the client must continue reading until no more
```
jlevon, line 1774: How does this look in vfio? Is a stream "safe", or would we ever want to support sparse read/writes in some way?

I'm not clear how this would even work if we wanted to do device state dirty tracking in PRE_COPY state.

docs/vfio-user.rst
```diff
@@ -515,18 +518,7 @@ Capabilities:
...
-| Name            | Type   | Description                                      |
-+=================+========+==================================================+
-| pgsize          | number | Page size of dirty pages bitmap. The smallest    |
-|                 |        | between the client and the server is used.       |
-+-----------------+--------+--------------------------------------------------+
-| max_bitmap_size | number | Maximum bitmap size in ``VFIO_USER_DIRTY_PAGES`` |
-|                 |        | and ``VFIO_DMA_UNMAP`` messages.  Optional,      |
-|                 |        | with a default value of 256MB.                   |
-+-----------------+--------+--------------------------------------------------+
+The migration capability is currently an empty object.
```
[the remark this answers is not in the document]
tmakatos: I don't think we need this anymore, migration is probed using the device features.

docs/vfio-user.rst
```diff
@@ -1468,6 +1460,292 @@ Reply
...
+  * ``VFIO_DEVICE_FEATURE_MASK`` is the 16-bit feature index of the relevant
+    feature.
+  
+  * ``VFIO_DEVICE_FEATURE_GET`` instructs the server to get the feature data.
+
+  * ``VFIO_DEVICE_FEATURE_SET`` instructs the server to set the feature data to
+    that given in the ``data`` field of the payload.
+
+  * ``VFIO_DEVICE_FEATURE_PROBE`` instructs the server to probe for feature
+    support. If ``VFIO_DEVICE_FEATURE_GET/SET`` are also set, the probe will
```
tmakatos, line 1510: nit: best to use full symbol name as it facilitates text searching

docs/vfio-user.rst
```diff
@@ -1468,6 +1460,292 @@ Reply
...
+^^^^^
+
+The reply payload must be the same as the request payload for setting or
+probing a feature. For getting a feature's data, the data is added in the data
+section and its length is added to ``argsz``.
+
+Device Features
+^^^^^^^^^^^^^^^
+
+The only device features supported by vfio-user are those related to migration.
```
tmakatos, line 1530: Not sure whether it's worth calrifying that this may change in the future?

docs/vfio-user.rst
```diff
@@ -1468,6 +1460,292 @@ Reply
...
+
+The data field of the reply message, as well as the ``SET`` request message, is
+structured as follows:
+
++--------------+--------+------+
+| Name         | Offset | Size |
++==============+========+======+
+| device_state | 0      | 4    |
++--------------+--------+------+
+| data_fd      | 4      + 4    |
```
tmakatos, line 1592: s/+/|

docs/vfio-user.rst
```diff
@@ -1468,6 +1460,292 @@ Reply
...
+| VFIO_DEVICE_STATE_RESUMING     | 4     | The device is stopped and is loading a new internal state.          |
++--------------------------------+-------+---------------------------------------------------------------------+
+| VFIO_DEVICE_STATE_RUNNING_P2P  | 5     | (not used in vfio-user)                                             |
++--------------------------------+-------+---------------------------------------------------------------------+
+| VFIO_DEVICE_STATE_PRE_COPY     | 6     | The device is running normally but tracking internal state changes. |
++--------------------------------+-------+---------------------------------------------------------------------+
+| VFIO_DEVICE_STATE_PRE_COPY_P2P | 7     | (not used in vfio-user)                                             |
++--------------------------------+-------+---------------------------------------------------------------------+
+
+* *data_fd* is unused in vfio-user, as the ``VFIO_USER_MIG_DATA_READ`` and
```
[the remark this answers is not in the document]
tmakatos: Probably, the only reason why we could keep it is to slightly simplify QEMU implementation, though I'm not sure that's a strong reason.

tmakatos (commented):
I've skimmed through it, need to sit down and compare with VFIO v2.

Could you include in the commit message from which commit/kernel version you're basing this? If master doesn't differ too much from latest stable version, then we could probably use the latter.

include/vfio-user.h
```diff
@@ -228,6 +229,44 @@ struct vfio_user_bitmap_range {
 
 #endif /* VFIO_REGION_TYPE_MIGRATION */
 
+/* Analogous to vfio_device_feature */
+struct vfio_user_device_feature {
+	uint32_t	argsz;
+	uint32_t	flags;
+#define VFIO_DEVICE_FEATURE_MASK	(0xffff)  /* 16-bit feature index */
```
tmakatos, line 232: If an old kernel is used these won't be defined, see how we handle `VFIO_IOMMU_DIRTY_PAGES_FLAG_START`.

include/vfio-user.h
```diff
@@ -228,6 +229,44 @@ struct vfio_user_bitmap_range {
...
+#define VFIO_DEVICE_FEATURE_MIGRATION 1
+
+/* Analogous to vfio_device_feature_mig_state */
+struct vfio_user_device_feature_mig_state {
+    uint32_t    device_state;
+    uint32_t    data_fd;
+} __attribute__((packed));
+#define VFIO_DEVICE_FEATURE_MIG_DEVICE_STATE 2
+
+enum vfio_device_mig_state {
```
tmakatos, line 255: How come this doesn't conflict with `linux/include/uapi/linux/vfio.h`?

lib/libvfio-user.c
```diff
@@ -1052,6 +1052,45 @@ handle_dirty_pages(vfu_ctx_t *vfu_ctx, vfu_msg_t *msg)
     return ret;
 }
 
+static int
+handle_device_feature(vfu_ctx_t *vfu_ctx, vfu_msg_t *msg)
+{
+    struct vfio_user_device_feature *req = msg->in.iov.iov_base;
```
tmakatos, line 1031: nit: here and everywhere else, `assert` arguments

lib/libvfio-user.c
```diff
@@ -1052,6 +1052,45 @@ handle_dirty_pages(vfu_ctx_t *vfu_ctx, vfu_msg_t *msg)
...
+
+    if (vfu_ctx->migration == NULL) {
```
tmakatos, line 1033: Is this necessary? Won't `migration_feature_supported` fail anyway?

lib/migration.c
tmakatos: We can drop `VFIO_MIGRATION_PRE_COPY` for now. @w-henderson have you checked the kernel whether we need to return `RUNNING ` and `ERROR` as they're always supported? If we don't have to return them, we should have a comment.

test/py/libvfio_user.py
tmakatos: nit: no need for parenthesis

tmakatos (commented):
Since this WIP, can you explain in the commit message how libvfio-user handles migration v2? E.g. it understands all migration-related operations but fails them, etc

include/libvfio-user.h
```diff
@@ -615,83 +602,13 @@ typedef struct {
...
-    /*
-     * Function that is called to read migration data. offset and size can be
-     * any subrange on the offset and size previously returned by prepare_data.
-     * The function must return the amount of data read or -1 on error, setting
-     * errno.
-     *
-     * This function can be called even if the migration data can be memory
-     * mapped.
-     */
+
```
tmakatos, line 609: I suppose you're re-introduce documentation in this commit?

lib/libvfio-user.c
```diff
@@ -120,16 +115,6 @@ dev_get_caps(vfu_ctx_t *vfu_ctx, vfu_reg_info_t *vfu_reg, bool is_migr_reg,
...
-        type = (struct vfio_region_info_cap_type *)header;
-        type->header.id = VFIO_REGION_INFO_CAP_TYPE;
-        type->header.version = 1;
-        type->header.next = 0;
-        type->type = VFIO_REGION_TYPE_MIGRATION;
-        type->subtype = VFIO_REGION_SUBTYPE_MIGRATION;
-        vfio_reg->cap_offset = sizeof(struct vfio_region_info);
-    }
-
     if (vfu_reg->mmap_areas != NULL) {
```
tmakatos, line 118: While you're at it, can you add a separate commit where you invert this check so that identation is decreased?

docs/vfio-user.rst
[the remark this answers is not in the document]
tmakatos: Actually, we haven't discussed whether we're going to support memory mapped data transfers for migration (this FD is something we could use), @w-henderson have you given any thought to this?

lib/migration_priv.h
```diff
@@ -33,93 +33,28 @@
...
-        uint64_t size;
-    } iter;
 };
 
-struct migr_state_data {
-    uint32_t state;
-    const char *name;
+/* valid migration state transitions 
+   indexed by vfio_device_mig_state enum */
+static const bool transitions[8][8] = {
```
tmakatos, line 45: What's each row and column?

tmakatos: Also, since all you need is 8x8 you can use bits instead in `char`.

tmakatos (commented):
Could you also explain what this commit does?

lib/migration.c
```diff
@@ -218,6 +218,61 @@ migration_feature_set(vfu_ctx_t *vfu_ctx, uint32_t flags, void *buf)
...
+        return -EINVAL;
+    }
+
+    if (migr->state != VFIO_DEVICE_STATE_PRE_COPY
+        && migr->state != VFIO_DEVICE_STATE_STOP_COPY) {
+        return -EINVAL;
+    }
+
+    msg->out.iov.iov_len = sizeof(req) + req->size;
+    msg->out.iov.iov_base = calloc(1, msg->out.iov.iov_len);
```
tmakatos, line 241: check for memory allocation failure

lib/migration.c
```diff
@@ -218,6 +218,61 @@ migration_feature_set(vfu_ctx_t *vfu_ctx, uint32_t flags, void *buf)
...
+
+    struct vfio_user_mig_data_with_data *res = msg->out.iov.iov_base;
+
+    ssize_t ret = migr->callbacks.read_data(vfu_ctx, &res->data, req->size);
+
+    res->size = ret;
+    res->argsz = ret + sizeof(req);
+
+    if (ret < 0) {
+        return -1;
```
tmakatos, line 251: `return ret`;

lib/migration.c
```diff
@@ -218,6 +218,61 @@ migration_feature_set(vfu_ctx_t *vfu_ctx, uint32_t flags, void *buf)
...
+    struct migration *migr = vfu_ctx->migration;
+    struct vfio_user_mig_data_without_data *req = msg->in.iov.iov_base;
+
+    if (vfu_ctx->migration == NULL) {
+        return -EINVAL;
+    }
+
+    if (migr->state != VFIO_DEVICE_STATE_PRE_COPY
+        && migr->state != VFIO_DEVICE_STATE_STOP_COPY) {
+        return -EINVAL;
```
tmakatos, line 237: a debug log message would be helpful

lib/migration.c
tmakatos: Also, we shouldn't blindly allocate memory as `msg->out.iov.iov_len` might be too large causing us to crash. I think there's some maximum message parameter specified during the negotiation phase?

lib/migration.c
```diff
@@ -218,6 +218,61 @@ migration_feature_set(vfu_ctx_t *vfu_ctx, uint32_t flags, void *buf)
...
+    }
+
+    msg->out.iov.iov_len = sizeof(req) + req->size;
+    msg->out.iov.iov_base = calloc(1, msg->out.iov.iov_len);
+
+    struct vfio_user_mig_data_with_data *res = msg->out.iov.iov_base;
+
+    ssize_t ret = migr->callbacks.read_data(vfu_ctx, &res->data, req->size);
+
+    res->size = ret;
```
tmakatos, line 247: If `ret < -1` then what do `argsz` and `size` mean?

lib/migration.c
```diff
@@ -218,6 +218,61 @@ migration_feature_set(vfu_ctx_t *vfu_ctx, uint32_t flags, void *buf)
...
+{
+    struct migration *migr = vfu_ctx->migration;
+    struct vfio_user_mig_data_with_data *req = msg->in.iov.iov_base;
+
+    if (vfu_ctx->migration == NULL) {
+        return -EINVAL;
+    }
+
+    if (migr->state != VFIO_DEVICE_STATE_RESUMING) {
+        return -EINVAL;
```
tmakatos, line 268: a debug log message would be helpful

lib/migration.c
```diff
@@ -218,6 +218,61 @@ migration_feature_set(vfu_ctx_t *vfu_ctx, uint32_t flags, void *buf)
...
+    }
+
+    ssize_t ret = migr->callbacks.write_data(vfu_ctx, &req->data, req->size);
+
+    if (ret < 0) {
```
tmakatos, line 273: `return migr->callbacks.write_data(vfu_ctx, &req->data, req->size);`

samples/client.c
```diff
@@ -252,7 +240,7 @@ get_region_vfio_caps(struct vfio_info_cap_header *header,
         }
         header = (struct vfio_info_cap_header*)((char*)header + header->next - sizeof(struct vfio_region_info));
     }
-    return migr;
+    return false;
```
tmakatos, line 245: no point in returning `bool`, convert to `void`

samples/client.c
```diff
@@ -526,6 +513,100 @@ access_region(int sock, int region, bool is_write, uint64_t offset,
     return 0;
 }
 
+static int
+set_migration_state(int sock, uint32_t state)
+{
+    static int msg_id = 0xfab1;
+    struct vfio_user_device_feature req = {
+        .argsz = 16,
```
tmakatos, line 522: Why 16? Is this `sizeof struct vfio_user_device_feature + struct vfio_user_device_feature_mig_state`?

samples/client.c
```diff
@@ -526,6 +513,100 @@ access_region(int sock, int region, bool is_write, uint64_t offset,
...
+        .flags = VFIO_DEVICE_FEATURE_SET | VFIO_DEVICE_FEATURE_MIG_DEVICE_STATE
+    };
+    struct vfio_user_device_feature_mig_state change_state = {
+        .device_state = state,
+        .data_fd = 0
```
tmakatos, line 527: `.data_fd = -1`

samples/client.c
```diff
@@ -526,6 +513,100 @@ access_region(int sock, int region, bool is_write, uint64_t offset,
...
+            .iov_base = &req,
+            .iov_len = sizeof(req)
+        },
+        [2] = {
+            .iov_base = &change_state,
+            .iov_len = sizeof(change_state)
+        }
+    };
+    
+    int ret = tran_sock_msg_iovec(sock, msg_id--, VFIO_USER_DEVICE_FEATURE,
```
tmakatos, line 542: just `return tran_sock_msg_iovec(...`

samples/client.c
```diff
@@ -526,6 +513,100 @@ access_region(int sock, int region, bool is_write, uint64_t offset,
...
+        .argsz = 8,
+        .size = len
+    };
+    struct iovec send_iovecs[2] = {
+        [1] = {
+            .iov_base = &req,
+            .iov_len = sizeof(req)
+        }
+    };
+    struct vfio_user_mig_data_with_data *res = calloc(1, sizeof(req) + len);
```
tmakatos, line 573: check for memory allocation failures

samples/client.c
```diff
@@ -526,6 +513,100 @@ access_region(int sock, int region, bool is_write, uint64_t offset,
...
+
+    int ret = tran_sock_msg_iovec(sock, msg_id--, VFIO_USER_MIG_DATA_READ,
+                                  send_iovecs, 2, NULL, 0, NULL,
+                                  res, sizeof(req) + len, NULL, 0);
+
+    if (ret < 0) {
+        free(res);
```
tmakatos, line 582: `res` must be always freed, even in success case

samples/client.c
```diff
@@ -526,6 +513,100 @@ access_region(int sock, int region, bool is_write, uint64_t offset,
...
+
+    return res->size;
+}
+
+static int
+write_migr_data(int sock, void *buf, size_t len)
+{
+    static int msg_id = 0x2023;
+    struct vfio_user_mig_data_with_data req = {
+        .argsz = 8 + len,
```
tmakatos: What's this `8`?

samples/client.c
```diff
@@ -526,6 +513,100 @@ access_region(int sock, int region, bool is_write, uint64_t offset,
...
+            .iov_base = &req,
+            .iov_len = sizeof(req)
+        },
+        [2] = {
+            .iov_base = buf,
+            .iov_len = len
+        }
+    };
+
+    int ret = tran_sock_msg_iovec(sock, msg_id--, VFIO_USER_MIG_DATA_WRITE,
```
tmakatos, line 613: ditto

samples/client.c
```diff
@@ -753,61 +834,34 @@ static size_t
...
-        }
+    for (i = 0; i < nr_iters && is_more; i++) {
 
-        ret = access_region(sock, VFU_PCI_DEV_MIGR_REGION_IDX, false,
-                            offsetof(struct vfio_user_migration_info, data_size),
-                            &data_size, sizeof(data_size));
-        if (ret < 0) {
-            err(EXIT_FAILURE, "failed to read data_size");
-        }
+        migr_iter[i].iov_len = 1024;
```
tmakatos: Why exactly 1K? Generally page size is preferrable. You can allocate outside this loop once and reuse the buffer. Need to check for memory allocation failures.

samples/client.c
```diff
@@ -753,61 +834,34 @@ static size_t
...
-        /*
-         * XXX read pending_bytes again to indicate to the server that the
-         * migration data have been consumed.
-         */
-        ret = access_region(sock, VFU_PCI_DEV_MIGR_REGION_IDX, false,
-                            offsetof(struct vfio_user_migration_info, pending_bytes),
-                            &pending_bytes, sizeof(pending_bytes));
-        if (ret < 0) {
-            err(EXIT_FAILURE, "failed to read pending_bytes");
+        if (ret < (int)migr_iter[i].iov_len) {
```
tmakatos, line 857: Can you explain what this code is about? It's useful to document everything as the sample is used for other people as an example to understand how to use the API. E.g. what does negative length mean?

samples/client.c
```diff
@@ -884,10 +938,8 @@ migrate_from(int sock, size_t *nr_iters, struct iovec **migr_iters,
      * XXX set device state to pre-copy. This is technically optional but any
      * VMM that cares about performance needs this.
      */
-    device_state = VFIO_DEVICE_STATE_V1_SAVING | VFIO_DEVICE_STATE_V1_RUNNING;
-    ret = access_region(sock, VFU_PCI_DEV_MIGR_REGION_IDX, true,
-                        offsetof(struct vfio_user_migration_info, device_state),
-                        &device_state, sizeof(device_state));
+    device_state = VFIO_DEVICE_STATE_PRE_COPY;
```
tmakatos, line 942: I suggest to skip this for now and add it later once you've got stop-and-copy working.

tmakatos (commented):
Does the client/server sample work?

lib/libvfio-user.c
```diff
@@ -1238,6 +1238,13 @@ handle_request(vfu_ctx_t *vfu_ctx, vfu_msg_t *msg)
         ret = handle_device_feature(vfu_ctx, msg);
         break;
 
+    case VFIO_USER_MIG_DATA_READ:
+        ret = handle_mig_data_read(vfu_ctx, msg);
+        break;
+
+    case VFIO_USER_MIG_DATA_WRITE:
+        ret = handle_mig_data_write(vfu_ctx, msg);
+
```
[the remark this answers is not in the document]
tmakatos: This fix should go into the original patch.

lib/libvfio-user.c
```diff
@@ -1044,8 +1043,10 @@ handle_device_feature(vfu_ctx_t *vfu_ctx, vfu_msg_t *msg)
     ssize_t ret;
 
     if (req->flags & VFIO_DEVICE_FEATURE_PROBE) {
-        msg->out.iov.iov_base = msg->in.iov.iov_base;
+        msg->out.iov.iov_base = malloc(msg->in.iov.iov_len);
```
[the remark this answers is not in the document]
tmakatos: This change should go into the original patch.

lib/libvfio-user.c
```diff
@@ -1055,11 +1056,12 @@ handle_device_feature(vfu_ctx_t *vfu_ctx, vfu_msg_t *msg)
         ret = migration_feature_get(vfu_ctx, req->flags,
                                     msg->out.iov.iov_base);
     } else if (req->flags & VFIO_DEVICE_FEATURE_SET) {
-        msg->out.iov.iov_base = msg->in.iov.iov_base;
+        msg->out.iov.iov_base = malloc(msg->in.iov.iov_len);
```
tmakatos, line 1059: check for memory allocation failure

lib/libvfio-user.c
```diff
@@ -2046,7 +2051,7 @@ vfu_setup_device_migration_callbacks(vfu_ctx_t *vfu_ctx, uint64_t flags,
         return ERROR_INT(ret);
     }
 
-    return 0;
+    return ret;
```
tmakatos, line 2054: spurious change

lib/migration.c
```diff
@@ -71,7 +71,11 @@ init_migration(const vfu_migration_callbacks_t *callbacks,
     migr->pgsize = sysconf(_SC_PAGESIZE);
 
     /* FIXME this should be done in vfu_ctx_realize */
-    migr->state = VFIO_DEVICE_STATE_RUNNING;
+    if (flags & LIBVFIO_USER_MIG_FLAG_START_RESUMING) {
```
tmakatos, line 74: Why do we need this? Can't the client transition from RUNNING to RESUMING?

lib/migration_priv.h
```diff
@@ -43,19 +43,16 @@ struct migration {
...
 static const bool transitions[8][8] = {
-    {0, 0, 0, 0, 0, 0, 0, 0},
-    {0, 0, 0, 1, 1, 0, 0, 0},
-    {0, 0, 0, 0, 0, 0, 1, 0},
-    {0, 1, 0, 0, 0, 0, 0, 0},
-    {0, 1, 0, 0, 0, 0, 0, 0},
-    {0, 0, 0, 0, 0, 0, 0, 0},
-    {0, 0, 1, 0, 0, 0, 0, 0},
-    {0, 0, 0, 0, 0, 0, 0, 0}
+    {0, 0, 0, 0, 0, 0, 0, 0}, // ERROR
```
tmakatos, line 46: We should also have description for the columns.

samples/client.c
```diff
@@ -62,6 +62,8 @@ static char const *irq_to_str[] = {
     [VFU_DEV_REQ_IRQ] = "REQ"
 };
 
+static pthread_mutex_t mutex = PTHREAD_MUTEX_INITIALIZER;
```
tmakatos, line 65: Why do we need the mutex?

samples/client.c
```diff
@@ -535,15 +536,23 @@ set_migration_state(int sock, uint32_t state)
             .iov_len = sizeof(change_state)
         }
     };
+    void* response = malloc(sizeof(req) + sizeof(change_state));
     
+    pthread_mutex_lock(&mutex);
     int ret = tran_sock_msg_iovec(sock, msg_id--, VFIO_USER_DEVICE_FEATURE,
                                   send_iovecs, 3, NULL, 0, NULL,
-                                  &req, sizeof(req), NULL, 0);
+                                  response, sizeof(req) + sizeof(change_state),
```
tmakatos, line 544: This change should go into the earlier commit that touched this code for migration v2.

samples/client.c
```diff
@@ -535,15 +536,23 @@ set_migration_state(int sock, uint32_t state)
...
+                                  NULL, 0);
+    pthread_mutex_unlock(&mutex);
 
     if (ret < 0) {
         return -1;
     }
 
+    assert(memcmp(&req, response, sizeof(req)) == 0);
```
tmakatos, line 552: Are you checking that the server succesfully changed migration state? If so then it should not be done with `assert` and we should be printing relevant values in case of error as this helps debugging.

samples/client.c
```diff
@@ -574,6 +585,8 @@ read_migr_data(int sock, void *buf, size_t len)
 
     memcpy(buf, res->data, res->size);
 
+    free(res);
```
tmakatos, line 588: ditto about original commit

samples/client.c
```diff
@@ -816,32 +831,17 @@ usage(char *argv0)
...
- * we simply accumulate the migration data of each iteration and apply it to
- * the destination server at the end.
- *
- * Performs as many migration loops as @nr_iters or until the device has no
- * more migration data (pending_bytes is zero), which ever comes first. The
- * result of each migration iteration is stored in @migr_iter.  @migr_iter must
- * be at least @nr_iters.
- *
- * @returns the number of iterations performed
- */
```
tmakatos, line 751: Is this comment not relevant any more?

samples/client.c
```diff
@@ -816,32 +831,17 @@ usage(char *argv0)
...
 static size_t
-do_migrate(int sock, size_t nr_iters, struct iovec *migr_iter)
+do_migrate(int sock, size_t max_iters, size_t max_iter_size,
+           struct iovec *migr_iter)
 {
     int ret;
+    size_t i;
```
tmakatos, line 839: spurious change

samples/client.c
```diff
@@ -1007,9 +1009,10 @@ migrate_to(char *old_sock_path, int *server_max_fds,
     if (ret == -1) {
         err(EXIT_FAILURE, "failed to fork");
     }
-    if (ret > 0) { /* child (destination server) */
+    if (ret == 0) { /* child (destination server) */
```
tmakatos, line 1012: This change should go into a separate commit (most likely a separate PR).

samples/server.c
```diff
@@ -62,7 +62,8 @@ struct server_data {
     size_t bar1_size;
     struct dma_regions regions[NR_DMA_REGIONS];
     struct {
-        uint64_t pending_bytes;
+        uint64_t pending_read;
```
tmakatos, line 65: Can you explain why we now need two variables?

samples/server.c
```diff
@@ -291,72 +293,121 @@ migration_read_data(vfu_ctx_t *vfu_ctx, void *buf, uint64_t size)
...
      * both BAR1 and BAR0. Since we always copy BAR1 in the stop-and-copy state,
      * copying BAR1 in the pre-copy state is pointless. Fixing this requires
      * more complex state tracking which exceeds the scope of this sample.
      */
 
-    if (size != server_data->migration.pending_bytes) {
-        errno = EINVAL;
-        return -1;
+    if (server_data->migration.pending_read == 0 || size == 0) {
+        return 0;
```
tmakatos, line 303: nit: can we have a debug log message clarifying what's going on?

tmakatos (commented):
I haven't reviewed from `migration_read_data` onwards, can you explain in the commit message how the code changes and why? Also, some comments in the code would also help, I know it wasn't there in v1 but it should have been, let's improve with v2.

include/vfio-user.h
```diff
@@ -66,10 +66,11 @@ enum vfio_user_command {
     VFIO_USER_DMA_READ                  = 11,
     VFIO_USER_DMA_WRITE                 = 12,
     VFIO_USER_DEVICE_RESET              = 13,
-    VFIO_USER_DIRTY_PAGES               = 14,
-    VFIO_USER_DEVICE_FEATURE            = 15,
-    VFIO_USER_MIG_DATA_READ             = 16,
-    VFIO_USER_MIG_DATA_WRITE            = 17,
+    VFIO_USER_DIRTY_PAGES               = 14, // TODO remove
+    VFIO_USER_REGION_WRITE_MULTI        = 15, // TODO implement
```
tmakatos, line 70: `VFIO_USER_REGION_WRITE_MULTI` should be added to the commit where you pulled the new spec.

lib/libvfio-user.c
w-henderson: Yes, is this the sort of thing we want an assertion for or do we want it to fail gracefully?

lib/migration.c
w-henderson: If we're trying to call `SET` on `VFIO_DEVICE_FEATURE_MIGRATION`, this isn't an allowed operation, but we allow `GET` and `PROBE`. For `VFIO_DEVICE_FEATURE_MIG_DEVICE_STATE` we allow all three operations, and since these are the only supported features (apart from the DMA logging one which I haven't started implementing yet), we return `false` for anything else.

lib/libvfio-user.c
w-henderson: It's not invalid - you set `GET` and `SET` to specify which operations you are probing support for.

From the spec:
> `VFIO_DEVICE_FEATURE_PROBE` instructs the server to probe for feature support. If `VFIO_DEVICE_FEATURE_GET/SET` are also set, the probe will only return success if all of the indicated methods are supported.

lib/libvfio-user.c
w-henderson: We set the reply to equal the request, as this is what VFIO and hence vfio-user dictates for the success response - this has since been replaced by the following, as the aliasing caused a double free: 😳

```c
msg->out.iov.iov_base = malloc(msg->in.iov.iov_len);
msg->out.iov.iov_len = msg->in.iov.iov_len;
memcpy(msg->out.iov.iov_base, msg->in.iov.iov_base,
       msg->out.iov.iov_len);
```

lib/migration.c
w-henderson: Does it not make sense to check support for the feature support all in one place? To me at least it wouldn't make sense to return `true` from `migration_feature_supported` in the case, for example, where `flags = VFIO_DEVICE_FEATURE_MIGRATION | VFIO_DEVICE_FEATURE_SET` - in this case, the feature is not supported for setting?

lib/libvfio-user.c
w-henderson: Yes:

> The reply payload must be the same as the request payload for setting or probing a feature.

lib/migration.c
w-henderson: We don't need to return `RUNNING` or `ERROR` because they're always supported... interestingly we do need to return `STOP_COPY` despite that also always being supported! From the kernel:

> If this `GET` succeeds, the `RUNNING` and `ERROR` states are always supported. Support for additional states is indicated via the flags field; at least `VFIO_MIGRATION_STOP_COPY` must be set.

I'll add a comment.

lib/libvfio-user.c
jlevon: yeah, so I read after.

lib/libvfio-user.c
tmakatos: I haven't checked that code, if it's something the client can control then it should be handled gracefully.

lib/libvfio-user.c
jlevon: Such an API would not be useful. I think you're supposed to respond with the set of flags known to the server. (And, if get, or set, is set, whether the server supports that verb for the given flag(s)).

lib/migration.c
jlevon: This really is an implementation of "probe" if I understand things right. So you could have a "probe()" function in the general code, that delegates to migration.c for the migration flags (which of course are currently all of them!). Up to you.

lib/migration.c
jlevon: I'm not sure what my question was here.

lib/migration.c
jlevon: Sure, but in general, PRs should be self-contained and make sense on their own.

include/vfio-user.h
w-henderson: It does, I'll rename it to `vfio_user_device_mig_state`.

lib/libvfio-user.c
w-henderson: At the moment migration is all-or-nothing and `migration_feature_supported` assumes we support migration... I don't think it makes sense to support another option until we also implement DMA logging (which will definitely be optional). What do you think?

lib/migration_priv.h
w-henderson: `transitions[FROM][TO]` gives whether or not the transition from state `FROM` to state `TO` is allowed. I'll change it to bits.

tmakatos: What I meant was that each row and column should be documented, e.g. how we document valid transitions in the spec here: https://github.com/nutanix/libvfio-user/blob/master/docs/vfio-user.rst#id107

include/libvfio-user.h
w-henderson: Yes, it needed to be re-written though so I removed the existing documentation for now - I will definitely add it later.

lib/libvfio-user.c
w-henderson: I feel like it makes more sense to call it `migration_feature_flags` and return the supported flags for the given feature, then have the check for migration unsupported elsewhere?

include/libvfio-user.h
tmakatos: I assume you're create an individual PR for this commit and fix ups can appear as separate commits, such as this one?

lib/libvfio-user.c
tmakatos: makes sense

lib/migration.c
w-henderson: Undefined, what do you think we should do about errors here?

include/vfio-user.h
```diff
@@ -267,13 +267,13 @@ enum vfio_device_mig_state {
 // used for read request and write response
 struct vfio_user_mig_data_without_data {
     uint32_t    argsz;
-    uint32_t    size;
+    uint64_t    size;
```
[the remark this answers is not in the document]
w-henderson: If I'm honest I can't remember why I changed this, should I change it back?

lib/migration.c
w-henderson: The client is not allowed to transition from RUNNING to RESUMING in the kernel's VFIO protocol so we don't allow it here.

lib/migration.c
tmakatos: IIRC it's documented in the spec how responses work, so `argsz` should contain the size of the response header. There should be other parts of the code that handle similar scenario so you can look there.

include/vfio-user.h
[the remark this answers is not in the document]
tmakatos: yes, we should change it back to 32-bit

lib/migration.c
tmakatos: OK, then we should document the purpose of this flag where it gets defined.

docs/vfio-user.rst
```diff
@@ -1468,6 +1460,292 @@ Reply
...
++-------+--------+--------------------------------+
+| data  | 8      | variable                       |
++-------+--------+--------------------------------+
+
+* *argsz* is the size of the above structure.
+
+* *flags* defines the action to be performed by the server and upon which
+  feature:
+
+  * ``VFIO_DEVICE_FEATURE_MASK`` is the 16-bit feature index of the relevant
```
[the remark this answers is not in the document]
w-henderson: How do I phrase what the first 16 bits are, just "feature"?

docs/vfio-user.rst
```diff
@@ -1468,6 +1460,292 @@ Reply
...
+    feature.
+  
+  * ``VFIO_DEVICE_FEATURE_GET`` instructs the server to get the feature data.
+
+  * ``VFIO_DEVICE_FEATURE_SET`` instructs the server to set the feature data to
+    that given in the ``data`` field of the payload.
+
+  * ``VFIO_DEVICE_FEATURE_PROBE`` instructs the server to probe for feature
```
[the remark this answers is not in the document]
w-henderson: I think the idea is if you don't set GET/SET, it is successful if the feature is supported at all, but if you set GET/SET it is only successful if whatever commands you probed for are also supported. How would you suggest I word this more clearly?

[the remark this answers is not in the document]
jlevon: I would add a whole section on probe, saying something like:

Probing
----------

If just `VFIO_DEVICE_FEATURE_PROBE` is set, the server should return the corresponding flags that are supported.

If `VFIO_DEVICE_FEATURE_PROBE` is set along with `VFIO_DEVICE_FEATURE_GET`, `VFIO_DEVICE_FEATURE_SET`, or both, the server should return whether the corresponding operation(s) are supported for the given flags.

docs/vfio-user.rst
```diff
@@ -1468,6 +1460,292 @@ Reply
...
+* *data* is specific to the particular feature. It is not used for probing.
+
+This part of the request is analogous to VFIO's ``struct vfio_device_feature``.
+
+Reply
+^^^^^
+
+The reply payload must be the same as the request payload for setting or
+probing a feature. For getting a feature's data, the data is added in the data
+section and its length is added to ``argsz``.
```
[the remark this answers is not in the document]
w-henderson: You're right, in fact I've implemented it completely wrong and am returning the data by itself instead of within the payload... I'll sort that out.

docs/vfio-user.rst
```diff
@@ -1468,6 +1460,292 @@ Reply
...
+``VFIO_DEVICE_FEATURE_MIGRATION``
+"""""""""""""""""""""""""""""""""
+
+This feature indicates that the device can support the migration API through
+``VFIO_DEVICE_FEATURE_MIG_DEVICE_STATE``. If ``GET`` succeeds, the ``RUNNING``
+and ``ERROR`` states are always supported. Support for additional states is
+indicated via the flags field; at least ``VFIO_MIGRATION_STOP_COPY`` must be
+set.
+
+There is no data field of the request message.
```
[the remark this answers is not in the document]
w-henderson: This bit of the spec is pretty much just lifted from VFIO, I don't believe you can call SET on this feature...?

https://elixir.bootlin.com/linux/v6.3.8/source/include/uapi/linux/vfio.h#L814

docs/vfio-user.rst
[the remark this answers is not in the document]
w-henderson: I haven't thought about this - I quite like the idea of following VFIO as closely as we can so my preference is probably to leave it in.

lib/libvfio-user.c
w-henderson: Are we sure we want to diverge from VFIO here?

From documentation of VFIO_DEVICE_FEATURE:
> Get, set, or probe feature data of the device.  The feature is selected using the FEATURE_MASK portion of the flags field.  Support for a feature can be probed by setting both the FEATURE_MASK and PROBE bits.  A probe may optionally include the GET and/or SET bits to determine read vs write access of the feature respectively.  Probing a feature will return success if the feature is supported and all of the optionally indicated GET/SET methods are supported.  The format of the data portion of the structure is specific to the given feature.  The data portion is not required for probing.  GET and SET are mutually exclusive, except for use with PROBE.

docs/vfio-user.rst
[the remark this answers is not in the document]
tmakatos: > support memory mapped data transfers for migration (this FD is something we could use)

What I said doens't make sense, the FD needs to be passed via ancillary data. I don't have a strong opinion about keeping it, perhaps qemu-devel may help us decide.

docs/vfio-user.rst
tmakatos: > I believe in VFIO we never "seek", only read/write until no more reading/writing can be done.

That's correct.

I'm not sure I understand the rest of the discussion though, especially how dirty page tracking is related to device state tracking.

I have the following question (perhaps we're talking about the same thing?), regarding PRE_COPY, from VFIO:
```
 * RUNNING -> PRE_COPY
 * RUNNING_P2P -> PRE_COPY_P2P
 * STOP -> STOP_COPY
 *   PRE_COPY, PRE_COPY_P2P and STOP_COPY form the "saving group" of states
 *   which share a data transfer session. Moving between these states alters
 *   what is streamed in session, but does not terminate or otherwise affect
 *   the associated fd.
 *
 *   These arcs begin the process of saving the device state and will return a
 *   new data_fd. The migration driver may perform actions such as enabling
 *   dirty logging of device state when entering PRE_COPY or PER_COPY_P2P.
 *
 *   Each arc does not change the device operation, the device remains
 *   RUNNING, P2P quiesced or in STOP. The STOP_COPY state is described below
 *   in PRE_COPY_P2P -> STOP_COPY.
 ```
IIUC PRE_COPY returns an FD as this starts a new copy session, so in vfio-user this means that the client can start sending ``VFIO_USER_MIG_DATA_READ`` messages until the server returns 0 length read (equivalnt to the `data_fd` returning no more data in VFIO). While in PRE_COPY state, the device migth generate more device data, how does the VFIO client know this? Does reading from the `data_fd` return data (it would previously return EOF).

samples/client.c
w-henderson: Yes.

samples/client.c
w-henderson: I've added additional checks afterwards so this is no longer relevant.

samples/client.c
w-henderson: `sizeof(struct vfio_user_mig_data)`.

samples/client.c
w-henderson: Later changes have added a mutex around this so we have to unlock the mutex before we can return.

samples/client.c
w-henderson: This has been since resolved.

samples/client.c
w-henderson: We know we've finished transferring data when less is returned than is requested. Negative `ret` is error (which is handled earlier), but I agree it makes more sense to cast `ret` to `size_t` than `iov_len` to `int` (since negative length has no meaning).

samples/client.c
w-henderson: I think I've got this working.

samples/client.c
w-henderson: I'm not sure why I removed it, it still applies. I'll put it back.

samples/client.c
w-henderson: The fake guest and the client use the same socket so we have to ensure the responses do not get mixed up.

samples/client.c
tmakatos: OK then could you replace it?

samples/client.c
tmakatos: What changed and we need a mutex?

samples/client.c
tmakatos: Adding a comment would be really helpful.

samples/server.c
w-henderson: I thought it made it clearer but now I realise it's unnecessary so I'll remove it.

include/vfio-user.h
w-henderson: I'm not sure about this - it isn't a spec change, it's an implementation change...?

tmakatos: good point

samples/client.c
tmakatos: This problem should exist in the original version of the code as well, right? Do your changes make an existing bug easier to trigger? If so, I think it's simpler to have the fake guest not require a resposne since it only writes to BAR1.

samples/client.c
w-henderson: I have done 👍

docs/vfio-user.rst
w-henderson: VFIO defines `PRE_COPY` to mean

> The device is running normally but tracking internal state changes

which maybe implies that we don't actually *transfer* any data in that state, rather just do whatever we can in the background to prepare for transfer and track the changes to keep this up to date? I'm not sure if this makes sense though.

docs/vfio-user.rst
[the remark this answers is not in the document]
w-henderson: I went for something similar to VFIO

> The feature is selected using the 16-bit ``VFIO_DEVICE_FEATURE_MASK`` portion of the flags field.

docs/vfio-user.rst
w-henderson: I think it means that the device does not do anything differently from the client's perspective, only that it may be doing something slightly differently in the background (i.e. tracking state changes in PRE_COPY or actually copying the data in STOP_COPY)

tmakatos: I'm confused, the documentation says:
> Each arc does not change the device operation, the device remains RUNNING, P2P quiesced or in STOP.

while you say:
> It definitely doesn't stay in the exact same state though.

w-henderson: I'm a bit confused too, I had been thinking of them as distinct states but I guess it's supposed to be more like different behaviour in the same state? That sounds a bit dodgy?

swapnili:
@w-henderson Is this PR up for review or WIP I see that the build is failing? Are you going to split this into small logical PRs?

w-henderson:
@swapnili This PR has since been split into the smaller #745 #746 #747 #754 which are at various different stages of development, I'll close this one since it's no longer relevant.

PR #744, opened 2023-07-12
update to latest version of the spec

w-henderson:
This pulls in the version of the spec that was submitted to qemu-devel: https://lists.nongnu.org/archive/html/qemu-devel/2023-06/msg06567.html

This version of the spec removes live migration support as it will be added later on following v2 of the VFIO migration protocol. However, we will not remove the code for live migration v1 until we implement v2.

We do not implement all of the other new changes in this spec update but are otherwise compatible.

### DIFF
--- a/docs/vfio-user.rst
+++ b/docs/vfio-user.rst
@@ -1,5 +1,4 @@
 .. include:: <isonum.txt>
-
 ********************************
 vfio-user Protocol Specification
 ********************************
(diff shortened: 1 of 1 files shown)


tmakatos (commented):
Looks good.

I don't think we can merge it to master yet as it has to bring all other commits (which are currently in the form of PRs). Perhaps it's best to create a branch off master and merge each approved commit there, while you work on your PRs? E.g. it can be called `master-migv2` (or something). Once all of your PRs are merged we can merge that branch into master without a review.

PR #747, opened 2023-07-12
update samples to use migration v2

w-henderson:


samples/client.c
```diff
@@ -868,13 +921,15 @@ migrate_from(int sock, size_t *nr_iters, struct iovec **migr_iters,
         .crcp = crcp
     };
 
+    size_t max_iter_size = 4096;
```
tmakatos, line 924: get it from `sysconf(_SC_PAGE_SIZE)`

samples/client.c
```diff
@@ -868,13 +921,15 @@ migrate_from(int sock, size_t *nr_iters, struct iovec **migr_iters,
...
+
     ret = pthread_create(&thread, NULL, fake_guest, &fake_guest_data);
     if (ret != 0) {
         errno = ret;
         err(EXIT_FAILURE, "failed to create pthread");
     }
 
-    *nr_iters = 2;
+    *nr_iters = 8;
```
tmakatos, line 932: Why?

samples/client.c
```diff
@@ -960,9 +1009,10 @@ migrate_to(char *old_sock_path, int *server_max_fds,
     if (ret == -1) {
         err(EXIT_FAILURE, "failed to fork");
     }
-    if (ret > 0) { /* child (destination server) */
+    if (ret == 0) { /* child (destination server) */
```
tmakatos, line 1012: That should be a separate fix.

samples/client.c
```diff
@@ -1056,6 +1072,15 @@ migrate_to(char *old_sock_path, int *server_max_fds,
     if (dst_crc != src_crc) {
         fprintf(stderr, "client: CRC mismatch: %u != %u\n", src_crc, dst_crc);
         abort();
+    } else {
+        fprintf(stdout, "client: CRC match, we did it! :)\n");
```
tmakatos, line 1076: `printf` is better.

samples/gpio-pci-idio-16.c
```diff
@@ -77,49 +77,23 @@ migration_device_state_transition(vfu_ctx_t *vfu_ctx, vfu_migr_state_t state)
...
-static int
-migration_data_written(UNUSED vfu_ctx_t *vfu_ctx, uint64_t count)
-{
-    assert(count == sizeof(pin));
-    return 0;
+    if (dirty) {
+        memcpy(buf, &pin, sizeof(pin));
+        dirty = false;
+        return sizeof(pin);
+    } else {
```
tmakatos, line 89: nit: no need for `else` after return.

tmakatos (commented):
I haven't gone through all of the changes.

Can you explain in the commit message the changes you're making? E.g. the replacement of `pending_bytes` with `pending_read` and `pending_write`, the change in number of iterations, etc.

samples/client.c
```diff
@@ -62,6 +62,8 @@ static char const *irq_to_str[] = {
     [VFU_DEV_REQ_IRQ] = "REQ"
 };
 
+static pthread_mutex_t mutex = PTHREAD_MUTEX_INITIALIZER;
```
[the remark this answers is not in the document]
w-henderson: We don't, I think it's a remnant of some debugging from last week that I accidentally forgot to remove - I'll get rid of it!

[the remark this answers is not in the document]
w-henderson: The fake guest is calling `access_region` while we do the migration so we have to ensure that the responses don't get mixed up.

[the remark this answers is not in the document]
tmakatos: Why is this a problem with migration v2 and not v1?

samples/client.c
```diff
@@ -526,6 +514,114 @@ access_region(int sock, int region, bool is_write, uint64_t offset,
...
+
+    return ret;
+}
+
+static int
+read_migr_data(int sock, void *buf, size_t len)
+{
+    static int msg_id = 0x6904;
+    struct vfio_user_mig_data_without_data req = {
+        .argsz = 12,
```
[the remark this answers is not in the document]
w-henderson: I'll replace it with `sizeof(struct vfio_user_mig_data)`.

samples/client.c
```diff
@@ -526,6 +514,114 @@ access_region(int sock, int region, bool is_write, uint64_t offset,
...
+
+    return res->size;
+}
+
+static int
+write_migr_data(int sock, void *buf, size_t len)
+{
+    static int msg_id = 0x2023;
+    struct vfio_user_mig_data_with_data req = {
+        .argsz = 12 + len,
```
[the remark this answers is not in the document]
w-henderson: Again, I'll replace it with `sizeof(struct vfio_user_mig_data)`.

samples/client.c
w-henderson: This was actually incorrect behaviour - I just set to the page size because I didn't realise that what we actually want is the maximum supported transfer size, as determined by the negotiation when we connect. In the client example, `CLIENT_MAX_DATA_XFER_SIZE` is 1024, which allows us to test the chunking.

samples/client.c
w-henderson: I'm quite confused about how the `nr_iters` stuff works - is it literally just a mechanism to ensure that the server cannot just keep sending data indefinitely to stop the client? In which case, would it make more sense to set `nr_iters` to something like `ceil(expected_data / max_iter_size)`?

samples/client.c
w-henderson: I'll revert it in this one then.

samples/client.c
tmakatos: I _think_ `nr_iters` was used to avoid reading the entire migration data in one go, in order to make the sample a bit more realistic. This could have been better documented.
If you do have to change it, then please improve the (currently non-existent) documentation of `nr_iters` and also explain why it has to be changed.

w-henderson: Okay, this makes sense.

samples/client.c
tmakatos: That was for DMA, not sure how it affects migration, can you investigate?

samples/client.c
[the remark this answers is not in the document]
w-henderson: I think it might've been an issue with v1 too - every so often running the sample from the master branch you randomly get a CRC mismatch (but it doesn't seem to completely fall over like with v2 without the mutex).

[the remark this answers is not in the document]
w-henderson: This doesn't seem to fix the problem - I have it now setting the `no_reply` flag on writes to BAR1 in the master version but the CRC mismatch still occurs...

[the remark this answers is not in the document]
w-henderson: Wait I've figured it out, it's not a concurrency bug, it's that in `fake_guest` we calculate the CRC with the previous CRC as the initialisation bytes (incorrectly), so if the fake guest does more than one iteration the CRC will not match.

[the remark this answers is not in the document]
tmakatos: That's a different problem then, let's skip fixing master for now.

Back to v2, if you set `no_reply` you shouldn't be getting mixed responses, so there's shouldn't be a need for additional locking to prevent this. CRC mismatch errors should be irrelevant here.

Regarding CRC mismatch errors, here's what I had in mind when implementing the client/server sample:

- the fake guest constantly writes to BAR1 and store the CRC of what it wrote
- the main thread sets migration state to PRE_COPY and does a few, pointless save iterations. It doesn't care about CRC. IIUC the fake guest does at least one iteration.
- the main thread stops the fake guest, sets migration state to SAVING and does a few saving iterations to read device state. Ideally it should do at least two such "iterations", where it should read BAR1 in at least two steps to be more realistic (I realize that iterations is the wrong term, it should be steps).
- the main thread forks and creates the destination servers, sets device state to RESUMING, applies device state in at lest two steps (again, misusing term iterations), sets device state to running, validates CRC but does not restart the fake guest.

So I _believe_ there shouldn't be any CRC mismatch errors, obviously if you still get them then there's something wrong. First of all, could you add this description somewhere? Second, could you fix the misuse of iterations, e.g. replace with steps? Third, can you add some assertions where we execute at least two save/resume steps?

samples/client.c
```diff
@@ -267,40 +257,6 @@ do_get_device_region_info(int sock, struct vfio_region_info *region_info,
     }
 }
 
-static void
-mmap_sparse_areas(int *fds, struct vfio_region_info *region_info,
```
swapnili, line 271: Why are you removing this? I think still mapping sparse areas is valid test

samples/client.c
```diff
@@ -526,6 +474,126 @@ access_region(int sock, int region, bool is_write, uint64_t offset,
...
+    
+    pthread_mutex_lock(&mutex);
+    int ret = tran_sock_msg_iovec(sock, msg_id--, VFIO_USER_DEVICE_FEATURE,
+                                  send_iovecs, 3, NULL, 0, NULL,
+                                  response, sizeof(req) + sizeof(change_state),
+                                  NULL, 0);
+    pthread_mutex_unlock(&mutex);
+
+    if (ret < 0) {
+        return ret;
```
swapnili, line 514: free(response) before return

samples/client.c
```diff
@@ -690,40 +758,40 @@ get_dirty_bitmap(int sock, struct vfio_user_dma_map *dma_region)
...
+    struct vfio_user_device_feature *res;
+    struct vfio_user_device_feature_dma_logging_report *report;
+    
     char *bitmap;
-    size_t size;
-    void *data;
     int ret;
 
-    size = sizeof(*dirty_pages) + sizeof(*range) + bitmap_size;
+    ssize_t size = sizeof(*res) + sizeof(*report) + bitmap_size;
```
swapnili, line 768: Why change to signed?

samples/client.c
```diff
@@ -690,40 +758,40 @@ get_dirty_bitmap(int sock, struct vfio_user_dma_map *dma_region)
...
                         NULL, data, size);
     if (ret != 0) {
         err(EXIT_FAILURE, "failed to get dirty page bitmap");
     }
 
     printf("client: %s: %#llx-%#llx\t%#x\n", __func__,
-           (ull_t)range->iova,
-           (ull_t)(range->iova + range->size - 1), bitmap[0]);
+           (ull_t)report->iova,
+           (ull_t)(report->iova + report->length - 1), bitmap[0]);
```
swapnili, line 795: free(res->argsz)

samples/client.c
```diff
@@ -1108,6 +1118,14 @@ int main(int argc, char *argv[])
     uint32_t crc;
     size_t bar1_size = 0x3000; /* FIXME get this value from region info */
 
+    struct vfio_user_device_feature *dirty_pages_feature;
+    struct vfio_user_device_feature_dma_logging_control *dirty_pages_control;
+    size_t dirty_pages_size = sizeof(*dirty_pages_feature) +
+                               sizeof(*dirty_pages_control);
+    void* dirty_pages = malloc(dirty_pages_size);
```
swapnili, line 1125: free(dirty_pages) at the end

swapnili (commented):
Some initial comments. I haven't finished reviewing whole patch.

samples/client.c
w-henderson: `mmap_sparse_areas` was only ever called on the migration region so it's not applicable anymore - I'm not sure I understand what it's doing either so I don't know how I could adapt it to demonstrate another region?

samples/client.c
w-henderson: I have no idea, must've been a mistake! I'll change it back.

samples/client.c
w-henderson: `res->argsz` is just a `uint32_t` struct field? Isn't it already freed when we free `data` which is the allocation behind the whole of `res`, `report` and the bitmap?

samples/client.c
swapnili: In server.c VFU_PCI_DEV_BAR1_REGION_IDX region also has sparse mmaps

w-henderson: Why does it expect 2 FDs for VFU_PCI_DEV_BAR1_REGION_IDX? I get that there are two sparse areas but don't we have one file backing them?

samples/client.c
```diff
@@ -526,6 +508,129 @@ access_region(int sock, int region, bool is_write, uint64_t offset,
...
+        [1] = {
+            .iov_base = &req,
+            .iov_len = sizeof(req)
+        },
+        [2] = {
+            .iov_base = &change_state,
+            .iov_len = sizeof(change_state)
+        }
+    };
+    void* response = malloc(sizeof(req) + sizeof(change_state));
```
jlevon, line 534: nit, "void *response"

samples/client.c
```diff
@@ -690,40 +795,40 @@ get_dirty_bitmap(int sock, struct vfio_user_dma_map *dma_region)
...
     char *bitmap;
-    size_t size;
-    void *data;
     int ret;
 
-    size = sizeof(*dirty_pages) + sizeof(*range) + bitmap_size;
+    size_t size = sizeof(*res) + sizeof(*report) + bitmap_size;
 
-    data = calloc(1, size);
+    void* data = calloc(1, size);
```
jlevon, line 807: nit, void *data

samples/client.c
```diff
@@ -750,64 +855,33 @@ usage(char *argv0)
...
-        /*
-         * XXX read pending_bytes again to indicate to the server that the
-         * migration data have been consumed.
-         */
-        ret = access_region(sock, VFU_PCI_DEV_MIGR_REGION_IDX, false,
-                            offsetof(struct vfio_user_migration_info, pending_bytes),
-                            &pending_bytes, sizeof(pending_bytes));
-        if (ret < 0) {
-            err(EXIT_FAILURE, "failed to read pending_bytes");
+        // We know we've finished transferring data when less is returned
```
jlevon, line 880: please fix as per other discussion

samples/client.c
```diff
@@ -1108,6 +1160,14 @@ int main(int argc, char *argv[])
     uint32_t crc;
     size_t bar1_size = 0x3000; /* FIXME get this value from region info */
 
+    struct vfio_user_device_feature *dirty_pages_feature;
+    struct vfio_user_device_feature_dma_logging_control *dirty_pages_control;
+    size_t dirty_pages_size = sizeof(*dirty_pages_feature) +
+                               sizeof(*dirty_pages_control);
+    void* dirty_pages = malloc(dirty_pages_size);
```
jlevon, line 1167: void *dirty_pages

samples/server.c
```diff
@@ -263,16 +263,17 @@ migration_device_state_transition(vfu_ctx_t *vfu_ctx, vfu_migr_state_t state)
...
         case VFU_MIGR_STATE_PRE_COPY:
-            /* TODO must be less than size of data region in migration region */
             server_data->migration.pending_bytes = server_data->bar1_size;
             break;
         case VFU_MIGR_STATE_STOP:
             /* FIXME should gracefully fail */
             assert(server_data->migration.pending_bytes == 0);
             break;
         case VFU_MIGR_STATE_RESUME:
+            server_data->migration.pending_bytes = server_data->bar1_size + sizeof(time_t);
```
jlevon, line 273: should discuss why/if we have pending_bytes at all still

samples/server.c
```diff
@@ -285,125 +286,120 @@ migration_device_state_transition(vfu_ctx_t *vfu_ctx, vfu_migr_state_t state)
...
+        // TODO check the following is always allowed
 
-static size_t
-nr_pages(size_t size)
-{
-    return (size / sysconf(_SC_PAGE_SIZE) +
-            (size % sysconf(_SC_PAGE_SIZE) > 1));
-}
+        memcpy(server_data->bar1 + write_start, buf, bytes_written);
+    } else if (
```
jlevon, line 377: fix formatting please?
If you find you need comments in the middle of an if, that's a good indication you need to think about a refactor!

w-henderson: Yep, I'm not sure why I had the second case not just be an `else` at the end!

samples/server.c
w-henderson: Actually maybe `bytes_read` is not a good name given that we use it to track how much we've written too (not at the same time of course), maybe something like `bytes_transferred`? What do you think?

jlevon (approved):

samples/client.c
```diff
@@ -526,6 +508,129 @@ access_region(int sock, int region, bool is_write, uint64_t offset,
...
+               sizeof(change_state)) != 0) {
+        err(EXIT_FAILURE, "invalid response to set_migration_state (payload)");
+    }
+
+    free(response);
+
+    return ret;
+}
+
+static int
```
swapnili, line 566: Use ssize_t return type

samples/client.c
```diff
@@ -750,64 +855,33 @@ usage(char *argv0)
...
-        migr_iter[i].iov_base = malloc(data_size);
         if (migr_iter[i].iov_base == NULL) {
             err(EXIT_FAILURE, "failed to allocate migration buffer");
         }
 
         /* XXX read migration data */
-        ret = access_region(sock, VFU_PCI_DEV_MIGR_REGION_IDX, false,
-                            data_offset,
-                            (char *)migr_iter[i].iov_base, data_size);
+        ret = read_migr_data(sock, migr_iter[i].iov_base, migr_iter[i].iov_len);
```
swapnili, line 875: return value can be size so use ssize_t type

samples/client.c
```diff
@@ -884,16 +962,17 @@ migrate_from(int sock, size_t *nr_iters, struct iovec **migr_iters,
...
-                        offsetof(struct vfio_user_migration_info, device_state),
-                        &device_state, sizeof(device_state));
+    device_state = VFIO_USER_DEVICE_STATE_PRE_COPY;
+    ret = set_migration_state(sock, device_state);
     if (ret < 0) {
         err(EXIT_FAILURE, "failed to write to device state");
     }
 
-    _nr_iters = do_migrate(sock, 1, *migr_iters);
-    assert(_nr_iters == 1);
```
swapnili, line 896: Can you have similar assert() to check ret below?

w-henderson: Sorry I'm not sure what you mean - we're already checking the return value was success?

samples/client.c
tmakatos: Expecting 2 FDs is correct, the server might use multiple files for each sparse region of a device region. The fact that it wasn't clear means we should have a comment explaining that, could you do this in a separate commit?

samples/client.c
swapnili: Right I meant having assert for iters value

w-henderson: Okay cool.

samples/client.c
w-henderson: If it expects two FDs then are we just supposed to give it the same one twice or something? That seems awkward

swapnili: I remember having a look into server.c implementation, what I got was it sends just 1 fd.

PR #675, opened 2022-05-10
re-work SGL API

jlevon:
Harmonize and rename the vfu_*sg() APIs to better reflect their functionality:
in our case, there is no mapping happening as part of these calls, they are
merely housekeeping for range splitting, dirty tracking, and so on.

Signed-off-by: John Levon <john.levon@<a href="nutanix.com">nutanix.com</a>>

---

**Stack**:
- #677
- #676
- #675 ⬅

⚠️ *Part of a stack created by [spr](https://github.com/ejoffe/spr). Do not merge manually using the UI - doing so may have unexpected results.*

PR #748, opened 2023-07-21
fix: server sample not marking dirty pages

w-henderson:
The server sample is supposed to demonstrate dirty page logging, but it was not marking dirty pages. This commit both adds client-side dirty page tracking for pages dirtied with `vfu_sgl_write` and server-side dirty page tracking for pages directly dirtied by the server using `vfu_sgl_get/put`.

### DIFF
--- a/samples/client.c
+++ b/samples/client.c
@@ -54,12 +54,6 @@
 /* This is low, so we get testing of vfu_sgl_read/write() chunking. */
 #define CLIENT_MAX_DATA_XFER_SIZE (1024)
 
-/*
- * Our DMA regions are one page in size so we only need one bit to mark them as
- * dirty. This flag goes in the flags field of the vfio_user_dma_map struct.
- */
-#define CLIENT_DIRTY_DMA_REGION_FLAG (1 << 31)
-#define CLIENT_DIRTY_PAGE_TRACKING_ENABLED_FLAG (1 << 30)
 
 static char const *irq_to_str[] = {
     [VFU_DEV_INTX_IRQ] = "INTx",
@@ -67,6 +61,18 @@ static char const *irq_to_str[] = {
     [VFU_DEV_MSIX_IRQ] = "MSI-X",
     [VFU_DEV_ERR_IRQ] = "ERR",
     [VFU_DEV_REQ_IRQ] = "REQ"
+};
+
+struct client_dma_region {
+    uint32_t flags;
+/*
+ * Our DMA regions are one page in size so we only need one bit to mark them as
+ * dirty.
+ */
+#define CLIENT_DIRTY_PAGE_TRACKING_ENABLED_FLAG (1 << 0)
+#define CLIENT_DIRTY_DMA_REGION_FLAG (1 << 1)
+    struct vfio_user_dma_map map;
+    int fd;
 };
 
 void
@@ -567,8 +573,8 @@ wait_for_irq(int irq_fd)
 }
 
 static void
-handle_dma_write(int sock, struct vfio_user_dma_map *dma_regions,
-                 int nr_dma_regions, int *dma_region_fds)
+handle_dma_write(int sock, struct client_dma_region *dma_regions,
+                 int nr_dma_regions)
 {
     struct vfio_user_dma_region_access dma_access;
     struct vfio_user_header hdr;
@@ -595,18 +601,18 @@ handle_dma_write(int sock, struct vfio_user_dma_map *dma_regions,
         off_t offset;
         ssize_t c;
 
-        if (dma_access.addr < dma_regions[i].addr ||
-            dma_access.addr >= dma_regions[i].addr + dma_regions[i].size) {
+        if (dma_access.addr < dma_regions[i].map.addr ||
+            dma_access.addr >= dma_regions[i].map.addr + dma_regions[i].map.size) {
             continue;
         }
 
-        offset = dma_regions[i].offset + dma_access.addr;
+        offset = dma_regions[i].map.offset + dma_access.addr;
 
-        c = pwrite(dma_region_fds[i], data, dma_access.count, offset);
+        c = pwrite(dma_regions[i].fd, data, dma_access.count, offset);
 
         if (c != (ssize_t)dma_access.count) {
             err(EXIT_FAILURE, "failed to write to fd=%d at [%#llx-%#llx)",
-                    dma_region_fds[i], (ull_t)offset,
+                    dma_regions[i].fd, (ull_t)offset,
                     (ull_t)(offset + dma_access.count));
         }
 
@@ -615,7 +621,7 @@ handle_dma_write(int sock, struct vfio_user_dma_map *dma_regions,
          * to mark the newly-dirtied page as dirty.
          */
         if (dma_regions[i].flags & CLIENT_DIRTY_PAGE_TRACKING_ENABLED_FLAG) {
-            assert(dma_regions[i].size == PAGE_SIZE);
+            assert(dma_regions[i].map.size == PAGE_SIZE);
             dma_regions[i].flags |= CLIENT_DIRTY_DMA_REGION_FLAG;
         }
 
@@ -633,8 +639,8 @@ handle_dma_write(int sock, struct vfio_user_dma_map *dma_regions,
 }
 
 static void
-handle_dma_read(int sock, struct vfio_user_dma_map *dma_regions,
-                int nr_dma_regions, int *dma_region_fds)
+handle_dma_read(int sock, struct client_dma_region *dma_regions,
+                int nr_dma_regions)
 {
     struct vfio_user_dma_region_access dma_access, *response;
     struct vfio_user_header hdr;
@@ -661,18 +667,18 @@ handle_dma_read(int sock, struct vfio_user_dma_map *dma_regions,
         off_t offset;
         ssize_t c;
 
-        if (dma_access.addr < dma_regions[i].addr ||
-            dma_access.addr >= dma_regions[i].addr + dma_regions[i].size) {
+        if (dma_access.addr < dma_regions[i].map.addr ||
+            dma_access.addr >= dma_regions[i].map.addr + dma_regions[i].map.size) {
             continue;
         }
 
-        offset = dma_regions[i].offset + dma_access.addr;
+        offset = dma_regions[i].map.offset + dma_access.addr;
 
-        c = pread(dma_region_fds[i], data, dma_access.count, offset);
+        c = pread(dma_regions[i].fd, data, dma_access.count, offset);
 
         if (c != (ssize_t)dma_access.count) {
             err(EXIT_FAILURE, "failed to read from fd=%d at [%#llx-%#llx)",
-                    dma_region_fds[i], (ull_t)offset,
+                    dma_regions[i].fd, (ull_t)offset,
                     (ull_t)offset + dma_access.count);
         }
         break;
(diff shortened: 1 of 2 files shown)


tmakatos:
Could you also add a check in the client that pages do get dirtied by the server?

samples/client.c
```diff
@@ -725,6 +725,10 @@ get_dirty_bitmap(int sock, struct vfio_user_dma_map *dma_region)
            (ull_t)range->iova,
            (ull_t)(range->iova + range->size - 1), bitmap[0]);
 
+    if (bitmap[0] != 0) {
```
jlevon, line 728: isn't this hard-coding that the the dirty pages are in the first 8?
If so, maybe that's fine, but comment?

samples/server.c
```diff
@@ -224,6 +224,7 @@ static void do_dma_io(vfu_ctx_t *vfu_ctx, struct server_data *server_data)
         if (ret < 0) {
             err(EXIT_FAILURE, "vfu_sgl_write failed");
         }
+        vfu_sgl_mark_dirty(vfu_ctx, sg, 1);
```
jlevon, line 227: I'm not convinced this should be the server's responsibility; shouldn't we be marking dirty in vfu_sgl_write() ?

tmakatos: I think we had to make marking dirty explicit because of implications related to quiesce/migration (don't remember the details), PRs #672 and #675 are related.

w-henderson: I think usually if we use `vfu_sgl_get/put` it's done automatically but because we're using `vfu_sgl_write` we have to do it manually?

samples/client.c
w-henderson: Yes, I'm not sure about this - I just used `bitmap[0]` because the above `printf` only prints `bitmap[0]`. Is it correct that the dirty pages are in the first 8?

samples/server.c
jlevon: @tmakatos I don't see how vfu_sgl_write() is any different from vfu_sgl_put() in this situation

samples/client.c
jlevon: I didn't check the sample code, I was hoping you would :)

samples/server.c
tmakatos: > I'm not convinced this should be the server's responsibility; shouldn't we be marking dirty in vfu_sgl_write() ?

`vfu_sgl_write()` doesn't mark the page dirty. This might make sense (I'm not sure it was intentional to begin with), as the vfio-user client knows this, it handles the DMA write itself as it's an explicit message. However, it's not consistent with the expectation that the server must explicitly mark the page dirty when it directly modifies the memory-mapped guest RAM.

I see pros and cons in both cases, @jlevon thoughts?

In any case, I think this part of the code is not fit for purpose as it tries to do two things:

1. demonstrate the use of `vfu_sgl_write`
2. mark guest RAM dirty during migration (via `vfu_sgl_get`/`vfu_sgl_mark_dirty`).

@w-henderson if you do the above then you should be getting dirty pages because of (2), whether or not you need to adapt the sample to consider pages dirtied by (1) will depend what we decide about `VFIO_USER_DMA_WRITE` (whether it should mark and reported pages dirtied by it).

jlevon: > vfu_sgl_write() doesn't mark the page dirty

I'm arguing it should be doing so. vfu_sgl_mark_dirty() exists for consumers that don't always want to vfu_sgl_put() for long-running mappings. That's not relevant for vfu_sgl_write().

w-henderson: I think I probably agree with John that `vfu_sgl_write` should be marking the page dirty - I can't think of any situation in which you would want to be logging dirty pages for a certain region and then intentionally not log the dirtying of a page in that region...

tmakatos: > > In any case, I think this part of the code is not fit for purpose as it tries to do two things:
>
> I'm arguing it should be doing so.

My comment is no longer valid therefore the only fix required is for `vfu_sgl_write` to call `vfu_sgl_mark_dirty`.

w-henderson:
Replaced by #753.

w-henderson:
> Could you also add a check in the client that pages do get dirtied by the server?

I'll do this separately.

w-henderson:
Reopening this one to replace #753.

samples/client.c
w-henderson: Can you clarify what I would be asserting?

samples/client.c
```diff
@@ -1227,7 +1234,13 @@ int main(int argc, char *argv[])
     handle_dma_io(sock, dma_regions, nr_dma_regions, dma_region_fds);
 
     for (i = 0; i < nr_dma_regions; i++) {
-        get_dirty_bitmap(sock, &dma_regions[i]);
+        get_dirty_bitmap(sock, &dma_regions[i], &some_dirty);
+    }
+
+    if (some_dirty) {
```
tmakatos, line 1240: I think this should be just an `assert`.

samples/server.c
tmakatos: My interpretation of [John's comments](https://github.com/nutanix/libvfio-user/pull/753#discussion_r1272506888) was the client does the dirty page tracking in case of `VFIO_USER_DMA_WRITE`, so the server shouldn't have to call `vfu_sgl_mark_dirty` at all. To demonstrate dirty page tracking, you'll have to do what in my previous comment (which I said that was no longer valid, but now it is).

tmakatos:
@w-henderson I'm not sure why you requested another review, you haven't addressed my comment here: https://github.com/nutanix/libvfio-user/pull/748#discussion_r1273634169

w-henderson:
Sorry I missed that - I'm not sure I completely understand your comment, can you explain again what it is I need to do please?

tmakatos:
> Sorry I missed that - I'm not sure I completely understand your comment, can you explain again what it is I need to do please?

Sure I'll rephrase. In the server sample we try to demonstrate two things:

1. Use of `vfu_sgl_write`.
2. Tracking of dirty guest pages during live miration.

The server should not have to use `vfu_sgl_mark_dirty` when using `vfu_sgl_write`, it's the client's responsibility to track dirty pages in that case. Therefore, to demonstrate dirty page tracking in libvfio-user (point 2.), the server _also_ needs to write to guest RAM directly by using `vfu_sgl_get`, write to memory, mark dirty (`vfu_sgl_mark_dirty`), etc. The client should track pages dirtied by:
1. `VFIO_USER_DMA_WRITE` commands, and
2. by explicitly requesting which pages the server dirtied by directly accessing guest RAM, via `VFIO_IOMMU_DIRTY_PAGES_FLAG_GET_BITMAP`.

jlevon:
> The server should not have to use `vfu_sgl_mark_dirty` when using `vfu_sgl_write`, it's the client's responsibility to track dirty pages in that case.

I'd like to add something to memory-mapping.md for that, to describe our expectations here.
And a comment in the vfu_sgl_write() server code.

Perhaps even a comment in handle_dma_write() saying "here we'd mark these pages as dirty".
Going further, actually keep track of dirty pages in client, both from shared memory regions and the handle_dma_write(), then assert it's what we expect later.


 Therefore, to demonstrate dirty page tracking in libvfio-user (point 2.), the server _also_ needs to write to guest RAM directly by using `vfu_sgl_get`, write to memory, mark dirty (`vfu_sgl_mark_dirty`), etc. The client should track pages dirtied by:
> 
>     1. `VFIO_USER_DMA_WRITE` commands, and
> 
>     2. by explicitly requesting which pages the server dirtied by directly accessing guest RAM, via `VFIO_IOMMU_DIRTY_PAGES_FLAG_GET_BITMAP`.

tmakatos (commented):
Can you explain in the commit how the patch achieves what you state in the commit title?

samples/server.c
```diff
@@ -208,21 +209,46 @@ static void do_dma_io(vfu_ctx_t *vfu_ctx, struct server_data *server_data)
...
+            assert(iov.iov_len == (size_t)size);
+            memcpy(iov.iov_base, &buf[i * size], size);
+
+            /*
+             * When directly writing to client memory we are responsible for
+             * tracking dirty pages. We assert that all dirty writes are within
+             * the first page of region 1. In fact, all regions are only one
+             * page in size.
+             */
+            vfu_sgl_mark_dirty(vfu_ctx, sg, 1);
```
jlevon, line 249: I think we shuold probably do vfu_sgl_put() here instead. Servers are supposed to match them up.

samples/server.c
```diff
@@ -231,24 +257,39 @@ static void do_dma_io(vfu_ctx_t *vfu_ctx, struct server_data *server_data)
...
+            }
+        } else {
+            vfu_log(vfu_ctx, LOG_DEBUG, "%s: DIRECT READ   addr %p size %d",
+                    __func__, addr, 2 * size);
+            ret = vfu_sgl_get(vfu_ctx, sg, &iov, 1, 0);
+            if (ret < 0) {
+                err(EXIT_FAILURE, "vfu_sgl_get failed");
+            }
+            assert(iov.iov_len == 2 * (size_t)size);
+            memcpy(&buf[i * size], iov.iov_base, 2 * size);
```
jlevon, line 282: missing vfu_sgl_put

samples/server.c
```diff
@@ -231,24 +257,39 @@ static void do_dma_io(vfu_ctx_t *vfu_ctx, struct server_data *server_data)
...
         }
     }
 
     crc2 = rte_hash_crc(buf, sizeof(buf), 0);
 
     if (crc1 != crc2) {
         errx(EXIT_FAILURE, "DMA write and DMA read mismatch");
+    } else {
+        vfu_log(vfu_ctx, LOG_DEBUG, "%s: %s       success", __func__,
```
jlevon, line 291: nit, odd spacing?

samples/server.c
w-henderson: I agree, but how do we then demonstrate `vfu_sgl_mark_dirty`? Or do we not need to?

samples/server.c
w-henderson: It was for the purposes of aligning the output but I think it might be a bit unnecessary, I'll fix it.

samples/server.c
tmakatos: I think along with `vfu_sgl_put`, we should still have `vfu_sgl_mark_dirty` with a comment explaining that _in this example_ it isn't strictly necessary as it's already marked dirty by the former, and that it can be useful in cases where the server needs to mark guest memory dirty without releasing the mapping.

jlevon: Precisely

samples/server.c
```diff
@@ -208,21 +209,55 @@ static void do_dma_io(vfu_ctx_t *vfu_ctx, struct server_data *server_data)
...
+            vfu_log(vfu_ctx, LOG_DEBUG, "%s: DIRECT WRITE  addr %p size %d",
+                    __func__, addr, size);
+            ret = vfu_sgl_get(vfu_ctx, sg, &iov, 1, 0);
+            if (ret < 0) {
+                err(EXIT_FAILURE, "vfu_sgl_get failed");
+            }
+            assert(iov.iov_len == (size_t)size);
+            memcpy(iov.iov_base, &buf[i * size], size);
+
+
```
tmakatos, line 243: nit: spurious new line

samples/server.c
```diff
@@ -208,21 +209,55 @@ static void do_dma_io(vfu_ctx_t *vfu_ctx, struct server_data *server_data)
...
+            ret = vfu_sgl_get(vfu_ctx, sg, &iov, 1, 0);
+            if (ret < 0) {
+                err(EXIT_FAILURE, "vfu_sgl_get failed");
+            }
+            assert(iov.iov_len == (size_t)size);
+            memcpy(iov.iov_base, &buf[i * size], size);
+
+            
+            /*
+             * When directly writing to client memory we are responsible for
```
tmakatos, line 245: nit: s/we are/the server is

samples/server.c
```diff
@@ -605,14 +656,25 @@ int main(int argc, char *argv[])
...
+                 */
+                do_dma_io(vfu_ctx, &server_data, 1, false);
+                
                 /*
-                 * We also initiate some dummy DMA via an explicit message,
-                 * again to show how DMA is done. This is used if the client's
-                 * RAM isn't mappable or the server implementation prefers it
-                 * this way.  Again, the client expects the server to send DMA
-                 * messages right after it has triggered the IRQs.
+                 * We also do some dummy DMA via explicit messages to show how
```
tmakatos, line 670: I think it would bit a bit more representative to do messaged-based DMA writes to half of the regions and mmap-based to the other half. You can split the regions however you like, e.g. first half vs. second half, or even-based/odd-based. This way you won't have to abuse the `flags` field with `CLIENT_DIRTY_*`, right?

w-henderson: I was using `CLIENT_DIRTY_*` to let the client track dirty pages, how else would the client track this?

samples/client.c
```diff
@@ -62,6 +63,18 @@ static char const *irq_to_str[] = {
     [VFU_DEV_REQ_IRQ] = "REQ"
 };
 
+struct client_dma_region {
+    uint32_t flags;
+/*
```
jlevon, line 68: nit, I'd put these *above* flags.

samples/client.c
```diff
@@ -62,6 +63,18 @@ static char const *irq_to_str[] = {
...
+ * Our DMA regions are one page in size so we only need one bit to mark them as
+ * dirty.
+ */
+#define CLIENT_DIRTY_PAGE_TRACKING_ENABLED_FLAG (1 << 0)
```
jlevon, line 72: nit, "_FLAG" suffix isn't really useful

samples/server.c
```diff
@@ -206,21 +207,54 @@ static void do_dma_io(vfu_ctx_t *vfu_ctx, struct server_data *server_data)
...
+             * When directly writing to client memory the server is responsible
+             * for tracking dirty pages. We assert that all dirty writes are
+             * within the first page of region 1. In fact, all regions are only
+             * one page in size.
+             * 
+             * Note: this is not strictly necessary in this example, since we
+             * later call `vfu_sgl_put`, which marks pages dirty if the SGL was
+             * acquired with `PROT_WRITE`. However, `vfu_sgl_mark_dirty` is
+             * useful in cases where the server needs to mark guest memory dirty
+             * without releasing the memory.
```
jlevon, line 251: nit, "... without putting the sgl".

jlevon (commented):
just tiny nits

jlevon (approved):

PR #749, opened 2023-07-24
remove need for mutex in client sample

w-henderson:
For writes to BAR1 (which come from the fake guest thread), we now send the command with the `no_reply` flag set so we don't interfere with the main thread. This means we don't require a mutex.

### DIFF
--- a/samples/client.c
+++ b/samples/client.c
@@ -453,7 +453,7 @@ configure_irqs(int sock)
 
 static int
 access_region(int sock, int region, bool is_write, uint64_t offset,
-            void *data, size_t data_len)
+              void *data, size_t data_len, bool no_reply)
 {
     static int msg_id = 0xf00f;
     struct vfio_user_region_access send_region_access = {
@@ -492,20 +492,19 @@ access_region(int sock, int region, bool is_write, uint64_t offset,
     }
 
     /*
-     * For all accesses that aren't writes to BAR1, use the standard method.
-     * For writes to BAR1 (which come from the fake guest thread), we send
-     * the command with the `no_reply` flag set so we don't interfere with
-     * the main thread. This means we don't require a mutex.
+     * We set `no_reply` for writes from the fake guest thread so the replies
+     * don't get mixed up with those from the main thread. Since the fake guest
+     * only calls `access_region` with writes, this means we don't need a mutex.
      */
-    if (region != 1 || !is_write) {
+    if (no_reply) {
+        ret = tran_sock_send_iovec(sock, msg_id--, false, op,
+                                   send_iovecs, nr_send_iovecs,
+                                   NULL, 0, 0, true);
+    } else {
         ret = tran_sock_msg_iovec(sock, msg_id--, op,
                                   send_iovecs, nr_send_iovecs,
                                   NULL, 0, NULL,
                                   recv_data, recv_data_len, NULL, 0);
-    } else {
-        ret = tran_sock_send_iovec(sock, msg_id--, false, op,
-                                   send_iovecs, nr_send_iovecs,
-                                   NULL, 0, 0, true);
     }
 
     if (ret != 0) {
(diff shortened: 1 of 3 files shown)


samples/client.c
```diff
@@ -492,12 +491,48 @@ access_region(int sock, int region, bool is_write, uint64_t offset,
         err(EXIT_FAILURE, "failed to alloc recv_data");
     }
 
-    pthread_mutex_lock(&mutex);
-    ret = tran_sock_msg_iovec(sock, msg_id--, op,
```
tmakatos, line 496: Isn't it easier to e.g. extend `tran_sock_send_iovec` to accept a `no_reply` option

jlevon:
I'm missing the context as to why we want to make this change

samples/client.c
jlevon: tran_*() is a private interface not API/ABI

w-henderson:
@jlevon 

Context: https://github.com/nutanix/libvfio-user/pull/747#discussion_r1262517464

In v2 we make more use of the mutex as it's no longer `access_region` that's being called while the fake guest is running, it's also all the migration commands. In my implementation I just moved the mutex in `access_region` to be a global mutex to ensure that no commands can be sent to the server while still waiting for a response. However, if the fake guest just doesn't ask for responses (which it doesn't need to, as it's only writing to BAR1 and, being a sample, the writes shouldn't ever fail), we don't need a mutex at all as all responses are for the main thread.

samples/client.c
tmakatos: > That would make more sense but it would be a a pretty big breaking change, does that make a difference?

I don't think it's such a big change TBH, it's straightforward to modify the callers.

> tran_*() is a private interface not API/ABI

We've already been using tran_*() ins the client/server samples, should we remove it entirely from there or is it "OK" to use them as long as we don't require additional functionality (like the `no_reply` argument I suggested)?

w-henderson: In that case yes I'll change it.

jlevon: we can do whatever we like with our private code...
Although I would hope we're not using tran_*() in server: our APIs are supposed to cover everything a server needs.

jlevon: on that note can we remove tran_sock.h from server.c?

w-henderson: We're not using `train_*()` in `server.c` but yeah we still have the include, should I make another PR to remove that? What am I doing with this one? @jlevon

w-henderson: > on that note can we remove tran_sock.h from server.c?

#752

samples/client.c
```diff
@@ -492,12 +491,23 @@ access_region(int sock, int region, bool is_write, uint64_t offset,
...
-                              recv_data, recv_data_len, NULL, 0);
-    pthread_mutex_unlock(&mutex);
+    /*
+     * For all accesses that aren't writes to BAR1, use the standard method.
+     * For writes to BAR1 (which come from the fake guest thread), we send
+     * the command with the `no_reply` flag set so we don't interfere with
+     * the main thread. This means we don't require a mutex.
+     */
+    if (region != 1 || !is_write) {
+        ret = tran_sock_msg_iovec(sock, msg_id--, op,
```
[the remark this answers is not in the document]
w-henderson: I think it would be a bit weird to add the `no_reply` argument to `tran_sock_msg_iovec` given that all the `tran_sock_msg_*` functions send a message, wait for a reply, and put the reply in a given buffer? We use `tran_sock_send_iovec` for the `no_reply` one because it only does the sending part and doesn't wait for a reply.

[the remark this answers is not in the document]
jlevon: Agreed, I prefer what is here already

tmakatos (approved):

jlevon:
If previously we could access the socket from both threads simultaneously, why is it not still possible? I get that we'll no longer need to get replies in the fake guest thread, but can't we still race at the tran_sock layer? What prevents both threads from using the fd at the same time?

w-henderson:
You're right, I was working under the assumption that `sendmsg` is thread-safe, which [it is not](https://quark.tistory.com/235) - should I just close this PR then?

PR #750, opened 2023-07-24
fix: CRC calculation in client sample

w-henderson:
The previous implementation was using the last CRC value as the initialisation bytes for calculating the new CRC each iteration of the fake guest. This made it impossible to verify the CRC, so I've changed it to always use 0 as the initialisation value.

### DIFF
--- a/samples/client.c
+++ b/samples/client.c
@@ -827,12 +827,10 @@ fake_guest(void *arg)
     int ret;
     char buf[fake_guest_data->bar1_size];
     FILE *fp = fopen("/dev/urandom", "r");
-    uint32_t crc = 0;
 
     if (fp == NULL) {
         err(EXIT_FAILURE, "failed to open /dev/urandom");
     }
-
 
     do {
         ret = fread(buf, fake_guest_data->bar1_size, 1, fp);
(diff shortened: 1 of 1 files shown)


samples/client.c
```diff
@@ -844,7 +844,7 @@ fake_guest(void *arg)
         if (ret != 0) {
             err(EXIT_FAILURE, "fake guest failed to write garbage to BAR1");
         }
-        crc = rte_hash_crc(buf, fake_guest_data->bar1_size, crc);
+        crc = rte_hash_crc(buf, fake_guest_data->bar1_size, 0);
         __sync_synchronize();
```
tmakatos, line 848: In this case it doesn't make sense to calculate it on each loop as we discard the value except the very last one, moving to L851?

w-henderson: Yeah that's true, do you know why it was done like this in the first place? Because I feel like I must be misunderstanding something big as to how the CRC verification was supposed to work if the CRC depends on every BAR1 value past and present?

tmakatos: > Yeah that's true, do you know why it was done like this in the first place? Because I feel like I must be misunderstanding something big as to how the CRC verification was supposed to work if the CRC depends on every BAR1 value past and present?

I guess I did it this way to make it a bit more realistic, clearly I didn't think it through and it only worked by chance.

> And then do we even still need __sync_synchronize since we wait for fake_guest to return before using the CRC value?

Probably not, that could be a separate commit.

tmakatos (approved):

PR #753, opened 2023-07-24
make `vfu_sgl_write` mark appropriate pages as dirty

w-henderson:
See https://github.com/nutanix/libvfio-user/pull/748#discussion_r1272249894.

### DIFF
--- a/include/libvfio-user.h
+++ b/include/libvfio-user.h
@@ -828,7 +828,8 @@ vfu_sgl_get(vfu_ctx_t *vfu_ctx, dma_sg_t *sgl, struct iovec *iov, size_t cnt,
 
 /**
  * Mark scatter/gather entries (previously acquired via vfu_sgl_get())
- * as dirty (written to). This is only necessary if vfu_sgl_put() is not called.
+ * as dirty (written to). This is only necessary if neither of vfu_sgl_put() and
+ * vfu_sgl_write() are called.
  *
  * @vfu_ctx: the libvfio-user context
  * @sg: array of scatter/gather entries to mark as dirty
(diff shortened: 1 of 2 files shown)


lib/libvfio-user.c
```diff
@@ -2273,7 +2273,15 @@ vfu_sgl_write(vfu_ctx_t *vfu_ctx, dma_sg_t *sgl, size_t cnt, void *data)
...
     }
 
-    return vfu_dma_transfer(vfu_ctx, VFIO_USER_DMA_WRITE, sgl, data);
+    int ret = vfu_dma_transfer(vfu_ctx, VFIO_USER_DMA_WRITE, sgl, data);
+
+    if (ret < 0) {
+        return ret;
+    }
+
+    vfu_sgl_mark_dirty(vfu_ctx, sgl, cnt);
```
tmakatos, line 2282: nit: I guess you can just:
```
vfu_sgl_mark_dirty(vfu_ctx, sgl, cnt);
return vfu_dma_transfer(vfu_ctx, VFIO_USER_DMA_WRITE, sgl, data);
```
I don't think it's a problem marking something dirty when it's actually clean in case of error?

tmakatos (commented):
Could you add a unit test for this?

include/libvfio-user.h
```diff
@@ -828,7 +828,8 @@ vfu_sgl_get(vfu_ctx_t *vfu_ctx, dma_sg_t *sgl, struct iovec *iov, size_t cnt,
 
 /**
  * Mark scatter/gather entries (previously acquired via vfu_sgl_get())
- * as dirty (written to). This is only necessary if vfu_sgl_put() is not called.
+ * as dirty (written to). This is only necessary if neither of vfu_sgl_put() and
```
jlevon, line 831: s/and/nor/

lib/libvfio-user.c
jlevon: The marking has to happen after the write finishes.

tmakatos: Oh right otherwise the _client_ can race.

jlevon:
can you change docs/memory-mapping.md ?

lib/libvfio-user.c
jlevon: I was referring to this race:

1. page is marked as dirty
2. dirty page get returns bitmap, clears bit
3. page is copied into client
4. client receives DMA write request, writes to page, but it's no longer known to be dirty


But now I'm starting to think maybe the right thing is for the *client* to handle the dirtiness. It is, after all, the one doing the writes. (Sorry @w-henderson ).

PR #754, opened 2023-07-28
update existing and add new tests for migration v2

w-henderson:
This commit introduces additional testing for migration v2, as well as updating/removing the old tests for migration v1. Additionally, a number of tests are moved from `unit-tests.c` to the Python tests.


test/unit-tests.c
```diff
@@ -402,176 +402,313 @@ static void
...
+        free(LAST_WRITE);
+        LAST_WRITE = NULL;
+    }
+
     return 0;
 }
 
 static void
-test_setup_migration_region_size_ok(void **state)
+test_setup_migration_callbacks(void **state)
```
jlevon, line 531: I think we can test this in python, and drop these units tests too?

test/unit-tests.c
```diff
@@ -402,176 +402,313 @@ static void
...
+    assert_int_equal(LAST_STATE, VFU_MIGR_STATE_STOP);
+    
+    r = handle_device_state(p->v, migr, VFIO_USER_DEVICE_STATE_RUNNING, true);
+    assert_int_equal(0, r);
+    assert_int_equal(LAST_STATE, VFU_MIGR_STATE_RUNNING);
 }
 
 static void
-test_setup_migration_callbacks_without_migration_region(void **state)
+test_handle_mig_data_read(void **state)
```
jlevon, line 570: ditto?

jlevon (commented):
Thanks for adding all this testing!
As I mention, if we can get the necessary test done in python code, let's drop the unit-test.c code, no need to test things twice, and I don't think a lot of that code is of much value.

test/py/test_migration.py
```diff
@@ -1,7 +1,7 @@
 #
 # Copyright (c) 2021 Nutanix Inc. All rights reserved.
 #
-# Authors: Thanos Makatos <thanos@nutanix.com>
+# Authors: John Levon <john.levon@nutanix.com>
```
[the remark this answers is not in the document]
w-henderson: Whoops - I rewrote this file from scratch but I copied all the license information from some other random file so in the diff it looks like I just got rid of Thanos... I'll change it to my name.

test/unit-tests.c
```diff
@@ -402,176 +402,313 @@ static void
 test_migration_state_transitions(void **state UNUSED)
 {
     bool (*f)(uint32_t, uint32_t) = vfio_migr_state_transition_is_valid;
```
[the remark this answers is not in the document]
w-henderson: Yep, I agree.

test/py/test_dirty_pages.py
```diff
@@ -99,60 +92,43 @@ def test_dirty_pages_setup():
 
     msg(ctx, sock, VFIO_USER_DMA_MAP, payload)
 
+    f2 = tempfile.TemporaryFile()
+    f2.truncate(0x10 << PAGE_SHIFT)
 
-def test_dirty_pages_short_write():
```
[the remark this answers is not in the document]
w-henderson: How do we want to check whether `argsz` is correct? Just checking like `msg->in.iov.iov_len == req->argsz` or something else?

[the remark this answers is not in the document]
jlevon: it would depend on the exact semantics of the value as they can differ depending on the operation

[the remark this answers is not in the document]
w-henderson: I was under the impression that it always just means "length of this message"?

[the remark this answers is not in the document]
jlevon: see here
https://github.com/nutanix/libvfio-user/blob/master/docs/vfio-user.rst#message-sizes

I haven't looked at these specific proto commands though

test/py/test_migration.py
```diff
@@ -1,7 +1,7 @@
 #
-# Copyright (c) 2021 Nutanix Inc. All rights reserved.
+# Copyright (c) 2023 Nutanix Inc. All rights reserved.
 #
-# Authors: Thanos Makatos <thanos@nutanix.com>
+# Authors: William Henderson <william.henderson@nutanix.com>
```
jlevon, line 4: add, don't replace, authors please.

jlevon (commented):
just one nit, from an admittedly not super-close look, this looks good

test/py/test_migration.py
w-henderson: Whoops - I rewrote the whole file from scratch so I didn't realise it was going to look like I got rid of Thanos! I'll fix that.

test/py/test_migration.py
```diff
@@ -28,139 +29,504 @@
...
+    assert result[len(vfio_user_mig_data()):] == data
+
+
+def test_handle_mig_data_read_too_long():
+    global read_data
+
+    transition_to_state(VFIO_USER_DEVICE_STATE_RUNNING)
+
+    payload = vfio_user_mig_data(
+        argsz=len(vfio_user_mig_data()) + 8,
```
tmakatos, line 309: ditto about using `len(data)` instead of 8

test/py/test_migration.py
```diff
@@ -28,139 +29,504 @@
...
+    transition_to_state(VFIO_USER_DEVICE_STATE_PRE_COPY)
+    read_data = bytes([1, 2, 3, 4, 5, 6, 7, 8])
+    msg(ctx, sock, VFIO_USER_MIG_DATA_READ, payload, expect=errno.EINVAL)
+
+
+def test_handle_mig_data_read_invalid_state():
+    global read_data
+
+    payload = vfio_user_mig_data(
+        argsz=len(vfio_user_mig_data()) + 4,
```
tmakatos, line 326: ditto about using len(data) instead of 4

test/py/test_migration.py
```diff
@@ -28,139 +29,504 @@
...
+        size=4
+    )
+
+    data = bytes([1, 2, 3, 4])
+
+    transition_to_state(VFIO_USER_DEVICE_STATE_RUNNING)
+    read_data = data
+    msg(ctx, sock, VFIO_USER_MIG_DATA_READ, payload, expect=errno.EINVAL)
+
+    transition_to_state(VFIO_USER_DEVICE_STATE_STOP)
```
tmakatos, line 336: Shouldn't we be testing the other states as well (apart from the `_COPY` ones)?

test/py/test_migration.py
```diff
@@ -28,139 +29,504 @@
...
-    print("received reply")
+def test_handle_mig_data_read_failed_callback():
+    global fail_callbacks
+
+    transition_to_state(VFIO_USER_DEVICE_STATE_PRE_COPY)
+
+    fail_callbacks = True
+
+    payload = vfio_user_mig_data(
+        argsz=len(vfio_user_mig_data()) + 4,
```
tmakatos, line 349: ditto about using len(data) instead of 4

test/py/test_migration.py
```diff
@@ -28,139 +29,504 @@
...
+        size=4
+    )
+
+    msg(ctx, sock, VFIO_USER_MIG_DATA_READ, payload, expect=errno.EINVAL)
+
+    fail_callbacks = False
+
+
+def test_handle_mig_data_read_short_write():
+    payload = struct.pack("I", 8)
```
tmakatos, line 359: Can you explain a bit more?

test/py/test_migration.py
```diff
@@ -28,139 +29,504 @@
...
+
+    msg(ctx, sock, VFIO_USER_MIG_DATA_READ, payload, expect=errno.EINVAL)
+
+
+def test_handle_mig_data_write():
+    payload = vfio_user_mig_data(
+        argsz=len(vfio_user_mig_data()) + 4,
```
tmakatos, line 366: ditto about using len(data) instead of 4

test/py/test_migration.py
```diff
@@ -28,139 +29,504 @@
...
+    data = bytes([1, 2, 3, 4])
+
+    transition_to_state(VFIO_USER_DEVICE_STATE_RESUMING)
+    msg(ctx, sock, VFIO_USER_MIG_DATA_WRITE, bytes(payload) + data)
+    assert write_data == data
+
+
+def test_handle_mig_data_write_too_long():
+    payload = vfio_user_mig_data(
+        argsz=len(vfio_user_mig_data()) + 8,
```
tmakatos, line 379: ditto about using len(data) instead of 8

test/py/test_migration.py
```diff
@@ -28,139 +29,504 @@
...
+
+    data = bytes([1, 2, 3, 4, 5, 6, 7, 8])
+    transition_to_state(VFIO_USER_DEVICE_STATE_RESUMING)
+    msg(ctx, sock, VFIO_USER_MIG_DATA_WRITE, bytes(payload) + data,
+        expect=errno.EINVAL)
+
+
+def test_handle_mig_data_write_invalid_state():
+    payload = vfio_user_mig_data(
+        argsz=len(vfio_user_mig_data()) + 4,
```
tmakatos, line 395: ditto about using len(data) instead of 4

test/py/test_migration.py
```diff
@@ -28,139 +29,504 @@
...
+    # data transfer size is 4 bytes. Here we test to check that a transfer of 8
+    # bytes fails.
+
+    data = bytes([1, 2, 3, 4, 5, 6, 7, 8])
+    transition_to_state(VFIO_USER_DEVICE_STATE_RESUMING)
+    msg(ctx, sock, VFIO_USER_MIG_DATA_WRITE, bytes(payload) + data,
+        expect=errno.EINVAL)
+
+
+def test_handle_mig_data_write_invalid_state():
```
tmakatos, line 393: nit: move before `test_handle_mig_data_write_too_long` so that all tests that test state are grouped together

test/py/test_migration.py
```diff
@@ -28,139 +29,504 @@
...
+
+def test_handle_mig_data_write_failed_callback():
+    global fail_callbacks
+
+    transition_to_state(VFIO_USER_DEVICE_STATE_RESUMING)
+
+    fail_callbacks = True
+
+    payload = vfio_user_mig_data(
+        argsz=len(vfio_user_mig_data()) + 4,
```
tmakatos, line 414: ditto about using len(data) instead of 4

test/py/test_migration.py
```diff
@@ -28,139 +29,504 @@
...
+
+    data = bytes([1, 2, 3, 4])
+
+    msg(ctx, sock, VFIO_USER_MIG_DATA_WRITE, bytes(payload) + data,
+        expect=errno.EINVAL)
+
+    fail_callbacks = False
+
+
+def test_handle_mig_data_write_short_write():
```
tmakatos, line 426: Same comment as in `test_handle_mig_data_read_short_write`

test/py/test_migration.py
```diff
@@ -28,139 +29,504 @@
...
+    payload = struct.pack("I", 8)
+
+    msg(ctx, sock, VFIO_USER_MIG_DATA_WRITE, payload, expect=errno.EINVAL)
+
+
+def test_device_feature_migration():
+    payload = vfio_user_device_feature(
+        argsz=len(vfio_user_device_feature()) + 8,
```
tmakatos, line 434: 8?

test/py/test_migration.py
```diff
@@ -28,139 +29,504 @@
...
+    fail_callbacks = False
+
+
+def test_handle_mig_data_write_short_write():
+    payload = struct.pack("I", 8)
+
+    msg(ctx, sock, VFIO_USER_MIG_DATA_WRITE, payload, expect=errno.EINVAL)
+
+
+def test_device_feature_migration():
```
tmakatos, line 432: Either the test name should explain what's being tested or there should be a docstring.

test/py/test_migration.py
```diff
@@ -28,139 +29,504 @@
...
+    )
+
+    result = msg(ctx, sock, VFIO_USER_DEVICE_FEATURE, payload)
+    _, result = vfio_user_device_feature.pop_from_buffer(result)
+    flags, = struct.unpack("Q", result)
+
+    assert flags == VFIO_MIGRATION_STOP_COPY | VFIO_MIGRATION_PRE_COPY
+
+
+def test_device_feature_short_write():
```
tmakatos, line 445: Same comment as in `test_handle_mig_data_read_short_write`.

test/py/test_migration.py
```diff
@@ -28,139 +29,504 @@
...
+    payload = struct.pack("I", 8)
+
+    msg(ctx, sock, VFIO_USER_DEVICE_FEATURE, payload, expect=errno.EINVAL)
+
+
+def test_device_feature_unsupported_operation():
+    payload = vfio_user_device_feature(
+        argsz=len(vfio_user_device_feature()) + 8,
```
tmakatos, line 453: 8?

test/py/test_migration.py
```diff
@@ -28,139 +29,504 @@
...
+        flags=VFIO_DEVICE_FEATURE_GET | VFIO_DEVICE_FEATURE_MIGRATION
+    )
+
+    msg(ctx, sock, VFIO_USER_DEVICE_FEATURE, payload, expect=errno.EINVAL)
+
+
+def test_device_feature_bad_argsz_get_dma():
+    argsz = len(vfio_user_device_feature()) + \
+            len(vfio_user_device_feature_dma_logging_report()) + \
+            8  # bitmap size
```
tmakatos, line 481: Should calculate instead?

test/py/test_migration.py
```diff
@@ -28,139 +29,504 @@
...
+
+    msg(ctx, sock, VFIO_USER_DEVICE_FEATURE, payload, expect=errno.EINVAL)
+
+
+def test_migration_cleanup():
+    disconnect_client(ctx, sock)
+    vfu_destroy_ctx(ctx)
 
-# ex: set tabstop=4 shiftwidth=4 softtabstop=4 expandtab: #
+# ex: set tabstop=4 shiftwidth=4 softtabstop=4 expandtab:
```
tmakatos, line 532: spurious change

test/py/test_quiesce.py
```diff
@@ -195,36 +195,6 @@ def test_allowed_funcs_in_quiesced_dma_unregister_busy(mock_quiesce,
...
 
 
-@patch('libvfio_user.migr_trans_cb', side_effect=_side_effect)
-@patch('libvfio_user.quiesce_cb')
-def test_allowed_funcs_in_quiesed_migration(mock_quiesce,
-                                            mock_trans):
-
-    global ctx, sock
-    _map_dma_region(ctx, sock)
-    data = VFIO_DEVICE_STATE_V1_SAVING.to_bytes(c.sizeof(c.c_int), 'little')
```
tmakatos, line 205: Shouldn't these be replaced with which migration states allow these functions in quiesced state?

test/py/test_request_errors.py
```diff
@@ -189,24 +185,29 @@ def test_disconnected_socket_quiesce_busy(mock_quiesce):
...
-    # read the get_pending_bytes register, it should close the socket causing
-    # the reply to fail
-    read_region(ctx, sock, VFU_PCI_DEV_MIGR_REGION_IDX,
-                vfio_user_migration_info.pending_bytes.offset,
-                vfio_user_migration_info.pending_bytes.size, rsp=False,
-                busy=True)
+    mock_migr_trans_cb.side_effect = migr_trans_cb_side_effect
+
+    # change the state, it should close the socket causing the reply to fail
+    feature = vfio_user_device_feature(
```
tmakatos, line 201: Maybe I've wrong but after having reviewed all this code I start feeling this is getting repeated, if that's the case then could we create helper function?

test/unit-tests.c
```diff
@@ -583,19 +407,16 @@ test_device_is_stopped_and_copying(UNUSED void **state)
     size_t i;
     struct migration migration;
     vfu_ctx.migration = &migration;
-    for (i = 0; i < ARRAY_SIZE(migr_states); i++) {
-        if (migr_states[i].name == NULL) {
-            continue;
-        }
-        migration.info.device_state = i;
+    for (i = 0; i < 8; i++) {
```
tmakatos, line 410: 8?

tmakatos (commented):
just a couple of comments

test/py/libvfio_user.py
```diff
@@ -678,10 +724,15 @@ def connect_sock():
...
-def connect_client(ctx):
+def connect_client(ctx, max_data_xfer_size=None):
     sock = connect_sock()
 
-    json = b'{ "capabilities": { "max_msg_fds": 8 } }'
+    if max_data_xfer_size is None:
+        json = b'{ "capabilities": { "max_msg_fds": 8 } }'
+    else:
+        json = b'{ "capabilities": { "max_msg_fds": 8, "max_data_xfer_size": '\
+            + str(max_data_xfer_size).encode("utf-8") + b' } }'
```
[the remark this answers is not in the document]
w-henderson: Yes, I'm not sure why I didn't think of that, much less horrible!

test/py/libvfio_user.py
```diff
@@ -1244,4 +1253,9 @@ def fds_are_same(fd1: int, fd2: int) -> bool:
     return s1.st_dev == s2.st_dev and s1.st_ino == s2.st_ino
 
 
+def get_bitmap_size(size: int, pgsize: int) -> int:
```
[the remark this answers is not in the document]
w-henderson: It does the same as `get_bitmap_size` in `common.h` but that's internal so I thought it made more sense to have a Python version...? It doesn't assume it's a 64-bit machine - the bitmap size is always a multiple of 64 bits on all architectures.

test/py/test_dirty_pages.py
```diff
@@ -99,60 +92,43 @@ def test_dirty_pages_setup():
 
     msg(ctx, sock, VFIO_USER_DMA_MAP, payload)
 
+    f2 = tempfile.TemporaryFile()
```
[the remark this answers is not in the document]
w-henderson: There definitely was a reason at some point but not any more, I'll remove it.

test/py/test_dirty_pages.py
```diff
@@ -99,60 +92,43 @@ def test_dirty_pages_setup():
...
-        expect=errno.EINVAL)
+    feature = vfio_user_device_feature(
+        argsz=len(vfio_user_device_feature()) +
+              len(vfio_user_device_feature_dma_logging_control()) +
+              len(ranges),
+        flags=VFIO_DEVICE_FEATURE_DMA_LOGGING_START | VFIO_DEVICE_FEATURE_SET)
 
+    payload = vfio_user_device_feature_dma_logging_control(
+        page_size=page_size,
+        num_ranges=(1 if addr is not None else 0),
```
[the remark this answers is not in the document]
w-henderson: `len(ranges)` would give us the length in bytes but we want 1 or 0.

test/py/test_dirty_pages.py
```diff
@@ -161,155 +137,85 @@ def test_dirty_pages_start():
...
-    dirty_pages = vfio_user_dirty_pages(argsz=argsz,
-        flags=VFIO_IOMMU_DIRTY_PAGES_FLAG_GET_BITMAP)
-    bitmap = vfio_user_bitmap(pgsize=PAGE_SIZE, size=1)
-    br = vfio_user_bitmap_range(iova=0x10 << PAGE_SHIFT,
-                                size=0x10 << PAGE_SHIFT, bitmap=bitmap)
-
-    payload = bytes(dirty_pages) + bytes(br)
+def test_dirty_pages_start_different_pgsize():
+    start_logging(page_size=0, expect=errno.EINVAL)
+    start_logging(page_size=PAGE_SIZE >> 1, expect=errno.EINVAL)
```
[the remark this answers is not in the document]
w-henderson: Yep, once we start logging at a certain page size, all requests to start logging at a different page size are rejected. I'll add a comment.

test/py/test_dirty_pages.py
```diff
@@ -161,155 +137,85 @@ def test_dirty_pages_start():
...
+        assert b == 0
 
-    dirty_pages = vfio_user_dirty_pages(argsz=argsz,
-        flags=VFIO_IOMMU_DIRTY_PAGES_FLAG_GET_BITMAP)
-    bitmap = vfio_user_bitmap(pgsize=PAGE_SIZE, size=8)
-    br = vfio_user_bitmap_range(iova=0x40 << PAGE_SHIFT,
-                                size=0x10 << PAGE_SHIFT, bitmap=bitmap)
 
-    payload = bytes(dirty_pages) + bytes(br)
+def get_dirty_page_bitmap(addr=0x10 << PAGE_SHIFT, length=0x10 << PAGE_SHIFT,
```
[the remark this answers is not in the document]
w-henderson: Yep, I can refactor the entire `test_dirty_pages_get_unmodified` to this!
```c
bitmap = get_dirty_page_bitmap()
assert bitmap == 0
```

test/py/test_dirty_pages.py
```diff
@@ -427,68 +354,48 @@ def test_dirty_pages_get_modified():
...
 
-def test_dirty_pages_stop():
-    # FIXME we have a memory leak as we don't free dirty bitmaps when
-    # destroying the context.
-    payload = vfio_user_dirty_pages(argsz=len(vfio_user_dirty_pages()),
-                                    flags=VFIO_IOMMU_DIRTY_PAGES_FLAG_STOP)
-
-    msg(ctx, sock, VFIO_USER_DIRTY_PAGES, payload)
-
+def test_dirty_pages_invalid_address():
```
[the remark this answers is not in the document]
w-henderson: I'll rename it to `test_dirty_pages_invalid_arguments`.

test/py/libvfio_user.py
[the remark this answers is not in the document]
tmakatos: It's fine to have a Python equivalent, we still need some docstring.

> the bitmap size is always a multiple of 64 bits on all architectures.

Is this documented somewhere? We should use a define.

test/py/test_migration.py
```diff
@@ -28,139 +29,504 @@
...
+    ret = vfu_setup_device_migration_callbacks(ctx, cbs)
     assert ret == 0
 
     vfu_setup_device_quiesce_cb(ctx)
 
     ret = vfu_realize_ctx(ctx)
     assert ret == 0
 
-    sock = connect_client(ctx)
+    sock = connect_client(ctx, 4)
```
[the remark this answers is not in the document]
w-henderson: It's the max data transfer size, I'll make that clearer:
```py
max_data_xfer_size = 4 # for later tests
sock = connect_client(ctx, max_data_xfer_size)
```

test/py/test_migration.py
```diff
@@ -28,139 +29,504 @@
...
+    - The path cannot have saving group states as interior arcs, only start/end
+      states.
+
+    This test implements a breadth-first search to ensure that the paths taken
+    by the implementation correctly follow these rules.
     """
 
-    global ctx, sock
+    # states (vertices)
+    V = {VFIO_USER_DEVICE_STATE_ERROR, VFIO_USER_DEVICE_STATE_STOP,
```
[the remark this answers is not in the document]
w-henderson: Wouldn't that involve referring to all the states as `State.VFIO_USER_DEVICE_STATE_*` everywhere or something like that?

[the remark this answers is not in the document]
w-henderson: Actually we can use `E.keys()` like you suggested below.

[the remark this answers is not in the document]
tmakatos: > Wouldn't that involve referring to all the states as State.VFIO_USER_DEVICE_STATE_* everywhere or something like that?

It could be `VFIO_USER_DEVICE_STATE_.STOP`, `VFIO_USER_DEVICE_STATE_.PRE_COPY`, etc. Regardless of using `E.keys()`, it nice to have.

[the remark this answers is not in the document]
tmakatos: > but in my experience it is conventional and clearer to call the set of vertices V and the set of edges E. 

OK, you can keep it as is then.

test/py/test_migration.py
```diff
@@ -28,139 +29,504 @@
...
-    Tests transitioning to the saving state.
+    The spec dictates that complex state transitions are to be implemented as
+    combinations of the defined direct transitions, with the path selected
+    according to the following rules:
+
+    - Select the shortest path.
+    - The path cannot have saving group states as interior arcs, only start/end
+      states.
+
+    This test implements a breadth-first search to ensure that the paths taken
```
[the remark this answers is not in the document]
w-henderson: I guess it's not really searching for anything, just traversing the graph to find the shortest paths between every pair of nodes. Maybe I should say breadth-first traversal? But is that a term that's actually used?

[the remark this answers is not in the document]
tmakatos: I think BFS is for searching, perhaps it's pre-order traversal?

test/py/test_migration.py
```diff
@@ -28,139 +29,504 @@
...
+            VFIO_USER_DEVICE_STATE_RUNNING,
+            VFIO_USER_DEVICE_STATE_STOP_COPY
+        },
+        VFIO_USER_DEVICE_STATE_PRE_COPY_P2P: set()
+    }
+
+    # "saving states" which cannot be internal arcs
+    S = {VFIO_USER_DEVICE_STATE_PRE_COPY, VFIO_USER_DEVICE_STATE_STOP_COPY}
+
+    for source in V:
```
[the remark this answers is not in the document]
w-henderson: I'll add a bunch.

test/py/test_migration.py
```diff
@@ -28,139 +29,504 @@
...
-    global ctx, sock
-    mock_quiesce
 
-    data = VFIO_DEVICE_STATE_V1_SAVING.to_bytes(c.sizeof(c.c_int), 'little')
-    write_region(ctx, sock, VFU_PCI_DEV_MIGR_REGION_IDX, offset=0,
-                 count=len(data), data=data, rsp=False,
-                 busy=True)
+def test_migration_failed_callback():
+    global fail_callbacks
+    fail_callbacks = True
```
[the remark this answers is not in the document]
w-henderson: The Python callbacks don't set `errno` I don't think, how do we do that?

test/py/test_migration.py
```diff
@@ -28,139 +29,504 @@
...
+    transition_to_state(VFIO_USER_DEVICE_STATE_RUNNING, expect=errno.EINVAL)
+    fail_callbacks = False
 
-    ret = vfu_device_quiesced(ctx, 0)
-    assert ret == 0
 
-    get_reply(sock)
+def test_migration_get_state():
```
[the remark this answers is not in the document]
w-henderson: No, we never send a `VFIO_USER_DEVICE_FEATURE` message with the `GET` flag for the `VFIO_USER_DEVICE_FEATURE_MIG_STATE` feature.

test/py/test_migration.py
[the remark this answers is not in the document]
tmakatos: Look `fail_with_errno` in libvfio-user.py.

test/py/test_migration.py
w-henderson: It's only STOP, RUNNING and RESUMING that we can a) legally transition into and b) not call MIG_DATA_READ, so I'll add an extra test for RESUMING.

test/py/test_migration.py
w-henderson: I'm not sure why I did it like that, I'll explicitly make it send a `vfio_user_mig_data` payload with the data missing.

w-henderson: Upon further investigation it has arisen that there is an actual bug and it will just start giving out-of-bounds data to the write callback 😳 I will fix that in the implementation PR.

test/py/test_migration.py
```diff
@@ -28,139 +29,504 @@
...
+    assert result[len(vfio_user_mig_data()):] == data
+
+    transition_to_state(VFIO_USER_DEVICE_STATE_STOP_COPY)
+    read_data = data
+    result = msg(ctx, sock, VFIO_USER_MIG_DATA_READ, payload)
+    assert len(result) == argsz
+    assert result[len(vfio_user_mig_data()):] == data
+
+
+def test_handle_mig_data_read_too_long():
```
tmakatos, line 303: Please add docstring.

test/py/test_migration.py
```diff
@@ -28,139 +29,504 @@
...
 
-    global ctx, sock
+def test_handle_mig_data_read():
+    global read_data
 
-    data = VFIO_DEVICE_STATE_V1_RUNNING.to_bytes(c.sizeof(c.c_int), 'little')
-    write_region(ctx, sock, VFU_PCI_DEV_MIGR_REGION_IDX, offset=0,
-                 count=len(data), data=data, rsp=False,
-                 busy=True)
+    transition_to_state(VFIO_USER_DEVICE_STATE_RUNNING)
```
tmakatos, line 279: Can we use a loop of the core code on all valid states?

test/py/test_migration.py
```diff
@@ -28,139 +29,504 @@
...
-@patch('libvfio_user.migr_trans_cb', side_effect=fail_with_errno(errno.ENOTTY))
-def test_migration_trans_async_err(mock_trans, mock_quiesce):
-    """
-    Tests writing to the migration state register, the device not being able to
-    immediately quiesce, and then finally the device failing to transition to
-    the new migration state.
-    """
 
-    global ctx, sock
+def test_handle_mig_data_read():
```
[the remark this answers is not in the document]
tmakatos: Actually, it would make it easier to implement, maintain, and read if we executed the same code on all states and then validated the result (assuming we convert the states into an enum):
```
for state in VFIO_USER_DEVICE_STATES:
    transition_to_state(state.value)
     result = read_migr_data()
     if state in valid_states:
         assert result == True
     else:
         assert result == False
```

test/py/libvfio_user.py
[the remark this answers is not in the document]
w-henderson: In VFIO the bitmap is an array of `uint64_t`s so we just follow that, I don't think it's explicitly documented though...

test/py/test_migration.py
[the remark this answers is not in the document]
w-henderson: I'm pretty sure BFS is just a general algorithm for traversing a graph (whether or not you're searching for anything) - pre-order traversal is limited to trees.

test/py/libvfio_user.py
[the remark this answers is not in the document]
jlevon: can you file an issue on documenting this please?

test/py/test_migration.py
w-henderson: I'll add `get` in the title but otherwise it's pretty descriptive, it's testing calling GET on the VFIO_USER_DEVICE_FEATURE_MIGRATION feature.

test/py/test_migration.py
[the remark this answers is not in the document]
jlevon: I agree with Thanos that "search" is a bit weird, let's just say breadth-first traversal?

test/py/test_migration.py
w-henderson: `len(vfio_user_device_feature_migration())`

test/py/libvfio_user.py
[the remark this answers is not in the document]
tmakatos: VFIO and vfio-user say:
```
The mapping of IOVA to bits is given by:
bitmap[(addr - iova)/page_size] & (1ULL << (addr % 64))
```
However, VFIO earlier says:
```
bitmap is a pointer to an array of u64s that will hold the output bitmap with 1 bit reporting a page_size unit of IOVA.
```
But vfio-user doesn't. We need to clarify this on the vfio-user spec.

test/py/test_quiesce.py
w-henderson: You're absolutely right, I also didn't implement this in the implementation branch but I've done that now (will push in a moment).

test/unit-tests.c
w-henderson: `VFIO_USER_DEVICE_NUM_STATES`.

test/py/test_migration.py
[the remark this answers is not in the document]
tmakatos: > pre-order traversal is limited to trees

I had assumed it is a tree, but that might not be true. In fact, I have no idea what the graph looks like, it would be nice if we could visualize it: https://github.com/nutanix/libvfio-user/issues/778

test/py/test_migration.py
```diff
@@ -28,139 +29,504 @@
...
+    # When we set up the tests at the top of this file we specify that the max
+    # data transfer size is 4 bytes. Here we test to check that a transfer of 8
+    # bytes fails.
+
+    transition_to_state(VFIO_USER_DEVICE_STATE_PRE_COPY)
+    read_data = bytes([1, 2, 3, 4, 5, 6, 7, 8])
+    msg(ctx, sock, VFIO_USER_MIG_DATA_READ, payload, expect=errno.EINVAL)
+
+
+def test_handle_mig_data_read_invalid_state():
```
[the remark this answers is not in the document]
w-henderson: Since this is now covered in `test_handle_mig_data_read` this doesn't apply any more.

test/py/test_migration.py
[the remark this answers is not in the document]
w-henderson: Unless you feel really strongly about it I'd rather keep it as "breadth-first search" because that's the proper name of the algorithm?

test/py/libvfio_user.py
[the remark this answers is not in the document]
w-henderson: #780

test/py/test_migration.py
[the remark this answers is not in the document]
tmakatos: I'm fine with it.

test/py/test_migration.py
```diff
@@ -276,181 +314,193 @@ def test_migration_get_state():
...
-        argsz=len(vfio_user_mig_data()) + 4,
-        size=4
+        argsz=len(vfio_user_mig_data()) + len(data),
+        size=len(data)
     )
 
-    data = bytes([1, 2, 3, 4])
-
-    transition_to_state(VFIO_USER_DEVICE_STATE_RESUMING)
+    transition_to_state(ctx, sock, VFIO_USER_DEVICE_STATE_RESUMING)
```
tmakatos, line 404: nit: since `ctx` and `sock` are global, can we have a helper method local to this file that passes them to `transition_to_state`?

test/py/test_migration.py
```diff
@@ -276,181 +314,193 @@ def test_migration_get_state():
...
     msg(ctx, sock, VFIO_USER_MIG_DATA_WRITE, bytes(payload) + data)
     assert write_data == data
 
 
-def test_handle_mig_data_write_too_long():
+def test_handle_mig_data_write_invalid_state():
+    data = bytes([1, 2, 3, 4])
+
     payload = vfio_user_mig_data(
```
tmakatos, line 412: Introduce helper function that returns the payload given the data?

test/py/test_migration.py
```diff
@@ -276,181 +314,193 @@ def test_migration_get_state():
...
 
-def test_handle_mig_data_write_invalid_state():
+def test_handle_mig_data_write_too_long():
+    """
+    When we set up the tests at the top of this file we specify that the max
+    data transfer size is 4 bytes. Here we test to check that a transfer of 8
+    bytes fails.
+    """
+
+    data = bytes([1, 2, 3, 4, 5, 6, 7, 8])
```
tmakatos, line 429: I'd use `data = bytes([i for i in range(0, max_data_xfer_size + 1)])`

test/py/test_migration.py
```diff
@@ -276,181 +314,193 @@ def test_migration_get_state():
...
-    payload = struct.pack("I", 8)
+    payload = vfio_user_device_feature(
+        argsz=len(vfio_user_device_feature()) +
+              len(vfio_user_device_feature_migration()),
+        flags=VFIO_DEVICE_FEATURE_GET | VFIO_DEVICE_FEATURE_MIGRATION
+    )
 
-    msg(ctx, sock, VFIO_USER_DEVICE_FEATURE, payload, expect=errno.EINVAL)
+    payload = bytes(payload)
+    assert len(payload) == 8
```
tmakatos, line 494: 8?

test/py/test_quiesce.py
```diff
@@ -195,6 +195,31 @@ def test_allowed_funcs_in_quiesced_dma_unregister_busy(mock_quiesce,
     mock_dma_unregister.assert_called_once_with(ctx, mock.ANY)
 
 
+@patch('libvfio_user.migr_trans_cb', side_effect=_side_effect)
+@patch('libvfio_user.quiesce_cb')
+def test_allowed_funcs_in_quiesed_migration(mock_quiesce,
```
tmakatos, line 200: nit: quiesced

test/py/test_quiesce.py
```diff
@@ -195,6 +195,31 @@ def test_allowed_funcs_in_quiesced_dma_unregister_busy(mock_quiesce,
...
+                                            mock_trans):
+
+    global ctx, sock
+    _map_dma_region(ctx, sock)
+    transition_to_state(ctx, sock, VFIO_USER_DEVICE_STATE_STOP)
```
tmakatos, line 205: This implies that STOP is the only state we can transition to once quiesced. If this is true then we should have negative tests for all other migration states, no?

test/py/test_dirty_pages.py
```diff
@@ -124,69 +126,45 @@ def start_logging(addr=None, length=None, page_size=PAGE_SIZE, expect=0):
...
+    start_logging(page_size=PAGE_SIZE >> 1, expect=errno.EINVAL)
+    start_logging(page_size=PAGE_SIZE << 1, expect=errno.EINVAL)
 
 
 def get_dirty_page_bitmap(addr=0x10 << PAGE_SHIFT, length=0x10 << PAGE_SHIFT,
                           page_size=PAGE_SIZE, expect=0):
+    """
+    Get the dirty page bitmap from the server for the given region and page
+    size as a 64-bit integer. This only works for bitmaps that fit within a
+    64-bit integer.
```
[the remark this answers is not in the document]
w-henderson: The function returns the bitmap as an integer so we can easily compare it with binary literals in the tests, I'll clarify that this is a test function limitation.

test/py/test_migration.py
```diff
@@ -204,26 +218,40 @@ def test_migration_shortest_state_transition_paths():
...
+
     # "saving states" which cannot be internal arcs
-    S = {VFIO_USER_DEVICE_STATE_PRE_COPY, VFIO_USER_DEVICE_STATE_STOP_COPY}
+    saving_states = {VFIO_USER_DEVICE_STATE_PRE_COPY,
+                     VFIO_USER_DEVICE_STATE_STOP_COPY}
 
+    # Consider each vertex in turn to be the start state, that is, the state
+    # we are transitioning from.
     for source in V:
         back = {v: None for v in V}
```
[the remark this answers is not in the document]
w-henderson: Backpointers for each node (the previous node in the shortest path from the start to the node). We use this dictionary to retrieve the shortest paths to compare with the server's ones.

Suppose the shortest path from a to c was `a -> b -> c`, then `back[b] == a`.

test/py/test_migration.py
```diff
@@ -276,181 +314,193 @@ def test_migration_get_state():
...
+    data = bytes([1, 2, 3, 4])
 
-    msg(ctx, sock, VFIO_USER_MIG_DATA_READ, payload, expect=errno.EINVAL)
+    payload = vfio_user_mig_data(
+        argsz=len(vfio_user_mig_data()) + len(data),
+        size=len(data)
+    )
+
+    payload = bytes(payload)
+    assert len(payload) == 8
```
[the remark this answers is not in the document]
w-henderson: We can get rid of this and use `payload[0:len(payload) - 1]` instead because it's clearer that we're excluding one byte.

test/py/test_migration.py
w-henderson: This was to make it clear that the payload has length 8 and we're sending 7 bytes of it, I've changed this to now explicitly send one byte less than the payload length which I should've done in the first place!

test/py/test_quiesce.py
w-henderson: I have to be honest I don't really understand how this works - I just copied across the test from the previous version and changed it to the migration v2 API. I also changed the previous test's transition to STOP_COPY to a transition to STOP because of the following assertion that transition is called once, and it gets called multiple times because of the previous state if we transition to STOP_COPY.

In the actual code (both in v1 and v2) we allow any state transition to happen once quiesced.

test/py/test_migration.py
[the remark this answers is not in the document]
tmakatos: I can't get github to show enough context so I've no idea what `payload[0:len(payload) - 1]` means.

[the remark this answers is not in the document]
w-henderson: We're testing a short write so we create a valid payload, convert it to bytes, and `payload[0:len(payload) - 1]` gives us the payload with the final byte missing.

test/py/test_quiesce.py
tmakatos: > I just copied across the test from the previous version

Do you mean migration v1?

w-henderson: Yes

tmakatos: I `git diff`ed against master and can't find such a test?

w-henderson: https://github.com/nutanix/libvfio-user/blob/master/test/py/test_quiesce.py#L200

test/py/test_migration.py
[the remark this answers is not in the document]
tmakatos: Isn't `payload[:-1]` simpler?

test/py/test_quiesce.py
tmakatos: Right, why does it show that you're adding this function though?

test/py/test_migration.py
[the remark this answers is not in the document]
w-henderson: Yes, I'll change it.

test/py/test_quiesce.py
tmakatos: nvm, github being github, need to manually configure the base commit to show the right thing

tmakatos: > and it gets called multiple times because of the previous state if we transition to STOP_COPY.

Not sure I follow, is it executed mulitple times because of the intermediate transitions?

I realize that the base code isn't well-written. Ideally we should refactor it into a single function that attempts to execute allowed/disallowed API calls during queisce and see whether it was permitted or not.

test/py/test_migration.py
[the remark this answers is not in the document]
tmakatos: The implementation isn't sufficiently explained, the fact that BFS isn't enough because this isn't a textbook implementation. I would suggest refactorin and explaining further, e.g. the `while` loop at the beggining could be made a function where the input and output can be documented, etc.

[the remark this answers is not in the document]
w-henderson: Apart from the condition that we can't go through a saving state, it is a textbook implementation. However I do agree that having this condition embedded within the for loop is unnecessarily confusing (and also inefficient!) so I've moved it out of the for loop and documented exactly what it's doing and why it's there. Is this any better? @tmakatos

test/py/test_quiesce.py
tmakatos: Surreal discussion, no actual issue here.

tmakatos (approved):

PR #759, opened 2023-08-05
Fix MSI-X capability write logging opposite status

FlorianFreudiger:
Since the vfu_log call for whether MSI-X is enabled or not is using the not yet updated `msix->mxc.mxe`, it will always print the opposite value. ("enable MSI-X" <-> "disable MSI-X")

Instead of changing it to `new_msix.mxc.mxe`, I simply moved the assignment of both the `mxe` and the unaffected `fm` field before their log calls, so both variants are fine to use.

This is also how other functions like `handle_px_pxdc_write` do it, assignment before log.

### DIFF
--- a/lib/pci_caps.c
+++ b/lib/pci_caps.c
@@ -182,7 +182,7 @@ cap_write_msix(vfu_ctx_t *vfu_ctx, struct pci_cap *cap, char *buf,
 
     if (msix->mxc.fm != new_msix.mxc.fm) {
         msix->mxc.fm = new_msix.mxc.fm;
-        if (new_msix.mxc.fm) {
+        if (msix->mxc.fm) {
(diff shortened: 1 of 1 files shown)


lib/pci_caps.c
```diff
@@ -181,19 +181,19 @@ cap_write_msix(vfu_ctx_t *vfu_ctx, struct pci_cap *cap, char *buf,
      */
 
     if (msix->mxc.fm != new_msix.mxc.fm) {
+        msix->mxc.fm = new_msix.mxc.fm;
         if (new_msix.mxc.fm) {
```
jlevon, line 185: nit, could you make this refer to msix->mcx.fm now?

FlorianFreudiger: Sure!

jlevon (approved):

PR #761, opened 2023-08-14
Add another lspci test output variant

gierens:
On my Arch-based systems (Linux 6.1.44-lts) with pciutils 3.10.0 the lspci output is a little different causing the lspci test to fail. Do/Did you test it on a different version of pciutils, or is lspci just that on different distros?

Anyway I added the output I'm getting as and altered the test script accordingly.

### DIFF
--- a/test/lspci.expected.out.4
+++ b/test/lspci.expected.out.4
@@ -1,0 +1,30 @@
+00:00.0 Non-VGA unclassified device: Device 0000:0000
+	Control: I/O- Mem- BusMaster- SpecCycle- MemWINV- VGASnoop- ParErr- Stepping- SERR- FastB2B- DisINTx-
+	Status: Cap+ 66MHz- UDF- FastB2B- ParErr- DEVSEL=fast >TAbort- <TAbort- <MAbort- >SERR- <PERR- INTx-
+	Region 0: I/O ports at <unassigned> [disabled]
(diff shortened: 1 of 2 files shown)


gierens:
ah, just saw the workflow, yeah those distros all have other versions of pciutils apparently.

swapnili:
Currently we do not test on Arch platform in the github workflow. But feel free to add it and see if all test pass maybe other tests also needs some fixing. Thanks

jlevon (approved):

jlevon:
@gierens thanks. Would you mind altering your commit message to add a Signed-off-by, and also describe in more detail why you need the change (Arch version and lspci version please).

gierens:
So I added the signed-off and a few more details regarding the versions to the commit message ... sorry about forgetting the signed-off, just read that in the README must have overlooked it the first time. As to why I need this, well simply so the tests don't fail here. And since it merely adds a variant and doesn't affect anything for the distros you officially support or test against I though it couldn't hurt to at least mention this. At some point other distros might bump their pciutils versions as well.

jlevon:
To be clear, I wasn't asking why you need this, but to add that info to the commit message, which you've done now, thanks.

jlevon (approved):

jlevon:
@gierens if you like, we would also take a PR to add Arch to .github/workflows/pull_request.yml

PR #763, opened 2023-08-15
Introduce close_safely helper function

mnissler-rivos:
The helper function centralizes some extra checks and diligence desired by many/most current code paths but currently inconsistently applied. This includes bypassing the close call when the file descriptor is -1 already, resetting the file descriptor variable to -1 after closing, and preserving errno.

All calls to close are replaced by close_safely. Some warning log output is lost over this, but it doesn't seem like this was very useful anyways given that Linux always closes the file descriptor anyways.

### DIFF
--- a/lib/common.h
+++ b/lib/common.h
@@ -37,8 +37,10 @@
 #ifndef LIB_VFIO_USER_COMMON_H
 #define LIB_VFIO_USER_COMMON_H
 
+#include <errno.h>
 #include <limits.h>
 #include <stdint.h>
+#include <unistd.h>
 
 #define UNUSED __attribute__((unused))
 #define EXPORT __attribute__((visibility("default")))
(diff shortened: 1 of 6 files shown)


jlevon:
please signoff your commit

lib/common.h
```diff
@@ -79,6 +81,32 @@ _get_bitmap_size(size_t size, size_t pgsize)
...
 
+/*
+ * Closes the given file descriptor, and resets the value to -1. Preserves
+ * errno. Skips closing if *fd is -1.
+ */
+static inline void
+close_safely(int *fd)
+{
+    int errno_save = errno;
+    if (fd && *fd != -1) {
```
jlevon, line 92: nit, we compare pointers against NULL, so "fd != NULL" please.

lib/common.h
```diff
@@ -79,6 +81,32 @@ _get_bitmap_size(size_t size, size_t pgsize)
...
 }
 
+/*
+ * Closes the given file descriptor, and resets the value to -1. Preserves
+ * errno. Skips closing if *fd is -1.
+ */
+static inline void
+close_safely(int *fd)
+{
+    int errno_save = errno;
```
jlevon, line 91: super-nit: we call this "saved_errno" elsewhere.

jlevon (approved):
Thanks - but please don't use force push during PR review, as it deletes all context from previous reviews, and I can't tell what changed and so on.

mnissler-rivos:
> Thanks - but please don't use force push during PR review, as it deletes all context from previous reviews, and I can't tell what changed and so on.

I can still see the old commits, and the "Compare" links on the right show what has changed in a push. "View reviewed changes" seems broken though - is that what you are refering to?

Anyways, happy to follow whatever process you use for this project. Requested changes go into additional commits then?

jlevon:
"show changes since last review" gets broken. In addition, any context for previous review comments is lost. Obviously for a small change it's no big deal, but the next one will be different!

So yeah, during review, we use additional commits, and then "squash and merge" at the very end to have a single commit per logical change without a bunch of noise.

PR #765, opened 2023-08-15
More Distros for Pull-Request Workflow

gierens:
This does three things:

1. As suggested in #761 this adds a job for Arch Linux (base 20230723.0.166908), but since I was already at it also Debian 11 bullseye and Debian 12 bookworm, to the pull request workflow. 

2. During testing I stumbled across a linting error thrown by the newer flake8 version in the Arch job, see
Screenshot from 2023-08-15 16-06-34
I altered the mentioned equality checks in `test/py/libvfio_user.py` to fix this.

3. It also bumps the version of Github Actions to v3 in all workflows. v2 will apparently be deprecated over the summer, see https://github.blog/changelog/2023-06-13-github-actions-all-actions-will-run-on-node16-instead-of-node12-by-default/ and
Screenshot from 2023-08-15 16-36-55

I already tested this on my own runner, see https://github.com/gierens/libvfio-user/actions/runs/5868047364 or
Screenshot from 2023-08-15 16-35-27
seems to work. I'm also a bit surprised how fast the arch job is.

### DIFF
--- a/.github/settings.yml
+++ b/.github/settings.yml
@@ -108,6 +108,9 @@ branches:
           - ubuntu-20
           - ubuntu-22
           - fedora-35
+          - debian-11
+          - debian-12
+          - arch-202307
       # Required. Enforce all configured restrictions for administrators. Set to true to enforce required status checks for repository administrators. Set to null to disable.
       enforce_admins: true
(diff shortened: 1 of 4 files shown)


swapnili (approved):
thanks

jlevon:
thanks! Could you also add these to required status checks in settings.yml?

jlevon (approved):

PR #746, opened 2023-07-12
implement migration v2

w-henderson:
This commit adds the core implementation of migration v2 according to the spec.

PR #766, opened 2023-08-17
fix: incorrect number of dirty pages printed

w-henderson:
The `log_dirty_bitmap` function in `dma.c` would output the wrong number of dirty pages due to the `char` of the bitmap being sign-extended when implicitly being converted to `unsigned int` for `__builtin_popcount`. By adding an intermediate cast to `uint8_t` we avoid this incorrect behaviour.

See https://github.com/nutanix/libvfio-user/pull/746#discussion_r1297173318.


jlevon:
can you force push to correct the commit message too, including your paragraph explanation above?

jlevon (approved):

PR #771, opened 2023-08-29
Construct server capabilities using json-c

mnissler-rivos:
String formatting is hitting its limits: Adding another field is difficult given that we already branch on whether migration is enabled. This change constructs a JSON-C object instead so we can add what we need and serialize to a string afterwards.

### DIFF
--- a/lib/tran.c
+++ b/lib/tran.c
@@ -266,9 +266,9 @@ out:
 }
 
 /*
- * A json_object_object_add wrapper that takes ownership of *val unconditionally:
- * Resets *val to NULL and makes sure *val gets dropped, even when an error
- * occurs. Assumes key is new and a constant.
+ * A json_object_object_add wrapper that takes ownership of *val
+ * unconditionally: Resets *val to NULL and makes sure *val gets dropped, even
+ * when an error occurs. Assumes key is new and a constant.
  */
 static int
 json_add(struct json_object *jso, const char *key, struct json_object **val)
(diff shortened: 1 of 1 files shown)


lib/tran.c
```diff
@@ -265,36 +265,98 @@ recv_version(vfu_ctx_t *vfu_ctx, uint16_t *msg_idp,
...
+}
+
 static int
 send_version(vfu_ctx_t *vfu_ctx, uint16_t msg_id,
              struct vfio_user_version *cversion)
 {
     struct vfio_user_version sversion = { 0 };
     struct iovec iovecs[2] = { { 0 } };
-    char server_caps[1024];
+    struct json_object* jo_caps = NULL;
```
jlevon, line 351: " *jo_caps" please

lib/tran.c
```diff
@@ -265,36 +265,98 @@ recv_version(vfu_ctx_t *vfu_ctx, uint16_t *msg_idp,
...
+    const char* server_caps = NULL;
```
jlevon, line 352: ditto

lib/tran.c
```diff
@@ -265,36 +265,98 @@ recv_version(vfu_ctx_t *vfu_ctx, uint16_t *msg_idp,
...
-                    "\"max_data_xfer_size\":%u,"
-                    "\"migration\":{"
-                        "\"pgsize\":%zu"
-                    "}"
-                "}"
-             "}", SERVER_MAX_FDS, SERVER_MAX_DATA_XFER_SIZE,
-                  migration_get_pgsize(vfu_ctx->migration));
+    int ret;
+
+    if ((jo_caps = create_caps_object(vfu_ctx)) == NULL ||
```
jlevon, line 356: nit, I would rather this function returns a strdup()ed string and the json handling is entirely within it.

lib/tran.c
```diff
@@ -265,36 +265,98 @@ recv_version(vfu_ctx_t *vfu_ctx, uint16_t *msg_idp,
...
+#endif
+
+    if (ret < 0) {
+        json_object_put(*val);
+    }
+    *val = NULL;
+    return ret;
+}
+
+static json_object* create_caps_object(vfu_ctx_t *vfu_ctx) {
```
jlevon, line 294: * placement

jlevon: function brace on new line please

lib/tran.c
```diff
@@ -265,36 +265,98 @@ recv_version(vfu_ctx_t *vfu_ctx, uint16_t *msg_idp,
...
+    struct json_object *jo_migration = NULL;
+    struct json_object *jo_caps = NULL;
+    struct json_object *jo_top = NULL;
+    struct json_object *jo_tmp = NULL;
+
+    if ((jo_caps = json_object_new_object()) == NULL) {
+        goto error;
+    }
+
+    if ((jo_tmp = json_object_new_int64(SERVER_MAX_FDS)) == NULL ||
```
jlevon, line 304: consider a json_add_int64() on top of json_add() since you do this several times?

lib/tran.c
mnissler-rivos: fixing

lib/tran.c
mnissler-rivos: fixing

lib/tran.c
mnissler-rivos: fixing

lib/tran.c
mnissler-rivos: That should be simpler indeed. Will do.

lib/tran.c
mnissler-rivos: Good idea, will do.

jlevon (approved):

tmakatos (approved):
LGTM, just one nit

lib/tran.c
```diff
@@ -265,36 +265,113 @@ recv_version(vfu_ctx_t *vfu_ctx, uint16_t *msg_idp,
...
+json_add_uint64(struct json_object *jso, const char *key, uint64_t value)
+{
+    struct json_object *jo_tmp = NULL;
+
+    /*
+     * Note that newer versions of the library have a json_object_new_uint64
+     * function, but the int64 one is available also in older versions that we
+     * support, and our values don't require the full range anyways.
+     */
+    jo_tmp = json_object_new_int64(value);
```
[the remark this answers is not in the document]
mnissler-rivos: Yes, adding.

tmakatos (approved):

PR #772, opened 2023-08-29
Introduce client object in python tests

mnissler-rivos:
Thus far, the client end of the socket is the only piece of client state tracked in tests, for which a global `socket` variable has been used. In preparation to add more state, replace the `socket` global with a `client` global object that groups all client state.

### DIFF
--- a/test/py/libvfio_user.py
+++ b/test/py/libvfio_user.py
@@ -32,6 +32,7 @@
 #
 
 from types import SimpleNamespace
+import collections.abc
 import ctypes as c
 import array
 import errno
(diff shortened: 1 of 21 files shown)


test/py/libvfio_user.py
```diff
@@ -685,25 +686,60 @@ def connect_sock():
...
+        self.client_cmd_socket = None
+
+    def connect(self, ctx, capabilities={}):
+        self.sock = connect_sock()
+
+        effective_caps = {
+            "capabilities": {
+                "max_data_xfer_size": VFIO_USER_DEFAULT_MAX_DATA_XFER_SIZE,
+                "max_msg_fds": 8,
+                "twin_socket": False,
```
jlevon, line 703: I think this belongs in later PRs

mnissler-rivos: Correct, will move there.

test/py/libvfio_user.py
```diff
@@ -685,25 +685,42 @@ def connect_sock():
...
+
+        json = b'{ "capabilities": { "max_msg_fds": 8 } }'
+        # struct vfio_user_version
+        payload = struct.pack("HH%dsc" % len(json), LIBVFIO_USER_MAJOR,
+                              LIBVFIO_USER_MINOR, json, b'\0')
+        hdr = vfio_user_header(VFIO_USER_VERSION, size=len(payload))
+        self.sock.send(hdr + payload)
+        vfu_attach_ctx(ctx, expect=0)
+        fds, payload = get_reply_fds(self.sock, expect=0)
+        self.client_cmd_socket = socket.socket(fileno=fds[0]) if fds else None
```
tmakatos, line 706: This will never be assigned until the `twin_sock` changes, right?

test/py/libvfio_user.py
```diff
@@ -685,25 +685,42 @@ def connect_sock():
...
+        return self.sock
+
+    def disconnect(self, ctx):
+        self.sock.close()
+        self.sock = None
+        if self.client_cmd_socket is not None:
```
tmakatos, line 712: ditto

test/py/libvfio_user.py
mnissler-rivos: Ah, good catch, these lines ended up in the wrong patch. I'll move them out of this one.

tmakatos (approved):

PR #775, opened 2023-08-29
Describe the twin-socket feature in the spec

mnissler-rivos:


docs/vfio-user.rst
```diff
@@ -517,7 +549,11 @@ Reply
 ^^^^^
 
 The same message format is used in the server's reply with the semantics
-described above.
+described above. In case the client set ``twin_socket`` to true in its
+capabilities, the server may include a file descriptor to use for the
+server-to-client command channel in the reply. The index of the file descriptor
+in the ancillary data of the reply is given by the ``twin_socket`` capability
```
jlevon, line 555: is it? It's a boolean above.

mnissler-rivos: Good point. I had been reusing the name with a different type so it's clear this is about the same thing, but  `twin_socket_fd_index` will likely be less confusing. Let me know if you disagree.

jlevon: I think twin_socket on the outgoing (request) and twin_socket_fd_index on the reply ??

jlevon: actually, I think maybe have two: twin_socket both ways, and an *additional* twin_socket_fd_index in the reply

mnissler-rivos: What would be twin_socket's purpose in the reply? Do you have a scenario in mind where we'd want to encode twin_socket = false, but still provide a twin_socket_fd_index value?

If you're coming from a symmetry of feature presence point of view, then I would suggest we introduce `twin_socket` as a dictionary, carrying an (optional) `enable` boolean in the client message, and an optional `fd_index` in the server reply.

jlevon: dict sounds good

docs/vfio-user.rst
```diff
@@ -513,12 +537,44 @@ The migration capability contains the following name/value pairs:
...
+|          |         | in the reply message.                                  |
++----------+---------+--------------------------------------------------------+
+
 Reply
 ^^^^^
 
 The same message format is used in the server's reply with the semantics
 described above.
 
+If and only if the client has requested to enable twin-socket mode by setting
```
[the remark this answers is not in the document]
mnissler-rivos: Yes. Although as worded, the server would just omit the `fd_index` field as opposed to setting it to -1. Or it might leave out the `twin_socket` object entirely, matching what servers that don't know `twin_socket` would do.

I can add another paragraph to capture this if you like?

docs/vfio-user.rst
```diff
@@ -513,12 +537,57 @@ The migration capability contains the following name/value pairs:
...
+
+* If the field or the entire ``twin_socket`` object is missing, the server does
+  not support twin-socket mode or decided not to enable it. The client can
+  choose whether it wants to proceed without twin-socket mode, or close the
+  connection if not.
+* If ``twin_socket.fd_index`` is present, the client obtains the designated file
+  descriptor from the reply and monitors it for commands from the server.
+* If ``twin_socket.fd_index`` is present, but its value is not a valid index in
+  the file descriptor array, the client must abort and close the connection since
+  it would not be able to receive command from the server.
```
tmakatos, line 585: nit: commnads

tmakatos (approved):
LGTM

mnissler-rivos:
Btw. we might want to keep this open while the implementation MR is still in flight in case we end up deciding to make changes.

jlevon:
Sorry for the slow reply. On reflection I think I'd prefer two tweaks:

1. s/enable/supported/
2. put supported in the reply capabilities as well as the fd/fd index

My thinking here is that it's then much more like a negotiation of capabilities:

1. client: I support twin socket, please enable it if you do too
2. server: so do I, and here's the fd you should use


thoughts?

jlevon (changes requested):
per coment

mnissler-rivos:
> Sorry for the slow reply. On reflection I think I'd prefer two tweaks:
> 
>     1. s/enable/supported/
> 
>     2. put supported in the reply capabilities as well as the fd/fd index
> 
> 
> My thinking here is that it's then much more like a negotiation of capabilities:
> 
>     1. client: I support twin socket, please enable it if you do too
> 
>     2. server: so do I, and here's the fd you should use
> 
> 
> thoughts?

Works for me. And I'm assuming clients will bail and close the connection on inconsistent replies (such as `supported` being `true`, but `fd_index` absent).

jlevon:
> And I'm assuming clients will bail and close the connection on inconsistent replies (such as `supported` being `true`, but `fd_index` absent).

yes

tmakatos (approved):

PR #784, opened 2023-09-18
fix: minor memory bugs

w-henderson:


jlevon:
please add the actual reports from coverity to the commit msg

w-henderson:
@jlevon Will this actually resolve the report? Given that we're still doing pointer arithmetic on a pointer that doesn't point to an array?

jlevon:
> @jlevon Will this actually resolve the report? Given that we're still doing pointer arithmetic on a pointer that doesn't point to an array?

should do yes

jlevon (approved):

PR #786, opened 2023-10-02
replace bcopy() with memcpy()

jlevon:
For some unclear reason, clang-tidy believes bcopy() is insecure.
Regardless, it is deprecated, so replace usages with memcpy().


tmakatos (approved):
Can't find an answer online, perhaps because data flows from right to left and bcopy has src 1st and dst 2nd?

jlevon:
Hardly a "security" issue!
All I can think is that they have instrumentation for ASAN and so on that they didn't bother doing for bcopy..

Issue #770, opened 2023-08-29
arch linux CI workflow fails


tmakatos:
https://github.com/nutanix/libvfio-user/actions/runs/6011550391/job/16305137885?pr=769

```
pacman -Sy --noconfirm \

...

:: Proceed with installation? [Y/n] 
:: Retrieving packages...
Error: The operation was canceled.
```

@gierens any ideas?

jlevon:
It looks like it timed out.

gierens:
pacman has a download timeout of 10s if I'm not mistaken that can be disabled by:

![SmartSelect_20230831_064604_Samsung Internet.jpg](https://github.com/nutanix/libvfio-user/assets/49617392/0265e2d8-2305-4a82-af02-e638beec919f)

Not sure if you can also simply increase this, but one could simply use the timeout command on the pacman command. I can have a look at this later the day.

jlevon:
It's the entire workflow that's timing out though, I think, if you check the execution times. It shouldn't really take 10 minutes to update (!)

gierens:
Ah yeah, thank's for the hint, now I see it too in the workflow run. Hm, I'll sync my fork and try to reproduce this on my runner.

gierens:
I looked into this but I'm not able to reproduce this: https://github.com/gierens/libvfio-user/actions/runs/6066677383 ... ran it multiple times and the arch job always succeeds as the fastest.

What I did notice however is that the centos job sometimes comes dangerously close to the 10 minute mark (https://github.com/gierens/libvfio-user/actions/runs/6066677383/attempts/1):
![Screenshot from 2023-09-04 10-13-34](https://github.com/nutanix/libvfio-user/assets/49617392/bbd728b5-d75d-41aa-a094-9dfd02738304)

And also failed one time (https://github.com/gierens/libvfio-user/actions/runs/6066677383/attempts/3):
![Screenshot from 2023-09-04 10-13-39](https://github.com/nutanix/libvfio-user/assets/49617392/fde84c9f-5109-4aa1-a7b1-29e4fe742f56)
Apparently due to some connection issue:
![Screenshot from 2023-09-04 10-13-54](https://github.com/nutanix/libvfio-user/assets/49617392/322531a0-a018-46fa-b7f6-52a6fe865994)

So my best guess is, that something similar happened to the arch job on the mentioned PR ... some connection issue to the arch repos and then a timeout. Nothing to worry too much about.

[...]

PR #788, opened 2024-01-24
Fix DMA message size calculation

mnissler-rivos:
When performing DMA via VFIO-user commands over the socket, vfu_dma_transfer breaks large requests into chunks according to the client's maximum data transfer size negotiated at connection setup time. This change fixes the calculation of the chunk size for the case where the last chunk is less than the maximum transfer size.

Unfortunately, the existing test didn't catch this due to the request size being a multiple of that maximum data transfer size. Adjust the test to make the last chunk size a true remainder.


mnissler-rivos:
FWIW, the arch-202307 failure looks unrelated to this change.

jlevon:
arch appears to be fundamentally unstable, see #770

jlevon:
LGTM otherwise

mnissler-rivos:
Thanks for the quick reply!

> please add your signed-off-by

Gah, my favorite mistake. Fixed now.

tmakatos (approved):

jlevon (approved):

mnissler-rivos:
Thanks for the quick turnaround, much appreciated!

PR #792, opened 2024-03-26
Support 64 bits and prefetchable BARs

jfgd:
Add two new flags for lib user to request 64bits and/or prefetchable BARs.

Tested with a vfio-user client patched QEMU.


tmakatos (commented):
I don't know enough about 64-bit BARs I'm afraid, I'll have to read up a little bit.

tmakatos (commented):
Oh and some unit tests would be nice.

jlevon:
Thanks for the patch! As Thanos said, do you mind taking a look at adding a simple test in test/py/ ?

Can you share your qemu client patches?

Did you need further changes? Do you care about very large BAR sizes, or just locating them past 4g?

jfgd:
> Thanks for the patch! As Thanos said, do you mind taking a look at adding a simple test in test/py/ ?

Ok, I'll have a look at it

> Can you share your qemu client patches?

I simply used the branch  `vfio-user-p3.1` from `https://github.com/oracle/qemu.git`, no changes needed there

> Did you need further changes? Do you care about very large BAR sizes, or just locating them past 4g?

No, my use case was to expose a very large BAR of 2GiB (located above 4G), this patch is enough for this. Setting the PCI_BASE_ADDRESS_MEM_TYPE_64 bit made QEMU and the PCI host driver happy.

test/py/test_vfu_realize_ctx.py
```diff
@@ -73,6 +73,17 @@ def test_vfu_realize_ctx_pci_bars():
     ret = vfu_setup_region(ctx, index=VFU_PCI_DEV_BAR1_REGION_IDX, size=4096,
                            flags=(VFU_REGION_FLAG_RW | VFU_REGION_FLAG_MEM))
     assert ret == 0
+    ret = vfu_setup_region(ctx, index=VFU_PCI_DEV_BAR2_REGION_IDX,
+                           size=1048576,
```
tmakatos, line 77: Is there a specific reason why you use these values?

lib/libvfio-user.c
```diff
@@ -1729,6 +1729,12 @@ vfu_realize_ctx(vfu_ctx_t *vfu_ctx)
         if (!(vfu_ctx->reg_info[i].flags & VFU_REGION_FLAG_MEM)) {
             vfu_ctx->pci.config_space->hdr.bars[i].io.region_type |= 0x1;
         }
+        if ((vfu_ctx->reg_info[i].flags & VFU_REGION_FLAG_64_BITS)) {
```
tmakatos, line 1732: @jlevon in vfu_setup_region() we don't check whether the size is > 2^32 for 32-bit BARs, I think we should fail if `VFU_REGION_FLAG_64_BITS` isn't set.

jlevon: We can't support bars of that size anyway I think due to the region size uint32_t in private.h

test/py/test_vfu_realize_ctx.py
```diff
@@ -81,6 +92,8 @@ def test_vfu_realize_ctx_pci_bars():
     hdr = get_pci_header(ctx)
     assert hdr.bars[0].io == 0x1
     assert hdr.bars[1].io == 0
+    assert hdr.bars[2].io == 0x4
```
jlevon, line 95: I know :93 was already like this, but do you mind adding constants for these instead of bare values?

jlevon (changes requested):

lib/libvfio-user.c
```diff
@@ -1729,6 +1729,12 @@ vfu_realize_ctx(vfu_ctx_t *vfu_ctx)
...
+                vfu_ctx->pci.config_space->hdr.bars[i].raw |= PCI_BASE_ADDRESS_MEM_TYPE_64;
```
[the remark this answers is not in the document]
jfgd: ok I can, however linux/pci_regs.h does not have a convenient flag for this. So it will `mem.locatable = 2` or I can set new define in `pci_defs.h`

lib/libvfio-user.c
```diff
@@ -1729,6 +1729,12 @@ vfu_realize_ctx(vfu_ctx_t *vfu_ctx)
...
+        }
+        if ((vfu_ctx->reg_info[i].flags & VFU_REGION_FLAG_PREFETCH)) {
```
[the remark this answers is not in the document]
jlevon: > 1.3.2.2

> For a PCI Express Endpoint, 64-bit addressing must be supported for all BARs that have the
Prefetchable bit Set.

Am I misinterpreting this?

lib/libvfio-user.c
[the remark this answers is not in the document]
jlevon: please add a define there yeah, thanks.

jlevon (approved):

jlevon:
Thanks for the contribution!

PR #797, opened 2024-05-14
improve documentation on testing with qemu

jlevon:
Pull out the qemu testing documentation into a separate file, and expand
it with step-by-step instructions on how to use the gpio example.

In addition switch to the jlevon/master.vfio-user branch, which is much
more up to date than the previous Oracle series.

### DIFF
--- a/docs/qemu.md
+++ b/docs/qemu.md
@@ -1,0 +1,112 @@
+qemu usage walkthrough
+======================
+
+In this walk-through, we'll use an Ubuntu cloudimg along with the
+[gpio sample server](../samples/gpio-pci-idio-16.c) to emulate a very simple GPIO
+device.
+
+Building qemu
+-------------
+
+`vfio-user` client support is not yet merged into `qemu`. Instead, download and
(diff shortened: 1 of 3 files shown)


docs/qemu.md
```diff
@@ -0,0 +1,112 @@
...
+
+```
+rm -f /tmp/vfio-user.sock
+./build/samples/gpio-pci-idio-16 -v /tmp/vfio-user.sock &
+```
+
+Booting the guest OS
+--------------------
+
+Make sure your system has hugepages available:
```
tmakatos, line 58: nit, that shouldn't be necessary for the gpio sample

tmakatos (approved):

docs/qemu.md
jlevon: Yeah, I suppose that's true, but given we want it for SPDK usage, seems fine to say so here?

tmakatos: sure

swapnili (approved):

PR #833, opened 2026-02-05
Add support for IOMMUFD IOCTLs

benlwalker:
Linux added a new uapi called IOMMUFD for managing the IOMMU. In recent Linux releases, many of the vfio ioctls are now just legacy wrappers around this new interface. However, the new interface has a couple of important features that I'd like to use with vfio-user. Namely:
1) Support for allowing the server to allocate the IOVA. This is important because the server can return the true IOVA to the client directly, which can be used as an actual DMA address by hardware. If the client puts this address into a command, they don't have to rely on the server process to intercept the command and translate it.
2) Support for registering memory described by a dma-buf (e.g. memory on another PCI device)

I've tried to make this backward compatible by adding feature negotiation. Only if both client and server report support will it be used.

### DIFF
--- a/include/vfio-user.h
+++ b/include/vfio-user.h
@@ -52,6 +52,13 @@ extern "C" {
 
 #define VFIO_USER_DEFAULT_MAX_DATA_XFER_SIZE (1024 * 1024)
 
+/*
+ * vfio-user protocol commands
+ *
+ * Commands 1-18 are part of the base protocol. Commands 19+ implement
+ * optional protocol extensions that must be negotiated via capabilities
+ * during VFIO_USER_VERSION exchange.
+ */
 enum vfio_user_command {
     VFIO_USER_VERSION                   = 1,
     VFIO_USER_DMA_MAP                   = 2,
@@ -70,6 +77,22 @@ enum vfio_user_command {
     VFIO_USER_DEVICE_FEATURE            = 16,
     VFIO_USER_MIG_DATA_READ             = 17,
     VFIO_USER_MIG_DATA_WRITE            = 18,
+    /*
+     * iommufd protocol extension
+     *
+     * These commands require negotiating "iommufd": {"supported": true}
+     * capability during VFIO_USER_VERSION. They provide an alternative to
+     * the legacy VFIO_USER_DMA_MAP/UNMAP commands, mirroring the Linux
+     * kernel's iommufd API.
+     */
+    VFIO_USER_IOMMUFD_ALLOC_IOAS        = 19,
+    VFIO_USER_IOMMUFD_DESTROY_IOAS      = 20,
+    VFIO_USER_IOMMUFD_IOAS_MAP          = 21,
+    VFIO_USER_IOMMUFD_IOAS_UNMAP        = 22,
+    VFIO_USER_IOMMUFD_IOAS_COPY         = 23,
+    VFIO_USER_IOMMUFD_BIND              = 24,
+    VFIO_USER_IOMMUFD_ATTACH_PT         = 25,
+    VFIO_USER_IOMMUFD_DETACH_PT         = 26,
     VFIO_USER_MAX,
 };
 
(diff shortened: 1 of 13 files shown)


include/vfio-user.h
```diff
@@ -143,6 +166,90 @@ struct vfio_user_dma_unmap {
...
+struct vfio_user_iommufd_ioas_map {
+    uint32_t argsz;
+    uint32_t flags;
+#define VFIO_USER_IOMMUFD_MAP_WRITEABLE     (1 << 0)
+#define VFIO_USER_IOMMUFD_MAP_READABLE      (1 << 1)
+    uint32_t ioas_id;      /* Target IOAS */
+    uint32_t __reserved;
+    uint64_t user_va;      /* Userspace virtual address (for tracking) */
+    uint64_t length;
+    uint64_t iova;         /* Output: server-allocated IOVA */
```
benlwalker: In the real iommufd there's an extra flag defined that tells the system whether the client specifies the iova here or whether the server fills it in. I've only implemented the mode where the server fills it in. There's a few considerations here:

* If the client chooses the iova, then it's almost certainly not going to be the real DMA address that needs to be sent to the hardware. The system will assume that the server process intercepts and translates all commands.
* If the server chooses the iova then we can return the real DMA address and the server doesn't need to intercept anything (so in theory, clients could get more direct access to the NVMe queues some day, if we gain additional security features in other areas). But the client needs to translate from its own vaddr to the server returned iova on every I/O.

The biggest challenge comes in knowing when the server should do a translation. By only allowing server-choose mode with iommufd, then I know if we're in iommufd mode the server doesn't have to translate. If I allow both modes like in the real iommufd ioctl, I don't have a great way to figure out if the server should translate an address or not.

tmakatos (commented):
Thanks for the patch, just to be clear there's no interaction with `/dev/iommufd` but this add iommufd-like functionality, right? I need to familiarise myself first with iommufd before I can meaningfully review this.

lib/iommufd.h
```diff
@@ -0,0 +1,162 @@
...
+ * iommufd controller state.
+ * Stored per connection (in vfu_ctx), reset on client disconnect.
+ */
+typedef struct iommufd_controller {
+    struct vfu_ctx      *vfu_ctx;
+    bool                enabled;        /* iommufd mode negotiated */
+    bool                device_bound;   /* Device bound to iommufd */
+    uint32_t            bound_ioas_id;  /* Currently attached IOAS */
+    uint32_t            next_ioas_id;   /* For allocating new IOAS IDs */
+    TAILQ_HEAD(, iommufd_ioas) ioas_list;
```
tmakatos: An `iommufd_controller` has many IOASs and an IOAS has many IOVAs.

lib/iommufd.h
```diff
@@ -0,0 +1,162 @@
...
+ * Allocate a new IOAS.
+ * Returns the new IOAS ID on success, or -1 with errno set on failure.
+ */
+int
+iommufd_alloc_ioas(iommufd_controller_t *iommufd, uint32_t *ioas_id);
+
+/*
+ * Destroy an IOAS and all its mappings.
+ */
+int
```
tmakatos: -1 on error and sets `errno`? (elsewher as well?)

lib/libvfio-user.c
```diff
@@ -2225,13 +2275,39 @@ vfu_addr_to_sgl(vfu_ctx_t *vfu_ctx, vfu_dma_addr_t dma_addr,
     quiesce_check_allowed(vfu_ctx, __func__);
 #endif
 
+    /* Check iommufd mappings first if iommufd is enabled */
+    if (vfu_ctx->iommufd != NULL && iommufd_is_enabled(vfu_ctx->iommufd)) {
+        /* Validate that the IOVA is registered (for security) */
+        mapping = iommufd_find_mapping(vfu_ctx->iommufd, (uint64_t)(uintptr_t)dma_addr, len);
+        if (mapping != NULL) {
+            /* Found in iommufd - create SGL entry with special region index */
+            if (max_nr_sgs < 1) {
```
tmakatos: Need to check whether it's legitimate to pass `max_nr_sgs == 0` in the first place.

lib/libvfio-user.c
```diff
@@ -2225,13 +2275,39 @@ vfu_addr_to_sgl(vfu_ctx_t *vfu_ctx, vfu_dma_addr_t dma_addr,
...
+                return ERROR_INT(ERANGE) - 1; /* Need at least 1 SG entry */
+            }
+
+            /* Use INT_MAX as special region index to mark iommufd mapping */
```
tmakatos: Isn't it cleaner to a dedicated field for that?

lib/libvfio-user.c
```diff
@@ -2240,7 +2316,35 @@ vfu_sgl_get(vfu_ctx_t *vfu_ctx, dma_sg_t *sgl, struct iovec *iov, size_t cnt,
...
+            /* In iommufd mode, IOVA equals the server's virtual address, so use it directly */
+            iov[i].iov_base = sg->dma_addr;
+            iov[i].iov_len = sg->length;
+        } else {
+            /* Legacy DMA mapping - use existing dma_sgl_get logic */
+            if (sg->region >= vfu_ctx->dma->nregions) {
+                return ERROR_INT(EINVAL);
+            }
+
+            dma_memory_region_t *region = &vfu_ctx->dma->regions[sg->region];
```
tmakatos: Need to check why `dma_sgl_get` went away.

lib/libvfio-user.c
```diff
@@ -2401,6 +2505,223 @@ vfu_sg_is_mappable(vfu_ctx_t *vfu_ctx, dma_sg_t *sg)
...
+        vfu_log(vfu_ctx, LOG_ERR, "IOMMUFD_BIND: iommufd not enabled");
+        return ERROR_INT(EINVAL);
+    }
+
+    ret = iommufd_bind_device(vfu_ctx->iommufd);
+    if (ret < 0) {
+        return ret;
+    }
+
+    /* Return a dummy device ID (not used in vfio-user) */
```
tmakatos: AFAIK vfio-user doesn't use device IDs at all, check.

tmakatos (commented):
@benlwalker this is a very superficial, partial review, I haven't had a chance to read about iommufd just yet.

lib/libvfio-user.c
```diff
@@ -2225,13 +2275,39 @@ vfu_addr_to_sgl(vfu_ctx_t *vfu_ctx, vfu_dma_addr_t dma_addr,
...
+        if (mapping != NULL) {
+            /* Found in iommufd - create SGL entry with special region index */
+            if (max_nr_sgs < 1) {
+                return ERROR_INT(ERANGE) - 1; /* Need at least 1 SG entry */
+            }
+
+            /* Use INT_MAX as special region index to mark iommufd mapping */
+            /* In iommufd mode, IOVA equals the server's virtual address, so we can use it directly */
+            sgl[0].dma_addr = dma_addr;
+            sgl[0].region = INT_MAX;
```
jlevon: a more specific define than INT_MAX for marking these please.

lib/iommufd.c
```diff
@@ -0,0 +1,494 @@
...
+
+    iommufd->device_bound = true;
+    iommufd->bound_ioas_id = 0;
+
+    vfu_log(iommufd->vfu_ctx, LOG_DEBUG, "device bound to iommufd");
+
+    return 0;
+}
+
+int
```
jlevon: what does attach/detach actually *do* for vfio-user servers? The only difference I could see is that we check the bound ioas first in iommufd_find_mapping() - but we already have the mapping_hint there.

What am I missing?

sbates130272:
@benlwalker do you have an ETA on updates to this PR based on the comments above? I am keen on this given it addresses the issue to how a libvfio-user emulated device can access device memory passed into a client (like QEMU) via vfio-pci. I am happy to collaborate on this with you if you are open to that?

benlwalker:
> @benlwalker do you have an ETA on updates to this PR based on the comments above? I am keen on this given it addresses the issue to how a libvfio-user emulated device can access device memory passed into a client (like QEMU) via vfio-pci. I am happy to collaborate on this with you if you are open to that?

I think I am going to take a bit of a different approach and do a simpler extension to only support server-allocated IOVAs. My use case is also giving clients direct DMA access and I don't need full IOMMUFD ioctl support to do that. I'll try to push a new version today.

PR #842, opened 2026-04-09
File I/O DMA region access mode

mnissler-rivos:
This PR adds a new mode to perform DMA operations using file I/O system calls. This is useful in cases where the client can supply a file descriptor, but the file descriptor doesn't support `mmap()`. This is the case for `/proc/<pid>/mem` for example, which can be used to give the server access to the client's address space.


lib/libvfio-user.c
```diff
@@ -2395,6 +2413,40 @@ vfu_dma_transfer(vfu_ctx_t *vfu_ctx, enum vfio_user_command cmd,
         return ERROR_INT(EPERM);
     }
 
+    if (sg->region->access_mode == REGION_ACCESS_MODE_MMAP) {
```
jlevon, line 2416: hmm, there may be good reasons to want the client to service dma read/write even in the case of MODE_FILE_IO or MODE_MMAP. I wonder if we want a flags param for vfu_dma/read_write() so users can request this if needed?

include/vfio-user.h
```diff
@@ -122,6 +122,8 @@ struct vfio_user_dma_map {
     uint32_t argsz;
 #define VFIO_USER_F_DMA_REGION_READ     (1 << 0)
 #define VFIO_USER_F_DMA_REGION_WRITE    (1 << 1)
+#define VFIO_USER_F_DMA_REGION_MMAP     (1 << 2)
```
jlevon, line 125: this will need a qemu update to the protocol definition please, as well as ideally updating the qemu headers.

jlevon (commented):
Not sure I'm quite clear on the use case - when would I be sharing something like /proc files with the server?

include/vfio-user.h
mnissler-rivos: Yes, I'll be happy to send a patch for that, but wanted to align on the overall proposal here first.

lib/libvfio-user.c
mnissler-rivos: That thought crossed my mind as well. Can you think of any plausible reasons though? If the client did provide a file descriptor at registration time, it seems only fair that we use it :-D

I'll be happy to add parameter (or make separate functions) though if you think that's the best way forward.

mnissler-rivos:
> Not sure I'm quite clear on the use case - when would I be sharing something like /proc files with the server?

This is for a case where a non-VMM client wants to expose buffers to the server, and it isn't feasible in practice to change all memory allocation code paths in the client to use `memfd_create` (or equivalent) rather than `malloc`.

jlevon:
> > Not sure I'm quite clear on the use case - when would I be sharing something like /proc files with the server?
> 
> This is for a case where a non-VMM client wants to expose buffers to the server, and it isn't feasible in practice to change all memory allocation code paths in the client to use `memfd_create` (or equivalent) rather than `malloc`.

This seems like it would only be useful in full-trust scenarios - not something we have as a typical design principle - but regardless the basic plumbing functionality seems reasonable.

lib/libvfio-user.c
jlevon: I'd rather be paranoid in terms of API that we add a flags parameter, even if no bits are yet supported. Let me knock up a PR for that.

jlevon: one thing:

```
  764  * During live migration, this call does not mark any of the written pages as    
  765  * dirty; the client is expected to track this.                                  
```

We have an open issue on this, and whether we should in fact be marking them dirty ourselves, but that may be one reason for forcing this, potentially.

jlevon: (regardless, this comment needs updating for the pwrite case)

mnissler-rivos:
> > > Not sure I'm quite clear on the use case - when would I be sharing something like /proc files with the server?
> > 
> > 
> > This is for a case where a non-VMM client wants to expose buffers to the server, and it isn't feasible in practice to change all memory allocation code paths in the client to use `memfd_create` (or equivalent) rather than `malloc`.
> 
> This seems like it would only be useful in full-trust scenarios - not something we have as a typical design principle - but regardless the basic plumbing functionality seems reasonable.

Yes, this gives the server full access to the entire address space, and it's certainly not something that a VMM would want to do with a potentially hostile server. I would argue though that access via mmap can turn into a big foot-gun as well if not used carefully, so clients already have to be careful which file descriptors are OK to expose.

lib/libvfio-user.c
jlevon: When you rebase on my change please make sure to update the docs for these routines to explain behaviour for pwrite-based regions.

lib/dma.c
```diff
@@ -170,12 +166,12 @@ MOCK_DEFINE(dma_controller_remove_region)(dma_controller_t *dma,
 
     if (region->info.vaddr != NULL) {
         dma_controller_unmap_region(dma, region);
-    } else {
-        assert(region->fd == -1);
     }
 
     btree_iter_remove(&iter);
     dma_controller_increment_regions_generation(dma);
+    err = fd_cache_put(&region->fd);
```
jlevon, line 173: I think this is a bit confusing as it makes it look like region->fd is valid in all cases, but that's only true in MMAP|FILE_IO cases. Can we instead make :167 check for MMAP|FILE_IO flags and do the right thing in all 3 cases, including keeping the illustrative assert?

lib/dma.c
```diff
@@ -208,10 +205,10 @@ dma_controller_remove_all_regions(dma_controller_t *dma,
 
         if (region->info.vaddr != NULL) {
```
jlevon, line 206: as per above -  and seeing :200 I'm wondering should we use dma_controller_remove_region(region); here, and refactor with another
dma_controller_remove_region_by_addr(addr, size) on top, instead of having remove_region() take addr, size? 

So we're not duplicating the logic here.

lib/dma.c
jlevon, line 312: are we concerned at all about mmap vs file IO in this code?

lib/dma.c
```diff
@@ -342,7 +340,7 @@ MOCK_DEFINE(dma_controller_add_region)(dma_controller_t *dma,
         return ERROR_PTR(EINVAL);
     }
 
-    if (fd != -1) {
+    if (access_mode != REGION_ACCESS_MODE_MSG) {
         page_size = fd_get_blocksize(fd);
```
jlevon, line 344: please add assert(fd != -1); here

lib/dma.c
jlevon, line 411: please audit for these thata should be checking access_mode instead?

lib/libvfio-user.c
jlevon, line 2352: please make sure to update the docs for this to explain when it can be used

jlevon: I think we might want an explicit access_mode check here instead of the dma-level EFAULT? thoughts?

lib/libvfio-user.c
```diff
@@ -709,22 +718,31 @@ handle_dma_map(vfu_ctx_t *vfu_ctx, vfu_msg_t *msg,
         dma_map->flags &= ~VFIO_USER_F_DMA_REGION_WRITE;
     }
 
+    /* If not specified, default to mmap()-ed access if an fd is provided. */
+    access_mode = fd != -1 ? REGION_ACCESS_MODE_MMAP : REGION_ACCESS_MODE_MSG;
```
jlevon, line 722: bit confusing, I think, can we do instead:

if (REGION_MMAP) {
   if (fd == -1) error
   access_mode = mmap
} else if (REGION_FILE_IO) {
  if fd == -1 error
  access_mode = mmap
} else {
  /* Compatibility: if neither MMAP|FILE_IO, derive from fd */
  fd != -1 ? MMAP : MSG
}

i.e. put the compat case at the end instead of over-riding a default?

lib/libvfio-user.c
```diff
@@ -709,22 +718,31 @@ handle_dma_map(vfu_ctx_t *vfu_ctx, vfu_msg_t *msg,
...
         return ERROR_INT(EINVAL);
     }
 
-    if (msg->in.nr_fds > 0) {
-        fd = consume_fd(msg->in.fds, msg->in.nr_fds, 0);
-        if (fd < 0) {
-            vfu_log(vfu_ctx, LOG_ERR, "failed to add DMA region %s: %m", rstr);
-            return -1;
-        }
+    if (access_mode != REGION_ACCESS_MODE_MSG && fd == -1) {
```
jlevon, line 737: I think you can do these checks as I suggest above

lib/libvfio-user.c
mnissler-rivos: OK, I've now gated the fd-based access behind a flag and updated the documentation.

lib/dma.c
mnissler-rivos: Done.

lib/dma.c
mnissler-rivos: Done.

lib/dma.c
mnissler-rivos: No, checking file descriptor equivalence is appropriate for both cases.

lib/dma.c
mnissler-rivos: Done.

lib/dma.c
mnissler-rivos: I've reworked this code path to share code with region destruction, which addresses this too.

lib/libvfio-user.c
mnissler-rivos: I updated the docs, but I'm not sure what alternative you're suggesting for EFAULT?

lib/libvfio-user.c
mnissler-rivos: Done.

lib/libvfio-user.c
mnissler-rivos: Done.

tmakatos:
@mnissler-rivos are you waiting for a reveiw for this?

mnissler-rivos:
> @mnissler-rivos are you waiting for a reveiw for this?

Yes, this is ready for another round of review.

include/vfio-user.h
jlevon: sorry for the slow reply; could you draft up a patch for review please?

lib/libvfio-user.c
jlevon: I was suggesting that you add a check directly in vfu_sgl_get() that looks at access_mode and refuses the API directly as appropriate.

include/libvfio-user.h
```diff
@@ -697,7 +697,9 @@ vfu_addr_to_sgl(vfu_ctx_t *vfu_ctx, vfu_dma_addr_t dma_addr, size_t len,
  * vfu_sgl_put().
  *
  * This is only supported when a @dma_unregister callback is provided to
- * vfu_setup_device_dma().
+ * vfu_setup_device_dma(). In addition, the client must have registered the
+ * relevant DMA regions to allow mmap()-ed access and supplied file
+ * descriptors, otherwise this function will fail with EFAULT.
```
jlevon, line 702: as per my other comment, this should may be explicit check and ENOTSUP

include/libvfio-user.h
jlevon: needs updating

lib/dma.c
```diff
@@ -158,13 +158,42 @@ MOCK_DEFINE(dma_controller_unmap_region)(dma_controller_t *dma,
...
+        dma_unregister(data, &region->info);
+        dma->vfu_ctx->in_cb = CB_NONE;
+    }
+
+    switch (region->access_mode) {
+        case REGION_ACCESS_MODE_MSG:
+            assert(region->fd == -1);
+            break;
+        case REGION_ACCESS_MODE_MMAP:
+            if (region->info.vaddr != NULL) {
```
jlevon, line 184: is vaddr==NULL possible?

mnissler-rivos: Yes, in case we're getting called on `dma_controller_add_region`'s error exit path. IMO, better to make destructors robust to do the right thing also for half-constructed objects rather than duplicating teardown code.

lib/libvfio-user.c
mnissler-rivos: Hm, but then I would have to add a loop over the S/G list entries here, and the check would be redundant with the one in `dma_sgl_get`, so we'd never generate `EFAULT`, which seems like an unnecessary API change?

include/libvfio-user.h
mnissler-rivos: (see reply to other comment)

include/libvfio-user.h
mnissler-rivos: done

include/vfio-user.h
mnissler-rivos: I have a draft here: https://github.com/mnissler-rivos/qemu/compare/master...mnissler-rivos:qemu:vfio_user_access_mode

Let me post that to qemu-devel so we can do the spec change review there.

mnissler-rivos: FWIW: https://lists.gnu.org/archive/html/qemu-devel/2026-06/msg00356.html

lib/libvfio-user.c
jlevon: ok thanks

jlevon (approved):

test/mocks.c
```diff
@@ -112,13 +112,13 @@ unpatch_all(void)
 }
 
 dma_memory_region_t *
-dma_controller_add_region(dma_controller_t *dma, void *dma_addr,
-                          uint64_t size, int fd, off_t offset,
-                          uint32_t prot)
+dma_controller_add_region(dma_controller_t *dma, void *dma_addr, uint64_t size,
+                          int *fd, off_t offset, uint32_t prot,
```
mills-ng, line 116: @jlevon @mnissler-rivos This commit does not compile in my environment. The previous master commit ee13121be8cdd79b39e659b4371bfc8570752de9 compiles without issue.

I am in a Yocto 6.0 build environment with the following package versions:
gcc 15.2.0
cmake 4.3.1
cmocka 2.0.2
meson 1.10.2
ninja 1.13.2
json-c 0.18

Cmocka does not like the change from 'int fd' to 'int *fd'

`| x86_64-pokysdk-linux-gcc -march=x86-64 --sysroot=/opt/workspace/demo/yocto/build/target/work/x86_64-nativesdk-pokysdk-linux/nativesdk-libvfio-user/0.0.1/recipe-sysroot -Itest/unit_tests.p -Itest -I../sources/libvfio-user-0.0.1/test -Iinclude -I../sources/libvfio-user-0.0.1/include -Ilib -I../sources/libvfio-user-0.0.1/lib -I/opt/workspace/demo/yocto/build/target/work/x86_64-nativesdk-pokysdk-linux/nativesdk-libvfio-user/0.0.1/recipe-sysroot/usr/local/oe-sdk-hardcoded-buildpath/sysroots/x86_64-pokysdk-linux/usr/include/json-c -fdiagnostics-color=always -D_FILE_OFFSET_BITS=64 -Wall -Winvalid-pch -Wextra -std=gnu99 -O2 -fcanon-prefix-map -ffile-prefix-map=/opt/workspace/demo/yocto/build/target/work/x86_64-nativesdk-pokysdk-linux/nativesdk-libvfio-user/0.0.1/sources/libvfio-user-0.0.1=/usr/src/debug/nativesdk-libvfio-user/0.0.1 -ffile-prefix-map=/opt/workspace/demo/yocto/build/target/work/x86_64-nativesdk-pokysdk-linux/nativesdk-libvfio-user/0.0.1/build=/usr/src/debug/nativesdk-libvfio-user/0.0.1 -ffile-prefix-map=/opt/workspace/demo/yocto/build/target/work/x86_64-nativesdk-pokysdk-linux/nativesdk-libvfio-user/0.0.1/recipe-sysroot= -ffile-prefix-map=/opt/workspace/demo/yocto/build/target/work/x86_64-nativesdk-pokysdk-linux/nativesdk-libvfio-user/0.0.1/recipe-sysroot-native= -pipe -DUNIT_TEST -DWITH_TRAN_PIPE -D_GNU_SOURCE -Wno-missing-field-initializers -Wmissing-declarations -Wwrite-strings -DHAVE_LINUX_KCMP_H -MD -MQ test/unit_tests.p/mocks.c.o -MF test/unit_tests.p/mocks.c.o.d -o test/unit_tests.p/mocks.c.o -c ../sources/libvfio-user-0.0.1/test/mocks.c
| ../sources/libvfio-user-0.0.1/test/mocks.c: In function 'dma_controller_add_region':
| ../sources/libvfio-user-0.0.1/test/mocks.c:127:5: error: initialization of 'long int' from 'int *' makes integer from pointer without a cast [-Wint-conversion]
|   127 |     check_expected_int(fd);
|       |     ^~~~~~~~~~~~~~~~~~
| ../sources/libvfio-user-0.0.1/test/mocks.c:127:5: note: (near initialization for '(anonymous).int_val')
`

jlevon: please file an issue (and a PR if you have one)

jlevon: I think it should be check_expected_ptr() maybe

mills-ng: Yes, check_expected_ptr() works. I submitted #863.

PR #851, opened 2026-04-15
Fix scan-build lint

jlevon:
On the CentOS 7 build, scan-build incorrectly reports:

../../../../lib/tran_sock.c:104:26: warning: Access to field 'cmsg_level' results in a dereference of a null pointer (loaded from variable 'cmsg')

Add an assert for this.

### DIFF
--- a/lib/tran_sock.c
+++ b/lib/tran_sock.c
@@ -100,7 +100,9 @@ tran_sock_send_iovec(int sock, uint16_t msg_id, bool is_reply,
         msg.msg_control = buf;
         msg.msg_controllen = CMSG_SPACE(size);
 
-        struct cmsghdr * cmsg = CMSG_FIRSTHDR(&msg);
+        struct cmsghdr *cmsg = CMSG_FIRSTHDR(&msg);
+        /* Quieten clang-analyzer. */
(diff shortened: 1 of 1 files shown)


lib/tran_sock.c
```diff
@@ -100,7 +100,8 @@ tran_sock_send_iovec(int sock, uint16_t msg_id, bool is_reply,
         msg.msg_control = buf;
         msg.msg_controllen = CMSG_SPACE(size);
 
-        struct cmsghdr * cmsg = CMSG_FIRSTHDR(&msg);
+        struct cmsghdr *cmsg = CMSG_FIRSTHDR(&msg);
```
tmakatos, line 103: nit, add a comment that this specifically to silence a scan-build false positive (since we construct a correctly-sized header right above)?

(Also, TIL about `__builtin_assume`, which would avoid the need for the comment.)

tmakatos (approved):

lib/tran_sock.c
jlevon: I don't think that's supported by gcc

tmakatos (approved):

PR #860, opened 2026-05-26
always return ENOTCONN on connection reset

Hooollin:
When a connection reset (ENOMSG/ECONNRESET) triggers vfu_reset_ctx(), the early "return ret" would return the reset error instead of ENOTCONN, hiding the connection loss from the caller.


tmakatos:
As explained in #861 this is intentional. The call will evenutally get the connection loss error.

Hooollin:
> As explained in #861 this is intentional. The call will evenutally get the connection loss error.

The issue is on my side, I synced some community code into our repo and messed up a conflict resolution. Thank you for your help.

PR #865, opened 2026-06-14
pci: implement RW1C handling for PCI status register

zp78256pxd-ux:
## pci: implement RW1C handling for PCI status register

First of all, really impressed by libvfio-user — the codebase is clean and well-structured, and the Python test framework is a great approach for this kind of low-level emulation testing. Happy to be contributing.

### Summary

Implement RW1C (Read/Write 1 to Clear) handling for PCI Status register writes.

Previously, writes to `PCI_STATUS` were silently ignored, even though several status bits are defined as RW1C by the PCI 3.0 spec. A FIXME in the code acknowledged this:

```c
/* FIXME ignoring write completely is wrong as some bits are RW1C */
vfu_log(vfu_ctx, LOG_INFO, "write to status ignored");
```

### Changes

- Add `handle_status_write()` implementing correct RW1C semantics for all six RW1C bits 
      (`dpd`, `sta`, `rta`, `rma`, `sse`,`dpe`)
- Replace the ignored write path in `pci_hdr_write()` with a call to `handle_status_write()`
- Add regression test coverage in `test/py/test_pci_status.py`
- Wire new test into `test/py/meson.build`

### Validation

Executed:

```
meson test -C build --print-errorlogs
pytest test/py/test_pci_status.py -v
```

All unit and Python tests passed. The `test-lspci` failure appears unrelated to this change (environment/output formatting difference in the test environment).

### References
- PCI Local Bus Specification Revision 3.0, 6.2.3 (Device Status Register).


test/py/test_pci_status.py
```diff
@@ -0,0 +1,112 @@
+#
+# Copyright (c) 2026 Nutanix Inc.
```
tmakatos: wrong

lib/pci.c
```diff
@@ -251,6 +251,44 @@ handle_erom_write(vfu_ctx_t *ctx, vfu_pci_config_space_t *pci,
     return 0;
 }
 
+static int
+handle_status_write(vfu_ctx_t *ctx UNUSED,
+                    vfu_pci_config_space_t *pci,
+                    const char *buf)
+{
+    vfu_pci_hdr_sts_t w;
```
tmakatos: nit, rename to `device_status` or something more descriptive?

test/py/test_pci_status.py
```diff
@@ -0,0 +1,112 @@
...
+#
+
+from libvfio_user import *
+import struct
+
+ctx = None
+
+PCI_STATUS = 0x06
```
tmakatos: These should be in `libvfio_user.py`.

test/py/test_pci_status.py
```diff
@@ -0,0 +1,112 @@
...
+        sock,
+        VFU_PCI_DEV_CFG_REGION_IDX,
+        offset=PCI_STATUS,
+        count=2
+    )
+
+    return struct.unpack("<H", payload)[0]
+
+
+def test_pci_status_rw1c():
```
tmakatos, line 64: Can you add a docstring explaing what exactly what's being tested? There's 6 registers to test, are you testing all of them? Also, you should test that setting the RO registers doesn't have any effect.

test/py/test_pci_status.py
```diff
@@ -1,17 +1,37 @@
 #
-# Copyright (c) 2026 Nutanix Inc.
+# Copyright (c) 2021 Nutanix Inc. All rights reserved.
```
tmakatos: I should have been more explicit, you don't _have_ to add something, but if you it has to be correct. If in doubt, don't add anything.

test/py/test_pci_status.py
```diff
@@ -41,7 +61,16 @@ def read_status(sock):
 
 
 def test_pci_status_rw1c():
-
+     """
```
tmakatos: identation

tmakatos (commented):
BTW do you mind sharing what specific problem you've encountered that motivated you to fix this?

tmakatos:
> BTW do you mind sharing what specific problem you've encountered that motivated you to fix this?

Just to clear we don't hear that much of how libvfio-user is used, feel free to decline if you don't want to discuss this further.

test/py/test_pci_status.py
zp78256pxd-ux: tq for pointing out bro

zp78256pxd-ux:
I’m currently exploring ideas around reducing cloud infrastructure cost and building an eBPF-based agent, and while reading through virtualization-related components I came across libvfio-user.

This wasn’t triggered by a production issue — I noticed the PCI status handling and thought it would be a good area to contribute and learn more about the stack.

tmakatos (commented):
LGTM apart from the copyright header.

zp78256pxd-ux:
Okay brother will do.One idea I’ve been exploring is whether eBPF could complement traditional observability by collecting more direct kernel-level signals instead of relying only on exported metrics.

I’m curious whether you’ve seen similar approaches used for things like OOM root-cause analysis, VM efficiency, or infrastructure cost reduction.ur thought brother !!

test/py/test_pci_status.py
zp78256pxd-ux: Thanks for clarifying — I copied the copyright header from an existing test file when creating this test. Could you clarify what would be the correct header in this case? Should I remove it entirely or follow a different convention for new tests?

tmakatos: I was wrong about this, we do need the Copyright (year, company, author name) as well as the license.

tmakatos (commented):
We also need SOB in the commit message.

jlevon (approved):

jlevon:
thanks for contribution, as per above, we need SOB (you can make it a single commit), plus you have some python lint :)

zp78256pxd-ux:
Thanks for the review and feedback.

I updated the commit with Signed-off-by, cleaned up the commit message, and force-pushed the changes. I also reran the PCI status test locally and verified the RW1C coverage still passes.
Please take another look when convenient.

test/py/test_pci_status.py
tmakatos: You still have Nutanix as the company and John as the author.

tmakatos:
> Thanks for the review and feedback.
> 
> I updated the commit with Signed-off-by, cleaned up the commit message, and force-pushed the changes. I also reran the PCI status test locally and verified the RW1C coverage still passes. Please take another look when convenient.

As @jlevon said there's a lint error and you still have John and Nutanix on the header. Also, don't force push or squash commits, we'll do that when merging, makes it easier to ensure that the content of the commit stays the same.

Issue #461, opened 2021-05-04
need function-level reset support


jlevon:
Certain writes to the PCI Express capability are supposed to initiate a function-level reset (FLR). This should only happen if pxcap::pxdcap::flrc is set.

NVMe controllers are required to support FLR, and nvmf is dutifully setting this bit in the pxcap. But we just ignore writes to pxcap::pxdc::iflr today.

We need to add a new reset callback type to support FLR, wire it up in SPDK, and reset at least some of our internal state (not clear how much, but certainly not including detaching the socket client).

Cc @changpe1

changpe1:
OK, SPDK nvmf can support controller level reset, FLR will cause a controller level reset, some internal states are required to clear, such as live migration state.

tmakatos:
@swapnili can we close this?

jlevon:
we can either keep this open to track the SPDK-side work, or open a new ticket

swapnili:
Anything is fine, I kept it open just to track SPDK work @changpe1 is working on.

changpe1:
> Anything is fine, I kept it open just to track SPDK work @changpe1 is working on.

yeah, I'm working on it, let's keep it open until SPDK support it.

tmakatos:
From the spec (NVMe® over PCIe® Transport Specification, Revision 1.2), 3.8.5.3:

> Function Level Reset Capability (FLRC): A value of ‘1’ indicates the Function
supports the optional Function Level Reset mechanism. NVM Express controllers shall
support Function Level Reset.

jlevon:
We've done the necessary plumbing in libvfio-user, seems silly to have this open for SPDK changes. So closing in favour of https://github.com/spdk/spdk/issues/3903 need proper FLR support in vfio_user.c

PR #866, opened 2026-06-19
pci: fail FLR when reset callback is missing

zp78256pxd-ux:
## What this PR does

Return EINVAL when a guest triggers PCI Function Level Reset (FLR) but no reset callback is registered.

Previously the library logged:
"FLR callback is not implemented"
but returned success to the client.

This caused FLR requests to appear successful even though no reset occurred.

## Changes
- Return ERROR_INT(EINVAL) in handle_px_pxdc_write()
   when FLR is requested and reset callback is missing.
- Add regression test covering FLR with FLRC enabled and
   no registered reset callback.

## Testing

pytest test/py/test_pci_caps.py -k "write_px" -vv


tmakatos (commented):
This will break virtual NVMe in SPDK, see https://github.com/nutanix/libvfio-user/issues/461#issuecomment-4766975662

lib/pci_caps.c
```diff
@@ -336,6 +336,7 @@ handle_px_pxdc_write(vfu_ctx_t *vfu_ctx, struct pxcap *px,
 	    return call_reset_cb(vfu_ctx, VFU_RESET_PCI_FLR);
         } else {
             vfu_log(vfu_ctx, LOG_ERR, "FLR callback is not implemented");
+            return ERROR_INT(EINVAL);
```
jlevon: I think the idea was that the device might not have anything to do for FLR, but maybe that's unlikely.

I think we need to document this somewhere if we make this change?

tmakatos: @jlevon IIUC you're in favour of getting this merged? If we're not planning to work on proper support on SPDK then we should at least add a callback and make it a no-op?

jlevon: we do have one, right?

tmakatos: > we do have one, right?

 we do, got confused that this callback is specific to FLR but it isn't, we pass the reason

tmakatos:
> This will break virtual NVMe in SPDK, see [#461 (comment)](https://github.com/nutanix/libvfio-user/issues/461#issuecomment-4766975662)

Wrong, NVMe does provide a reset callback.

lib/pci_caps.c
tmakatos: > I think the idea was that the device might not have anything to do for FLR, but maybe that's unlikely.
> 
> I think we need to document this somewhere if we make this change?

If the device advertises FLR then it should provide a reset callback, so I think the correct fix should be to fail adding the Device Control capability (`pxdc`) if a reset callback hasn't been provided. This means we'll have to document that the reset callback must be configured before adding capabilities.

zp78256pxd-ux:
Thanks for the feedback — that makes sense to me. Catching this when the capability is added rather than when FLR is actually triggered does seem like the more correct place to enforce it, since it surfaces the misconfiguration at device setup time instead of waiting for a guest to hit it at runtime.

I’m fairly new to this codebase, so I don’t have much visibility into how existing consumers structure their setup calls. A couple of things I wanted to check before attempting a rework:

1. Would you want this approach folded into this PR, or is it more involved than fits here? It looks like it would mean checking the flrc bit inside vfu_pci_add_capability() rather than just in the FLR write path, so I wasn’t sure whether that’s something you’d prefer to handle separately or whether you’d like me to take a pass at it.
2. From the SPDK discussion earlier in the thread, is there any known consumer that currently registers the reset callback after adding the capability? If so, enforcing the ordering requirement could break that setup, so I wanted to flag it in case that affects the preferred approach.

Happy to take a shot at whichever direction makes sense — I just didn’t want to make assumptions given I don’t have the full picture of downstream usage yet.

tmakatos:
> Thanks for the feedback — that makes sense to me. Catching this when the capability is added rather than when FLR is actually triggered does seem like the more correct place to enforce it, since it surfaces the misconfiguration at device setup time instead of waiting for a guest to hit it at runtime.
> 
> I’m fairly new to this codebase, so I don’t have much visibility into how existing consumers structure their setup calls. A couple of things I wanted to check before attempting a rework:
> 
> 1. Would you want this approach folded into this PR, or is it more involved than fits here? It looks like it would mean checking the flrc bit inside vfu_pci_add_capability() rather than just in the FLR write path, so I wasn’t sure whether that’s something you’d prefer to handle separately or whether you’d like me to take a pass at it.

Yes it can be folded in this PR.

> 2. From the SPDK discussion earlier in the thread, is there any known consumer that currently registers the reset callback after adding the capability? If so, enforcing the ordering requirement could break that setup, so I wanted to flag it in case that affects the preferred approach.

Our API is technically unstable so it's acceptable to break it in this way.

> 
> Happy to take a shot at whichever direction makes sense — I just didn’t want to make assumptions given I don’t have the full picture of downstream usage yet.

lib/pci_caps.c
jlevon: I'm not sure myself, but perhaps this should be a failure at realize time - so setup can still be in any order.

tmakatos: That's better yes.

zp78256pxd-ux:
Updated per review feedback: moved the validation to vfu_realize_ctx() so the check happens during realization rather than when FLR is triggered, and added a regression test for the EINVAL case.

lib/libvfio-user.c
```diff
@@ -1798,7 +1798,18 @@ vfu_realize_ctx(vfu_ctx_t *vfu_ctx)
     if (vfu_ctx->pci.nr_caps != 0) {
         vfu_ctx->pci.config_space->hdr.sts.cl = 0x1;
     }
+    if (vfu_ctx->pci_cap_exp_off >= 0) {
+        struct pxcap *px;
 
+        px = (struct pxcap *)pci_config_space_ptr(vfu_ctx,
+		vfu_ctx->pci_cap_exp_off);
```
tmakatos: identation

tmakatos (commented):
just one nit

tmakatos:
@zp78256pxd-ux checks are failing

zp78256pxd-ux:
My current implementation validates this in vfu_realize_ctx(), which causes the samples/lspci example to fail because it advertises FLR support but does not register a reset callback.

Would you prefer:

1. keeping the validation in vfu_realize_ctx() and updating samples/lspci to provide a reset callback (or disable FLR), or
2. moving the validation to capability creation time as previously suggested?

tmakatos:
> My current implementation validates this in vfu_realize_ctx(), which causes the samples/lspci example to fail because it advertises FLR support but does not register a reset callback.
> 
> Would you prefer:
> 
> 1. keeping the validation in vfu_realize_ctx() and updating samples/lspci to provide a reset callback (or disable FLR), or
     
this

zp78256pxd-ux:
Updated samples/lspci to register a trivial reset callback when advertising FLR support. This keeps the sample compatible with validation in vfu_realize_ctx() requiring a reset callback when FLR is enabled.

Verified locally that samples/lspci builds and runs successfully with the change.

My local make pre-push currently fails due to cmocka 2.0.2 deprecation warnings in existing unit tests (expect_check() with -Werror), which appears unrelated to this change

samples/lspci.c
```diff
@@ -38,6 +38,15 @@
 
 #include "libvfio-user.h"
 
+static int
+lspci_reset_cb(vfu_ctx_t *vfu_ctx, vfu_reset_type_t type)
```
jlevon: use UNUSED instead of void casts please.

jlevon (changes requested):
sorry for slow response, I have one nit.

zp78256pxd-ux:
Updated as requested. The callback now uses `UNUSED` for the unused parameters instead of `(void)` casts.

jlevon (approved):

PR #872, opened 2026-07-07
samples: increase BAR1 size to 16 KiB

### LINKED ISSUE #871: samples/server fails to start: BAR1 size (0x3000) rejected by power-of-two validation from #832
Here's the exact text — just copy everything below and paste it into the GitHub description box:
## Description

After commit `850ded8` ("Support 64b bar region size"), `samples/server` fails during startup because BAR1 is configured with a non-power-of-two size.

The sample currently defines:

```c
const size_t bar1_size = 0x3000;
```

and documents it as:

```c
/*
 * Setup BAR1 to be 3 pages in size...
 */
```

However, `bar_region_valid()` now rejects BAR sizes that are not powers of two:

```c
if ((size & (size - 1)) != 0) {
    vfu_log(vfu_ctx, LOG_ERR,
            "BAR region size %zu must be a power of 2", size);
    return false;
}
```

As a result, running the sample fails with:
server: BAR region size 12288 must be a power of 2
server: failed to setup BAR1 region: Invalid argument

## Steps to reproduce

1.  Build the project from the current master branch.
2. Run:  ./build/asan/samples/server pipe

## Actual behavior

Initialization fails because BAR1 is configured with a 3-page (12 KiB) region.

## Additional context

`samples/server.c` (and `samples/client.c`, which shares the same `bar1_size = 0x3000`) predates the power-of-two validation introduced in #832. I'm not sure whether:

- `samples/server.c`/`samples/client.c` should be updated to use a power-of-two BAR size (e.g. `0x4000`), or
- the sample intentionally relies on a non-power-of-two BAR and the validation should be reconsidered.

Could you clarify which behavior is intended?

zp78256pxd-ux:
Summary

Increase the sample BAR1 size from 3 pages (12 KiB) to 4 pages (16 KiB), and update the sample client and accompanying comments to keep the sample migration flow consistent.

Changes

* Increase the sample server BAR1 size from 0x3000 to 0x4000.
* Update the sample client BAR1 size from 0x3000 to 0x4000.
* Update the migration iteration assertions to match the larger BAR1.
* Update the BAR1 comment to describe the new 4-page layout.

Validation

* Rebuilt the sample client and server.
* Verified the sample client/server migration flow completes successfully with the updated BAR1 size and migration iteration counts (16 and 17 iterations).

Fixes #871


jlevon (approved):

tmakatos (commented):
nit, can you briefly explain in the commit message why this fix is needed and and mention that this is is for client/server in the commit title?

Otherwise LGTM.

zp78256pxd-ux:
Thanks! I’ve updated the commit title to clarify that the change is for the sample client/server, and expanded the commit message to explain why the fix is needed.

tmakatos (approved):

PR #873, opened 2026-07-07
samples/client: get BAR1 size from region info

zp78256pxd-ux:
Summary

Replace the hardcoded BAR1 size in the sample client with the size reported by VFIO_USER_DEVICE_GET_REGION_INFO.

This removes the existing FIXME and ensures the client uses the BAR1 size provided by the server instead of relying on a hardcoded value.

Testing

* Built the sample client with ninja -C build/asan samples/client
* Ran the sample client/server migration workflow successfully

### DIFF
--- a/samples/client.c
+++ b/samples/client.c
@@ -301,7 +301,7 @@ mmap_sparse_areas(int fd, struct vfio_region_info *region_info,
 }
 
 static void
-get_device_region_info(int sock, uint32_t index)
+get_device_region_info(int sock, uint32_t index, size_t *region_size)
 {
     struct vfio_region_info *region_info;
     size_t cap_sz;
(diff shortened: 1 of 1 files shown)


zp78256pxd-ux:
While testing this change with ASan enabled, I noticed a memory leak reported in the sample client. It appears to originate somewhere in the migration path (do_migrate()/migrate_from()) and I don't believe this change introduces it, but I'm not certain of the exact source.Is this a known issue? Happy to investigate and address it in
a follow-up PR.

tmakatos:
> While testing this change with ASan enabled, I noticed a memory leak reported in the sample client. It appears to originate somewhere in the migration path (do_migrate()/migrate_from()) and I don't believe this change introduces it, but I'm not certain of the exact source.Is this a known issue? Happy to investigate and address it in a follow-up PR.

Not a known issue, feel free to fix.

tmakatos (approved):

